PR #16, opened 2015-04-03
Bower update

vrockai:
Enhancements:
* Travis build. We now will be able to check if the build is passing before merging it.

Fixes:
* I've updated the version, so now we shouldn't cause resolving issues to both our previous users and people using angular 1.3.*.
* After building, there were some uncommitted changes in dist. Probably forgotten to be added in some previous PR. I've added new dist/* files to this PR.
* No idea why bower.json is showing so many changes in the UI, I just changed few versions.

To Ken:
* Please publish this new version to the bower registry.


vrockai:
Tested with LiveOak

PR #18, opened 2015-07-13
Adding charting component to patternfly repo

erundle:
I added a card directive for use on dashboards in the main module (The HTML markup might need to be tweaked).

As for the charting. I am looking for some feedback here. I added a base c3 directive and quota/percentage used chart. I will be adding several other c3 based charts shortly. I did not add the patternfly.charts module to the main module since this requires the use of c3 which means that anyone who used the main module would also need to add c3 which they might want.

I now remember creating a project like this in the past where we ran into issues with requiring that project pull in extra resources. For example if one project decides to inject angular.sanitize they all need to if they are included in the main module. So for now I just left it out of the main module and if a team wants to use it they need to inject 'patternfly.charts' explicitly


.jshintrc
```diff
@@ -14,12 +14,13 @@
   "quotmark": "single",
   "regexp": true,
   "undef": true,
-  "unused": true,
+  "unused": "vars",
```
waldenraines: Why the softening of this jshint rule?

waldenraines:
I personally think that e49937f belongs in a separate pull request as it's not directly related to charting.

bower.json
```diff
@@ -33,7 +33,8 @@
   },
   "dependencies": {
     "angular": "1.2.25 - 1.3.*",
-    "patternfly": "1.1.1 - 1.1.4"
+    "patternfly": "1.3.0",
+    "c3": "https://github.com/masayuki0812/c3.git#0.4.11-rc1"
```
waldenraines: Why are we using the full path here instead of just `0.4.11-rc1`?

src/card/card-directive.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * @ngdoc directive
+ * @name patternfly.card:pfCard
+ * @restrict A
+ * @element ANY
+ * @param {headtitle=} Title for the card - required
+ * @param {subtitle=} Subtitle for the card - optional
+ *
+ * @description
+ * Directive for easiliy displaying a html with transcluded content
```
waldenraines: Some typos, I think this should read:

```
Directive for easily displaying html with transcluded content
```

or even "...displaying a card with..."

src/card/card-directive.js
```diff
@@ -0,0 +1,34 @@
...
+
+angular.module('patternfly.card', []).directive('pfCard', function() {
+    'use strict';
+
+    return {
+        restrict: 'A',
+        transclude: true,
+        templateUrl: 'card/card.html',
+        scope: {
+            headtitle: '@',
```
waldenraines: Should this be `headTitle` as it's two words?

src/charts/c3/c3.directive.js
```diff
@@ -0,0 +1,52 @@
...
+    var vm = this;
+    vm.chartId = 'myChartId';
+    vm.chartConfig = { 'labels': ['Created','Deleted'], 'tooltipSuffixes': ['Container Group','Container Group'] };
+ }
+ </file>
+ */
+(function(c3){
+  'use strict';
+
+  angular.module('patternfly.charts')
```
waldenraines: I wonder if we ought to put c3 into its own module?  Maybe `patternfly.charts.c3`?  That would allows us to list it as a dependency [here](https://github.com/patternfly/angular-patternfly/pull/18/files#diff-8cfb720458f8dbe475bc9fbd0e7050cfR8).

waldenraines: > I wonder if we ought to put c3 into its own module? Maybe patternfly.charts.c3? That would allows us to list it as a dependency here.
> 
> Or we could create an angular-c3 or some such as it's own upstream project.

Scratch that.  Are we not able to use either of these for wrapping c3?

https://github.com/GraFiddle/angular-chart
https://github.com/jettro/c3-angular-directive

src/charts/progress/progress-chart-directive.js
```diff
@@ -0,0 +1,60 @@
+/**
+ * @ngdoc directive
+ * @name patternfly.charts.directive:pfPercentageUsed
+ *
+ * @description
+ *   Directive for rendering a percentage used progress chart. Will render one or more
+ *   bars based on data 
+ *
+ *
+ * @param {string} charts model data to be displalayed
```
waldenraines: Typo: `displayed`.

src/charts/progress/progress-chart-directive.js
```diff
@@ -0,0 +1,60 @@
...
+          //Calculate the perentage used  
+          angular.forEach($scope.charts, function(chart, index) {
+            chart.percentageUsed = 100 * (chart.start/chart.end);
+          }, $scope.charts);
+
+          //Animate in the chart load.
+          $scope.animate = true;
+		      $timeout(function(){
+                $scope.animate = false;
+              }, 0);
```
waldenraines: The spacing here seems to be off off.

src/charts/progress/progress-chart-directive.js
```diff
@@ -0,0 +1,60 @@
...
+    restrict: 'A',
+    scope: {
+      charts: '=',
+    },
+    replace: true,
+    templateUrl: 'charts/progress/progress-chart.html',
+    link: function($scope) {
+      $scope.$watch('charts', function(newVal, oldVal){
+        if (typeof(newVal) !== 'undefined') {
+          //Calculate the perentage used
```
waldenraines: Typo: "percentage".

src/card/card.html
```diff
@@ -0,0 +1,9 @@
+<div class="tile-pf">
+    <div class="tile-pf-heading">
+        <h2 class="tile-pf-title">{{headtitle}}</h2>
+    </div>
+    <span ng-if="subtitle" class="tile-pf-subtitle">{{subtitle}}</span>
+    <div class="tile-pf-body">
+        <ng-transclude></ng-transclude>
+    </div>
```
waldenraines: The spacing is off in this file.  For html we use 2 spaces.

src/charts/progress/progress-chart.html
```diff
@@ -0,0 +1,12 @@
...
+	<div class="quota-chart" ng-repeat="chart in charts">
+    	<div class="quota-chart-title">
+       		<span>{{ chart.title }}</span>
+        	<span class="used">{{chart.start}} of {{chart.end}}</span>
+    	</div>
+    	<div class="quota-chart-bar">
+        	<div class="quota-chart-used" ng-class="{'animate': animate}" style="width:{{chart.percentageUsed}}%"></div>
+        	<div class="quota-chart-unused"></div>
+    	</div>
+	</div>
```
waldenraines: The spacing is off in this file.  For html we use 2 spaces.

waldenraines: This, by the way, is a good argument for using something like [`grunt-htmlhint`](https://github.com/yaniswang/grunt-htmlhint).

waldenraines:
Are you proposing a change in where we store our views?  I notice `src/card/card.html` and ` src/charts/progress/progress-chart.html` rather than `src/card/views/card.html` and `src/charts/progress/views/progress-chart.html`.

For a small project like this I actually like storing the HTML files in the same directory as the views as you have it above.  If we are going that route, however, I would like to see a github issue opened to update the other views for consistency.

src/charts/progress/progress-chart.html
waldenraines: We should add this to the `.editorconfig` file:

```
[*.html]
indent_size = 2
```

waldenraines:
The changes that move the form directive should be their own PR as they are not related to charting at all.

test/charts/progress/progress.spec.js
```diff
@@ -0,0 +1,36 @@
+describe('Directive: pfPercentageUsed', function() {
```
waldenraines: Shouldn't the name of this file match the name of the directive?  I would think this should be named `pf-percentage-used.spec.js`.

test/charts/c3/c3.spec.js
```diff
@@ -0,0 +1,25 @@
+describe('Directive: c3Chart', function() {
```
waldenraines: Shouldn't the name of this file match the name of the directive? I would think this should be named `c3-chart.spec.js`.

test/card/card.spec.js
```diff
@@ -0,0 +1,27 @@
+describe('Directive: pfCard', function() {
```
waldenraines: Shouldn't the name of this file match the name of the directive? I would think this should be named `pf-card.spec.js`.

waldenraines:
I believe `card-directive.js` should be renamed to `pf-card.directive.js` to match the name of the directive, our existing pattern, and [best practices](https://github.com/johnpapa/angular-styleguide#naming).  Same thing for `progress-chart-directive.js`.

dist/angular-patternfly.js
```diff
@@ -1,4 +1,13 @@
 /**
+ * @name  patternfly
+ *
+ * @description
+ *   Charts module for patternfly. Must Inlcude d3.js and c3.js to use
```
waldenraines: Typo: "include".

waldenraines:
The change to using the function form of `'use strict';`, while a good change, should be it's own commit and PR.

waldenraines:
> As for the charting. I am looking for some feedback here. I added a base c3 directive and quota/percentage used chart. I will be adding several other c3 based charts shortly. I did not add the patternfly.charts module to the main module since this requires the use of c3 which means that anyone who used the main module would also need to add c3 which they might want.
> 
> I now remember creating a project like this in the past where we ran into issues with requiring that project pull in extra resources. For example if one project decides to inject angular.sanitize they all need to if they are included in the main module. So for now I just left it out of the main module and if a team wants to use it they need to inject 'patternfly.charts' explicitly

I think this is a fine approach and, really, we should do this with all of our submodules and create a `patternfly-all` module that includes everything rather than including it all in the base `patternfly` module.

.jshintrc
erundle: If you have a signature that requires (x,y) but you only use y jshint will fail. I also prefer having the entire function signature since it makes code more readable.

bower.json
erundle: I generally list the entire GIT repo path if the library isn't well known. This way it is clear to what is being pulled in

erundle:
Yes I was suggesting that we get rid of the "view" folder since in 95% of cases you will only have one *.html file per component so it seems like overkill. I would prefer to see folders for each component and then be flat from there. I didn't do that for all files but I can make that change.

I would also like to add a watch task to generate the dist js files. Having this auto build helps greatly if you are using bower link which I highly recommend doing if you are doing this type of development work.

I will address the typos and spacing in the branch and will close this MR in couple of days once I get additional feedback. I didn't realize this included an editorconfig file so I was manually setting my IDE formatting settings :).

.jshintrc
waldenraines: Not according to the [jshint docs](http://jshint.com/docs/options/#unused):

```
The default (true) behavior is to allow unused parameters that are followed by a used parameter.
```

waldenraines:
> Yes I was suggesting that we get rid of the "view" folder since in 95% of cases you will only have one *.html file per component so it seems like overkill.

I tend to agree and am in favor of this flattening of the directory structure.

> I didn't do that for all files but I can make that change

I would prefer a filing a new issue for this and opening a new PR.  This PR is already overloaded with changes and should be reduced to just the charting changes as mentioned above by @dtaylor113 and myself.

> I would also like to add a watch task to generate the dist js files. Having this auto build helps greatly if you are using bower link which I highly recommend doing if you are doing this type of development work.

If we were to add something like that it would be nice to have it turned on by a [grunt parameter](http://gruntjs.com/api/grunt.option) as it's an expensive operation.

erundle:
Closing PR. Will put in a PR for flat directory restructure, another for grunt changes, and a 3rd for charts when ready.

PR #19, opened 2015-07-17
Removing the view folders and putting each component in its own folder

erundle:
Removing the view folders and putting each component in its own folder. Most of the time there is only one view per component so the folder seems unnecessary. I added "component" folders within the sub modules since this will help with organization as we increase the number of sub modules


erundle:
Does anyone have issues with this? Can I or someone else merge this in? There is no need to tag this change.

waldenraines:
Looks good to me.

jeff-phillips-18:
Agreed, looks good. I will merge shortly if no one objects.

PR #22, opened 2015-07-23
Update for patternfly 2.0.0

erundle:
- Updating dependency to patternfly 2.0.0
- Updating to angular 1.3.X
- Adding a datepicker directive to wrap bootstrap date picker


waldenraines:
> Adding a datepicker directive to wrap bootstrap date picker

This should be its own separate PR in my opinion.

erundle:
The datepicker wasn't available until Patternfly 2.0 which was the entire reason I made this upgrade. We also have community members waiting on this so I would like to move forward with this unless anyone else has a huge objection or sees issues. 
https://github.com/patternfly/patternfly/issues/71

kenfinnigan:
I agree with @waldenraines, there are far too many changes in this one PR.

My concern is that this one PR does:
 - Update to PatternFly 2
 - Modify angular-patternfly release version scheme
 - Add date picker component
 - Update to AngluarJS 1.3
 - Restrict the supported versions of AngularJS and PatternFly

Also, I'm not sure why Gruntfile.js has essentially a full file change. I can only presume there is some different formatting, but it's impossible to know for sure.

erundle:
Closing this PR and putting a new one in that restricts to Patternfly 2. Will then put in a second PR for the datepicker additions

PR #26, opened 2015-07-29
Adding datepicker directive to wrap pf jquery based bootstrap datepicker

erundle:
Adding date picker directive to wrap the datepicker that was added into base patternfly.


dist/angular-patternfly.js
```diff
@@ -76,6 +76,80 @@ angular.module('patternfly.autofocus', []).directive('pfFocused', function($time
...
+    },
+    link: function($scope, element) {
+
+      //Make sure the date picker is set with the correct options
+      element.datepicker($scope.options);
+
+      //Set the initial value of the date picker
+      element.datepicker('update', $scope.date);
+
+
```
waldenraines: Extra line here.

src/form/datepicker/datepicker.directive.js
```diff
@@ -0,0 +1,74 @@
...
+    },
+    link: function($scope, element) {
+
+      //Make sure the date picker is set with the correct options
+      element.datepicker($scope.options);
+
+      //Set the initial value of the date picker
+      element.datepicker('update', $scope.date);
+
+
```
waldenraines: Extra blank line here.

waldenraines:
Is there a reason why we can't use the [ui-bootstrap datepicker](https://angular-ui.github.io/bootstrap/#/datepicker)?

src/form/datepicker/datepicker.html
```diff
@@ -0,0 +1,3 @@
+<div class="input-group date">
+	<input type="text" class="form-control"/><span class="input-group-addon"><span class="fa fa-calendar"></span></span>
```
waldenraines: Looks like an extra tab here.

waldenraines: Also, it think this would be a lot more readable if the html was formatted thusly:

```html
<div class="input-group date">
  <input type="text" class="form-control"/>
  <span class="input-group-addon">
    <span class="fa fa-calendar"></span>
  </span>
</div>
```

test/form/datepicker/datepicker.directive.spec.js
```diff
@@ -0,0 +1,46 @@
...
+  beforeEach(module('patternfly.form', 'form/datepicker/datepicker.html'));
+
+  beforeEach(inject(function(_$compile_, _$rootScope_) {
+    $compile = _$compile_;
+    $scope = _$rootScope_;
+  }));
+
+  var compileDatepicker = function(markup, $scope) {
+    var el = $compile(markup)($scope);
+    $scope.$apply();
```
waldenraines: I've usually seen `$scope.$digest()` used in directive testing instead of `$scope.$apply()`.  Any reason for using `$apply()`?

test/form/datepicker/datepicker.directive.spec.js
```diff
@@ -0,0 +1,46 @@
...
+    return el;
+  };
+
+  it("should set the date picker input", function() {
+    $scope.options = {};
+
+    var datepicker = compileDatepicker('<form><div pf-datepicker options="options" date="date"></div></form>', $scope);
+    var dateInput = angular.element(datepicker).find('input');
```
waldenraines: I personally prefer declaring all `var`s together at the top of the function.

If we don't already have this codified I think we should.

test/form/datepicker/datepicker.directive.spec.js
```diff
@@ -0,0 +1,46 @@
...
+
+    var newDate = new Date("October 4, 2015 11:13:00");
+    dateInput.datepicker('update', newDate);
+
+    $scope.$digest();
+    console.log($scope.date);
```
waldenraines: Errant `console.log()`.

erundle:
Nothing is stopping a developer from using the version in angular ui but that is not the "recommended" Patternfly design. Base Patternfly is using the following: https://github.com/eternicode/bootstrap-datepicker and there are dom differences so it is not a 1-1 map. This directive wraps that datepicker

test/form/datepicker/datepicker.directive.spec.js
erundle: apply calls digest but you are right it isn't needed here so I will change this and also address your other comments.

erundle:
I added a commit with the changes you suggested to this PR. Please let me know if this is OK or if you would like me to do something else.

jeff-phillips-18:
looks good to me

PR #28, opened 2015-08-05
Adding basic card directive

erundle:
Used so that we can create dashboard like cards without specifying the wrapper HTML each time allowing us to focus on the development of the inner card content.


dist/angular-patternfly.js
```diff
@@ -1,4 +1,12 @@
 /**
+ * @name  patternfly card
+ *
+ * @description
+ *   Card module for patternfly.
+ *
+ */
+angular.module('patternfly.card', []);
+;/**
```
waldenraines: Errant `;`?

src/card/basic/card-basic.html
```diff
@@ -0,0 +1,9 @@
+<div ng-class="showTopBorder === 'true' ? 'card-pf card-pf-accented' : 'card-pf'" class="cardContainer">
```
waldenraines: What is the purpose of this cardContainer class?  I don't see it used in css anywhere.

erundle: Yes you are right. It was only for testing purposes but I can access it via card-pf instead

jeff-phillips-18:
LGTM

waldenraines:
Shouldn't the naming of the file "card-basic.directive.js" match the directive name of `pfCard`?

erundle:
Sure renamed to card.directive.js

PR #29, opened 2015-08-06
Donut Chart

dtaylor113:
Donut Chart: made thresholds user defined, added ng-docs, created unit test


Gruntfile.js
```diff
@@ -1,223 +1,225 @@
...
+    grunt.registerTask('ngdocs:view', ['build', 'connect:docs', 'watch']);
+    grunt.registerTask('lint', ['jshint', 'htmlhint']);
+    grunt.registerTask('test', ['karma']);
+    grunt.registerTask('help', ['availabletasks']);
+
+  }
+
+  init({});
+
+};
```
waldenraines: Need to fix your editor config, it seems like there is some issue with whitespace changes here.

src/charts/donut/donut-percent-chart-directive.js
```diff
@@ -1,3 +1,61 @@
...
+       $scope.data = {
+           'used': '950',
+           'total': '1000'
+       };
+
+   };
+ </file>
+ </example>
+ */
 angular.module('patternfly.charts').directive('pfDonutPctChart', ['ChartsMixin', '$timeout',
```
waldenraines: It's good practice to have the directive name match the filename.

src/charts/donut/donut-percent-chart-directive.js
```diff
@@ -68,6 +133,16 @@ angular.module('patternfly.charts').directive('pfDonutPctChart', ['ChartsMixin',
           scope.config.donutConfig.data = scope.getDonutData(scope);
           setupDonutChartTitle();
         });
+        attrs.$observe('donutConfig', function() {
```
waldenraines: Why not use `$scope.$watch()` here?  If this only happens once then you should be able to unbind the watcher after it happens and thus prevent an infinite loop.

test/charts/donut/donut-pct.spec.js
```diff
@@ -0,0 +1,46 @@
...
+    return el;
+  };
+
+  it("should set the middle label", function() {
+    element = compileDonut('<div pf-donut-pct-chart config="config" data="data"></div>', $scope);
+    //$timeout.flush();
+
+    // Unit Test fails at this point with "MutationObserver is not defined"
+    // Need to upgrade to PhantomJS 2.0  which has MutationObserver support built in
+    // we are currently using PhantomJS 1.9.8.
```
waldenraines: Is it possible to remove the `$timeout` in `setupDonutChartTitle()` instead?

dtaylor113:
Hi,
More requirements were defined today, so I'm going to close this pull request with out merging into #Charts.  I will address the comments here, add new requirements, and do a pull request into #master.  -Thanks

Issue #27, opened 2015-08-03
Bower.json 'main' has minified and unminified files


stackus:
The main section for bower.json lists both the unminified and minified versions of the code. Are both really needed? 

We're using wiredep to include our bower components automatically and it is picking up both of these files.

erundle:
I took a look at the other angular projects and it looks like they only include the non minified version in main even though the send down the minified and unminified files. Ex: Angular just has "main": "./angular.js", but sends down angular.js and angular.min.js. Would "main": "dist/angular-patternfly.js" be sufficient for you?

stackus:
It would work great! Either would work but having the non-minified version helps with debugging.

PR #30, opened 2015-08-11
Update main to only include non minified version

erundle:
Resolves https://github.com/patternfly/angular-patternfly/issues/27

This is similar to how the angular repos and jquery etc are structured. They deliver minified and unminified versions but only the unminified version are listed in main:

Angular:
"main": "./angular.js",
jQuery:
"main": "dist/jquery.js",

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -23,10 +23,7 @@
     "README.md",
     "bower.json"
   ],
-  "main": [
-    "dist/angular-patternfly.js",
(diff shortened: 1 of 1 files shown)


waldenraines:
:+1: 

"Only one file per filetype." - https://github.com/bower/bower.json-spec#main

LGTM

jeff-phillips-18:
LGTM

PR #33, opened 2015-08-19
Donut Chart directive, styles, ng-docs, and unit test

dtaylor113:
The styles for the donut chart need to live in angular-patternfly#master (in new angular-patternfly.css file) since the donut chart implementation is being generated by angular-patternfly (there is no equivalent in patternfly#master)


dist/styles/angular-patternfly.css
```diff
@@ -0,0 +1,29 @@
+.donut-chart-pf {
+	width: 100%;
```
waldenraines: This spacing seems a little extreme.  Can we use 2 or 4?

src/charts/c3/c3.mixin.js
```diff
@@ -0,0 +1,57 @@
+angular.module('patternfly.charts').factory('ChartsMixin', function() {
```
waldenraines: The name of the file or the factory name should match.  This should probably be `charts-mixin.factory.js` to match the established pattern.

src/charts/c3/c3.mixin.js
```diff
@@ -0,0 +1,57 @@
...
+        return {
+            title: title,
+            label: {
+                show: false
+              },
+              width: 10
+            };
+      };
+
+    var getDefaultDonutSize = function() {
```
waldenraines: It's best practice to declare all `var`s at the top of the function.

src/charts/c3/c3.mixin.js
```diff
@@ -0,0 +1,57 @@
+angular.module('patternfly.charts').factory('ChartsMixin', function() {
+    'use strict';
+
+    // Donut Chart Defaults
+    var getDefaultDonut = function(title) {
```
waldenraines: Don't we usually put a space after the `function` keyword ala `function (title) {`

src/charts/c3/c3.mixin.js
waldenraines: Should this be a factory?  I feel like you could accomplish the same thing using a constant.  Where is the title set from outside of this factory?

src/charts/donut/donut-pct-chart-directive.js
```diff
@@ -0,0 +1,354 @@
+/**
+ * @ngdoc directive
+ * @name patternfly.charts.directive:pfDonutPctChart
+ *
+ * @description
+ *   Directive for rendering a percentage used donut/radial chart.  The Used Percentage fill starts at 12 o’clock and
+ *   moves clockwise.  Whatever portion of the donut not Used, will be represented as Available, and rendered as a
```
waldenraines: Typo

>  Whatever portion of the donut not Used

Should probably be "Whatever portion of the donut is not used..."

src/charts/donut/donut-pct-chart-directive.js
```diff
@@ -0,0 +1,354 @@
...
+     $scope.custData = {
+         'used': '670',
+         'total': '1000'
+     };
+
+   };
+ </file>
+ </example>
+ */
+angular.module('patternfly.charts').directive('pfDonutPctChart', ['ChartsMixin', '$timeout',
```
waldenraines: This is obviously outside the scope of this PR but we should look into using [grunt-ng-annotate](https://github.com/mzgol/grunt-ng-annotate) so we don't have to use this very verbose syntax ([`'dependency, function (dependency) { ...`.

See https://github.com/patternfly/angular-patternfly/issues/34

src/charts/donut/donut-pct-chart-directive.js
```diff
@@ -0,0 +1,354 @@
...
+  function(chartsMixin, $timeout) {
+    'use strict';
+    return {
+      restrict: 'A',
+      scope: {
+        id: '@',
+        config: '=',
+        data: '=',
```
waldenraines: I think it's best practice to use prefixed attributes (i.e. `chart-id`, `chart-data`) so as to avoid a potential collision with actual HTML elements.  If you want to use the html ID of the element you could use the `$attrs` parameter that is passed into the controller.

src/charts/donut/donut-pct-chart-directive.js
```diff
@@ -0,0 +1,354 @@
...
+            var color = {
+              pattern: []
+            };
+            var percentUsed = scope.data.used / scope.data.total * 100.0;
+            color.pattern[0] = $scope.getStatusColor(percentUsed, scope.config.thresholds);
+            color.pattern[1] = '#D1D1D1';
+            return color;
+          };
+
+          var donutTooltip = function(scope) {
```
waldenraines: It's best practice to declare the `var` at the top of the controller to avoid any potential hoisting issues.

src/charts/donut/donut-pct-chart-directive.js
```diff
@@ -0,0 +1,354 @@
...
+              return '#0088CE';
+            } else {
+              if (used >= thresholds.error) {
+                return '#CC0000';
+              } else if (used >= thresholds.warning) {
+                return '#EC7A08';
+              } else {
+                return '#0088CE';
+              }
+            }
```
waldenraines: This logic is very confusing to me because of the negative condition and the multiple return points.  I suggest reworking it like this (untested):

```javascript
$scope.getStatusColor = function(used, thresholds) {
    var color = '#0088CE';

    if (thresholds) {
        if (used >= thresholds.error) {
            color = '#CC0000';
        } else if (used >= thresholds.warning) {
            color = '#EC7A08';
        }
    }
 
    return color;
};
```

src/charts/donut/donut-pct-chart-directive.js
```diff
@@ -0,0 +1,354 @@
...
+          };
+
+          $scope.statusDonutColor = function(scope) {
+            var color = {
+              pattern: []
+            };
+            var percentUsed = scope.data.used / scope.data.total * 100.0;
```
waldenraines: Same note about declaring all `var`s at once at the top of the function.

waldenraines:
Do we have a coding standard on `var blah = function(variable){};` vs `var blah = function (variable) {};`?  I see both in this PR.

src/charts/donut/donut-pct-chart-directive.js
```diff
@@ -0,0 +1,354 @@
...
+                if (scope.config.tooltipFn) {
+                  var tootipHtml = '<span class="donut-tooltip-pf" style="white-space: nowrap;">' +
+                                        scope.config.tooltipFn(d) +
+                                   '</span>';
+                  return tootipHtml;
+                } else {
+                  return '<span class="donut-tooltip-pf" style="white-space: nowrap;">' +
+                            Math.round(d[0].ratio * 100) + '%' + ' ' + $scope.config.units + ' ' + d[0].name +
+                         '</span>';
+                }
```
waldenraines: I personally prefer this style, as I find it much more readable:

```javascript
var result = false;

if (condition) {
    result = true;
}

return result;
```

src/charts/donut/donut-pct-chart-directive.js
```diff
@@ -0,0 +1,354 @@
...
+            var percentUsed = scope.data.used / scope.data.total * 100.0;
+            color.pattern[0] = $scope.getStatusColor(percentUsed, scope.config.thresholds);
+            color.pattern[1] = '#D1D1D1';
+            return color;
+          };
+
+          var donutTooltip = function(scope) {
+            return {
+              contents: function (d) {
+                if (scope.config.tooltipFn) {
```
waldenraines: What should happen here if `scope.config.tooltipFn` is falsy?  Is it an error condition?

src/charts/donut/donut-pct-chart-directive.js
dtaylor113: Hi, what var are you referring too, color & percentUsed?
Are you suggesting that it should be something like:
$scope.statusDonutColor = function(scope) {
            var color, percentUsed;
...
...
}

src/charts/donut/donut-pct-chart-directive.js
```diff
@@ -0,0 +1,354 @@
...
+                }
+              }
+            };
+          };
+
+          $scope.getDonutData = function(scope) {
+            return {
+              columns: [
+                ['Used', scope.data.used],
+                ['Available', scope.data.available]
```
waldenraines: Again, the lack of translations is outside the scope of this PR but it made me think of this:  https://github.com/patternfly/angular-patternfly/issues/35

src/charts/donut/donut-pct-chart-directive.js
waldenraines: In this case I'm talking about `var donutTooltip = function(scope) {`.  I personally like having those at the top and the `$scope` functions at the bottom but regardless the `var`s should be declared at the top (since they will be hoisted there anyway).

src/charts/donut/donut-pct-chart-directive.js
waldenraines:
```javascript
$scope.statusDonutColor = function(scope) {
   var color, percentUsed;

   color = ....
};
```

or 

```javascript
$scope.statusDonutColor = function(scope) {
  var percentUsed = scope.data.used / scope.data.total * 100.0,
      color = {
        pattern: []
      };
};

Although the latter looks a little strange here.
```

src/charts/donut/donut-pct-chart-directive.js
dtaylor113: Hi, you wrote: "Do we have a coding standard on var blah = function(variable){}; vs var blah = function (variable) {};? I see both in this PR."
Honestly, it took me a few re-reads to figure out the difference between A vs B (a single space!?).   Are we really going to be doing code reviews to this level of detail?  If so, code reviews are going to take much longer to do, and I'm not sure holding developers to such ridged standards is productive, or really makes the code any better.
I agree with many of the other best practice issues you have brought up, but I think we need to come to an agreement on what best practices it makes sense to follow.
Thanks,
- Dave

src/charts/donut/donut-pct-chart-directive.js
```diff
@@ -0,0 +1,354 @@
...
+          $scope.updateAll = function(scope) {
+            $scope.updateAvailable();
+            $scope.config.data = $scope.getDonutData($scope);
+            $scope.config.color = $scope.statusDonutColor($scope);
+            $scope.config.tooltip = donutTooltip(scope);
+          };
+
+          $scope.config = $.extend(true, chartsMixin.getDefaultDonutConfig(), $scope.config);
+          $scope.updateAll($scope);
+        }],
```
waldenraines: Shouldn't the alignment of this be even with the `link` below?

dist/styles/angular-patternfly.css
dtaylor113: Yeah, sorry, that was a cut & paste.  I'll change it to 2.

src/charts/c3/c3.mixin.js
dtaylor113: I believe it should be a factory.  The version we have in our cfmePOC actually has all the defaults for all the C3 chart types, so the file is injected into donut, sparkline, utilization, etc.. charts.
I will change it's filename.
-Thanks

src/charts/c3/c3.mixin.js
dtaylor113: Sorry, I'm not understanding this request.  Could you please provide an example?
Thanks

src/charts/donut/donut-pct-chart-directive.js
waldenraines: I don't think we'll have to point these out in a code review once I turn on `white: true` in `.jshintrc` since the builds will fail instead :stuck_out_tongue_winking_eye:

waldenraines: > I don't think we'll have to point these out in a code review once I turn on white: true in .jshintrc since the builds will fail instead

See https://github.com/patternfly/angular-patternfly/pull/36

dtaylor113: We've been wanting to speak to you about the .jshintrc for the ridged indentation as well ;-)  
It makes commenting out code for testing really difficult.  I guess I just never worked on a project where there was such close attention to formatting.  I can understand best practices for naming conventions, var declarations, etc., maybe even indentation, but spaces in method signatures!?   IMO this is not an efficient use of our time and it really doesn't make the code any more readable.

src/charts/donut/donut-pct-chart-directive.js
dtaylor113: scope.config.tooltipFn is an optional property

styles/angular-patternfly.css
```diff
@@ -0,0 +1,29 @@
+.donut-chart-pf {
+	width: 100%;
```
waldenraines: Same thing here with the spacing.

waldenraines:
Missing tests for the mixin factory.

src/charts/c3/c3.mixin.js
waldenraines: > The version we have in our cfmePOC...

You'll have to excuse my ignorance but what is cfmePOC?

You can inject an [angular constant](https://docs.angularjs.org/api/auto/service/$provide#constant) as well.  To me it makes sense to make changing the title dynamic but all of these static config values make me think of a constant.  I'll have to take a deeper look at how this is used as I don't see where the title is set at first glance.

src/charts/c3/c3.mixin.js
waldenraines: Sorry, what I mean is this:

```javascript
angular.module('patternfly.charts').factory('ChartsMixin', function() {
  var getDefaultDonut, getDefaultDonutSize;

  getDefaultDonut = function () {...};
  getDefaultDonutSize = function () {...};

};
```

This is actually equivalent to what you have currently once it's parsed.  The trouble is that by not declaring your `var`s up top you can run into oddities like this

```javascript

var blah = 1;
function someFunction () {
	if (!blah) {
		var blah = 10;
	}
	console.log(blah);
}
someFunction();
```

This results in 10.  Now this is obviously a contrived example but it's a best practice to declare `var`s up top, outright to avoid any funny business (javascript is already funny enough).

src/charts/donut/donut-pct-chart-directive.js
waldenraines: :+1:

src/charts/donut/donut-pct-chart-directive.js
waldenraines: > We've been wanting to speak to you about the .jshintrc for the ridged indentation as well ;-)

Haha, I'm certainly open to the discussion.  I can see the points on both sides.

> It makes commenting out code for testing really difficult.

Can you elaborate on this please?  

> I guess I just never worked on a project where there was such close attention to formatting. I can understand best practices for naming conventions, var declarations, etc., maybe even indentation, but spaces in method signatures!? IMO this is not an efficient use of our time and it really doesn't make the code any more readable.

Ultimately, I suppose that's where we disagree.  I believe having consistency does help make code more readable.

Although we may want to consider taking this discussion to the mailing list?

jeff-phillips-18: Can we allow these discrepancies for this PR and fix them if/when https://github.com/patternfly/angular-patternfly/pull/36/ gets accepted? If we decide to enforce these rules, then these issues can be fixed at that time.

jeff-phillips-18:
LGTM

src/charts/c3/c3.mixin.js
dtaylor113: Sorry, the cfmePOC is a dashboard for containers where we initially developed a lot of the c3 charts (trends, utilization, donut, sparkline, etc.).  We are slowly moving these charts into a-pf#master.

src/charts/c3/c3.mixin.js
dtaylor113: Ok, that is a good reason.  Thanks for explaining.

src/charts/donut/donut-pct-chart-directive.js
dtaylor113: In regards to the indentation making debugging difficult.  I often want to comment out an outer function, such as a $timeout, in order to run the code immediately:

         // $timeout(function() {
            var donutChartTitle = element[0].querySelector('text.c3-chart-arcs-title');
            ...
         // }, 300);

I comment out the outer function, then 'grunt build', and of course the .jshintrc fails this, then have to go back, fix indentation, rerun, the do all over again when I restore the outer function.
I realize I can temporarily comment out the property in .jshintrc, but that is just another step to remember to restore before PR.

src/charts/donut/donut-pct-chart-directive.js
dtaylor113: Link is in alignment with templateUrl and controller properties above, which also indent their contents by 2 spaces. LGTM?

src/charts/donut/donut-pct-chart-directive.js
jeff-phillips-18: Dave, you can always use --force on the grunt commands which will continue execution regardless of jshint of test failures.

waldenraines: > Can we allow these discrepancies for this PR and fix them if/when #36 gets accepted? If we decide to enforce these rules, then these issues can be fixed at that time.

Definitely.  I don't want to hold up this PR on something as of yet undecided.

I'll start a thread on the mailing list about this because it's outside of the context of this particular PR and when this is merged I have no problem running my regexes again to fix the spacing for #36.

> ... comment out the outer function, then 'grunt build'...

What if we used the grunt live reload server like most other grunt-based projects?  Is there a reason why we are building every single time we need to test a change?

P.S.: Check out [my website's gruntfile](https://github.com/waldenraines/waldenrainesdotcom/blob/master/Gruntfile.js#L28-L61) for an example (based off of yeoman's [generator-angular](https://github.com/yeoman/generator-angular)), particularly the watch config.  It runs  jshint on javascript files when they are changed but that doesn't kill the webserver.  It also reloads the web browser on changes and runs tests on test file changes.

src/charts/donut/donut-pct-chart-directive.js
waldenraines: I'm talking about the `}]` above.  It's difficult to tell once it's been commented on but I would think the end of the controller block (`}]`) and the `link` property right below it (the controller and the link are in the same column) should also be in the same column.

[Look at the code](https://github.com/dtaylor113/angular-patternfly/blob/addDonutChart/src/charts/donut/donut-pct-chart-directive.js#L305-L306) before my comment, it's easier to see what I mean.

src/charts/donut/donut-pct-chart-directive.js
dtaylor113: Hi, I added: 

controller: ['$scope', '$attrs',
        function($scope, $attrs) {
          var donutTooltip;

          $scope.donutChartId = 'donutChart';
          if ($attrs.id) {
            console.log('$attrs.id'+$attrs.id);
            $scope.donutChartId = $attrs.id + $scope.donutChartId;

...but am now getting an error in the unit tests:

PhantomJS 1.9.8 (Linux) Directive: pfDonutPctChart should trigger no threshold FAILED
	Error: [$injector:unpr] Unknown provider: $attrsProvider <- $attrs
	http://errors.angularjs.org/1.3.17/$injector/unpr?p0=%24attrsProvider%20%3C-%20%24attrs

I temporarily removed the unit test, and the actual code works fine, but once I restore the unit test, it fails.  I'm googling the issue, but not sure what is causing this.

waldenraines: It's `controller: function($scope, $element, $attrs, $transclude, otherInjectables) { ... },`

https://docs.angularjs.org/api/ng/service/$compile

dtaylor113: I am going to move this back into a scope.config property which it was originally.  I was trying to reuse the 'id' attr, but it is causing too much issues.

dist/styles/angular-patternfly.css
dtaylor113: fixed

src/charts/c3/c3.mixin.js
dtaylor113: Fixed

src/charts/donut/donut-pct-chart-directive.js
dtaylor113: Fixed

src/charts/donut/donut-pct-chart-directive.js
dtaylor113: Fixed

src/charts/donut/donut-pct-chart-directive.js
dtaylor113: Fixed

src/charts/donut/donut-pct-chart-directive.js
dtaylor113: Fixed, thanks for the code snippet

src/charts/donut/donut-pct-chart-directive.js
dtaylor113: Fixed

src/charts/donut/donut-pct-chart-directive.js
dtaylor113: lol, I tried to align it:

 }],
 link: function(scope, element, attrs) {

But am now getting:

Linting src/charts/donut/donut-pct-chart-directive.js ...ERROR
[L312:C7] W015: Expected '}' to have an indentation at 9 instead at 7.
      }],

waldenraines: Hah.  It's because of the verbose style of dependency injection (i.e. the []).  It would probably be happy if you kept the } at column 9 and put the ] on the next line at column 7 (since that would match the function declaration above).

src/charts/c3/c3.mixin.js
dtaylor113: Hi, changed this to 'angular.module('patternfly.charts').constant('c3ChartDefaults', {...' and renamed file accordingly.'

dtaylor113:
Hi Walden,
I believe I have addressed all of your issues that are within the scope of this PR.
Thanks for all the feedback,
- Dave
p.s. I will squash all commits into a single one before merge

dist/styles/angular-patternfly.css
```diff
@@ -0,0 +1,29 @@
...
+  font-weight: normal;
+  color: #333;
+}
+
+.donut-title-small-pf {
+  font-family: "Open Sans", Helvetica, Arial, sans-serif;
+  font-size: 12px;
+  font-weight: normal;
+  color: #333;
+}
```
waldenraines: Shouldn't this be minified?

dtaylor113: There is a minified version in 'dist/styles/angular-patternfly.min.js'

waldenraines: Yeah, sorry, wrong file!

waldenraines:
LGTM, thanks!

styles/angular-patternfly.css
```diff
@@ -0,0 +1,29 @@
...
+  color: #fff;
+  opacity: 0.9;
+  filter: alpha(opacity=90);
+  padding: 2px 6px;
+}
+
+.donut-title-big-pf {
+  font-family: "Open Sans", Helvetica, Arial, sans-serif;
+  font-size: 22px;
+  font-weight: normal;
```
erundle: We need to use font-weight: 400 instead of normal. The way the fonts are currently declared in the font file are a bit weird and normal isn't mapped to anything.

What is the reason for needing the specify the font-family? That should be on the body and picked up by default.

erundle: The rest looks good so after you have looked at this let me know and I can merge.

dtaylor113: Ok, fixed those css issues. thanks

dist/styles/angular-patternfly.css
dtaylor113: Hi Walden, 
Silly question.  I'm doing a code review and I click on the blue + to add a comment to a piece of code; how did you get the snippet to appear here in the Conversations, above the comment? 
Thanks,
- Dave

src/charts/donut/donut-pct-chart-directive.js
dtaylor113: Hi, irt. 'What if we used the grunt live reload server like most other grunt-based projects? Is there a reason why we are building every single time we need to test a change?"  
When I'm working on a directive, I use the ng-docs 'dist/docs' to view changes, even style changes.   Don't we need to do a full build in order to get a fully updated 'dist/docs'?

dtaylor113: irt. "P.S.: Check out my website's gruntfile" 
"...watch config. It runs jshint on javascript files when they are changed but that doesn't kill the webserver."
-- awsome! that's what I want in angular-patternfly

"It also reloads the web browser on changes, "
-- think a-pf already does this

"runs tests on test file changes, "
-- this would be nice to have

"and automatically updates asset includes when they are added/removed from the bower.json file using grunt-wiredep."
-- this is nice, but probably a separate discussion.

So, should I change the watch config to not stop on jshint errors and also rerun karma tests upon change, create a PR.  Or do you usually run the intent by the team first before coding?

Thanks,
- Dave

waldenraines: I'm actually working on updating the grunt file right now.  You're right that usemin and wiredep are pretty useless for this project since it's not a typical webapp (with an index.html, etc.).

PR #38, opened 2015-08-20
Add Sparkline Chart Directive

jeff-phillips-18:
Adding directive for sparkline charts. Includes unit tests and ng-docs.


src/charts/sparkline/sparkline-chart.directive.js
```diff
@@ -0,0 +1,327 @@
...
+ * @name patternfly.charts.directive:pfSparklineChart
+ *
+ * @description
+ *   Directive for rendering a sparkline chart.
+ *   <br><br>
+ *   See http://c3js.org/reference.html for a full list of C3 chart options.
+ *
+ * @param {object} config configuration settings for the sparkline chart:<br/>
+ * <ul style='list-style-type: none'>
+ * <li>.chartId            - the ID iof the container that the chart should bind to
```
waldenraines: Typo: "iof"

src/charts/sparkline/sparkline-chart.directive.js
```diff
@@ -0,0 +1,327 @@
...
+  function (c3ChartDefaults) {
+    'use strict';
+    return {
+      restrict: 'A',
+      scope: {
+        config: '=',
+        chartData: '=',
+        chartHeight: '@',
+        showXAxis: '@',
+        showYAxis: '@'
```
waldenraines: Shouldn't these use `=` instead of `@`?  This is why you have to use ` $scope.showXAxis === 'true'` below.

src/charts/sparkline/sparkline-chart.directive.js
```diff
@@ -0,0 +1,327 @@
...
+                    '  <table class="c3-tooltip">' +
+                    '    <tbody>' +
+                    '      <tr>' +
+                    '        <td class="name">' + percentUsed + '%' + '</td>' +
+                    '      </tr>' +
+                    '    </tbody>' +
+                    '  </table>' +
+                    '</div>';
+                }
+                else {
```
waldenraines: What about using a switch statement here instead of all the else ifs?

src/charts/sparkline/sparkline-chart.directive.js
```diff
@@ -0,0 +1,327 @@
...
+                    '      </tr>' +
+                    '    </tbody>' +
+                    '  </table>' +
+                    '</div>';
+                }
+                else if (scope.config.tooltipType === 'valuePerDay') {
+                  tip = '<div id="sparkline-tooltip" class="module-triangle-bottom">' +
+                    '  <table class="c3-tooltip">' +
+                    '    <tbody>' +
+                    '      <tr>' +
```
waldenraines: It looks like there is a lot of of redundant html structure in these different tooltip types and plus the HTML in the javascript makes this difficult to read.  What about creating a chart-tooltip directive that this directive can reference?

dist/angular-patternfly.js
```diff
@@ -635,6 +707,333 @@ angular.module('patternfly.charts').directive('pfDonutPctChart', ['c3ChartDefaul
...
+              x: chartData.xData[0],
+              columns: [
+                chartData.xData,
+                chartData.data
+              ],
+              type: 'area'
+            };
+          };
+
+          $scope.sparklineTooltip = function (scope) {
```
waldenraines: Seems very strange to me for scope to be passed in as an argument here.  What is the purpose of this?

src/charts/sparkline/sparkline-chart.directive.js
dtaylor113: Hi Walden,
How do you get the code snippet to appear in the Conversation here?  I goto the commit, click on the blue +, but I don't get code snippets to appear here
Thanks,
- Dave

src/charts/sparkline/sparkline-chart.directive.js
```diff
@@ -0,0 +1,327 @@
...
+                  var graphOffsetX = document.querySelector('#' + scope.sparklineChartId + ' g.c3-axis-y').getBoundingClientRect().right;
+                  var x = Math.max(0, center + graphOffsetX - chartBox.left - Math.floor(width / 2));
+
+                  return {
+                    top: top - height,
+                    left: Math.min(x, chartBox.width - width)
+                  };
+                }
+                catch (TypeError) {
+                }
```
waldenraines: We want to do nothing with this error (just making sure)?

src/charts/sparkline/sparkline-chart.directive.js
```diff
@@ -0,0 +1,327 @@
...
+           * Setup Axis options. Default is to not show either axis. This can be overridden in two ways:
+           *   1) in the config, setting showAxis to true will show both axis
+           *   2) in the attributes showXAxis and showYAxis will override the config if set
+           *
+           * By default only line and the tick marks are shown, no labels. This is a sparkline and should be used
+           * only to show a brief idea of trending. This cna be overridden by setting the config.axis options per C3
+           */
+
+          if ($scope.showXAxis === undefined)
+          {
```
waldenraines: This is strange syntax with the `{` on a new line by itself.

src/charts/sparkline/sparkline-chart.directive.js
```diff
@@ -0,0 +1,327 @@
...
+                }
+                catch (TypeError) {
+                }
+              }
+            };
+          };
+
+          /*
+           * Setup Axis options. Default is to not show either axis. This can be overridden in two ways:
+           *   1) in the config, setting showAxis to true will show both axis
```
waldenraines: "...both axes".

src/charts/sparkline/sparkline-chart.directive.js
```diff
@@ -0,0 +1,327 @@
...
+           *   2) in the attributes showXAxis and showYAxis will override the config if set
+           *
+           * By default only line and the tick marks are shown, no labels. This is a sparkline and should be used
+           * only to show a brief idea of trending. This cna be overridden by setting the config.axis options per C3
```
waldenraines: Typo: "cna".

src/charts/sparkline/sparkline-chart.directive.js
waldenraines: Yeah, that's all I do.  Go to the files tab and click the blue + on the line you want to comment on.

src/charts/sparkline/sparkline-chart.directive.js
```diff
@@ -0,0 +1,327 @@
...
+                }
+                return tip;
+              },
+              position: function (data, width, height, element) {
+                try {
+                  var center = parseInt(element.getAttribute('x'));
+                  var top = parseInt(element.getAttribute('y'));
+                  var chartBox = document.querySelector('#' + scope.sparklineChartId).getBoundingClientRect();
+                  var graphOffsetX = document.querySelector('#' + scope.sparklineChartId + ' g.c3-axis-y').getBoundingClientRect().right;
+                  var x = Math.max(0, center + graphOffsetX - chartBox.left - Math.floor(width / 2));
```
waldenraines: I personally prefer a single var (and that's what we've been doing elsewhere) although I am open to debating which is best.

src/charts/sparkline/sparkline-chart.directive.js
```diff
@@ -0,0 +1,327 @@
...
+
+          if ($scope.showYAxis === undefined)
+          {
+            $scope.showYAxis = ($scope.config.showAxis !== undefined) && $scope.config.showAxis;
+          }
+
+          $scope.defaultConfig = c3ChartDefaults.getDefaultSparklineConfig();
+          $scope.defaultConfig.axis = {
+            x: {
+              show: $scope.showXAxis === true || $scope.showXAxis === 'true',
```
waldenraines: See above note about '=' vs '@'.

dist/angular-patternfly.js
```diff
@@ -172,7 +172,79 @@ angular.module('patternfly.charts').constant('c3ChartDefaults', {
...
+      zerobased: true
+    };
+  },
+
+  getDefaultSparklineSize: function () {
+    return {
+      height: 60
+    };
+  },
+
```
dtaylor113: Ignore this, just a test comment.

src/charts/sparkline/sparkline-chart.directive.js
dtaylor113: Ok, I was clicking on Commit, then commenting there, works correctly under 'Files Changed'

src/charts/sparkline/sparkline-chart.directive.js
jeff-phillips-18: Been trying to use a directive here but to no avail. I will, however, create a utility method to put the common html in one place.

src/charts/sparkline/sparkline-chart.directive.js
jeff-phillips-18: Personally, I find it harder to read when all the vars are declared on one line. Almost seems like each var is not intentional. I also like to init on declaration when possible. Again, its a preference.

src/charts/sparkline/sparkline-chart.directive.js
jeff-phillips-18: Fixed

src/charts/sparkline/sparkline-chart.directive.js
jeff-phillips-18: Fixed, thanks

src/charts/sparkline/sparkline-chart.directive.js
jeff-phillips-18: Fixed

dist/angular-patternfly.js
jeff-phillips-18: Yeah, I agree. Don't know what that was done. Fixed.

src/charts/sparkline/sparkline-chart.directive.js
jeff-phillips-18: not returning a value puts the tooltip at 0,0 which is probably what I would return anyhow. I could add that if you feel its more robust/readable

src/charts/sparkline/sparkline-chart.directive.js
jeff-phillips-18: Fixed

src/charts/sparkline/sparkline-chart.directive.js
jeff-phillips-18: Fixed

src/charts/sparkline/sparkline-chart.directive.js
jeff-phillips-18: Fixed

src/charts/sparkline/sparkline-chart.directive.js
jeff-phillips-18: Fixed.

jeff-phillips-18:
Dave, fixed your comments as well.

src/charts/sparkline/sparkline-chart.directive.js
waldenraines: Yeah, I can see that.  It's probably more of a habitual thing for me since we do single line for `var`s in katello.  But I am fine with both styles as long as the `var`s are at the top.

src/charts/c3/c3-chart-defaults.js
```diff
@@ -37,5 +37,90 @@ angular.module('patternfly.charts').constant('c3ChartDefaults', {
...
+   * Tooltips
+   */
+  getTooltipTableHTML: function (tipRows) {
+    return '<div class="module-triangle-bottom">' +
+    '  <table class="c3-tooltip">' +
+    '    <tbody>' +
+           tipRows +
+    '    </tbody>' +
+    '  </table>' +
+    '</div>';
```
waldenraines: I personally think this should stay in the directive as that is really the only place you should have HTML (other than .html files) in Angular.

src/charts/sparkline/sparkline-chart.directive.js
```diff
@@ -0,0 +1,311 @@
...
+              tick: {
+                format: function () {
+                  return '';
+                }
+              }
+            }
+          };
+
+          // Setup the default configuration
+          $scope.defaultConfig.tooltip = $scope.sparklineTooltip($scope);
```
waldenraines: `$scope` isn't used in the function.

src/charts/sparkline/sparkline-chart.directive.js
waldenraines: It just seems odd to use an exception to handle this.  Would it be possible to do this check up front in a conditional?

src/charts/sparkline/sparkline-chart.directive.js
jeff-phillips-18: removed it

src/charts/c3/c3-chart-defaults.js
jeff-phillips-18: Moved the function into the directive

jeff-phillips-18:
I will squash the commits when you agree the issues are all resolved

waldenraines:
> I will squash the commits when you agree the issues are all resolved

I'm good with your last push.

PR #39, opened 2015-08-20
Use eslint to enforce coding standards.

waldenraines:
Enforce a series of checks in order to maintain readability of code.
Now before opening a pull request one can run `grunt check` and all
linting and testing will be executed.


dtaylor113:
Hi, I've looked through Gruntfile.js, but I'm still learning all of it's usages.
Can you please elaborate on this process?  So, was jshint checking removed from the build process, and should run 'grunt check' before creating a PR?
Thanks,
- Dave

waldenraines:
> So, was jshint checking removed from the build process, and it will only be run manually via 'grunt check' before creating a PR?

The linting will also be ran as part of the build process as can be seen in the [Travis CI job](https://travis-ci.org/patternfly/angular-patternfly/builds/76526029#L384).

I just added a `grunt check` for convenience so you don't have to run `grunt lint; grunt test`.

ehelms:
Why keep the .jshintc file? What about adding the angular eslint plugin? Would that be another PR on top of this change?

Overall +1 for this change.

waldenraines:
> Why keep the .jshintc file?

Will remove.

> What about adding the angular eslint plugin? Would that be another PR on top of this change?

Yeah, I will enable that in the future.  I wanted to first enable ng-annotate (and some other grunt goodies) rather than having to adjust the `ng_di` check twice.

jeff-phillips-18:
LGTM

waldenraines:
@erundle @kenfinnigan any objections to merging this?

erundle:
I am for the overall change but I think we may want to relax some of the coding standards since some of them are preference.

The only issue that I see is I believe "use strict" should be defined within a function not at the top of the file. Since all of the src files get mashed into 1 the "use strict" statements get sprinkled all over the dist. This could later be problematic if someone takes the dist/angular-patternfly.js and concats it in a build process with another library that doesn't work with "use strict" since it will break that code. So for the "good of others" I try to follow having it in the function as a best practice.

waldenraines:
> I am for the overall change but I think we may want to relax some of the coding standards since some of them are preference.

Which ones specifically?



> The only issue that I see is I believe "use strict" should be defined within a function not at the top of the file. Since all of the src files get mashed into 1 the "use strict" statements get sprinkled all over the dist. This could later be problematic if someone takes the dist/angular-patternfly.js and concats it in a build process with another library that doesn't work with "use strict" since it will break that code. So for the "good of others" I try to follow having it in the function as a best practice.

That's a really good point.  I will update this PR to use the function form of `'use strict';`.

waldenraines:
> I will update this PR to use the function form of 'use strict';.

Updated

erundle:
Thanks. Let's move ahead with this and if we see that some of the standards become too cumbersome maybe we can revist. I will go ahead and merge

PR #40, opened 2015-08-22
Updating patternfly reference impl to be a devDependency

erundle:
Angular-Patternfly is now being used in a ruby based project where they use they use the SASS version (https://github.com/patternfly/patternfly-sass) of the reference impl not the LESS version. 

Listing the LESS version as a dependency causes the library to automatically get pulled into their code base which means therubyracer to gets pulled in causing their code base to be slow. They would like to exclude the less version from getting pulled in so I am suggesting changing the patternfly reference impl to be a devDependency instead and then the projects can explicitly specify whether they want the LESS or SASS version of PF.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -32,15 +32,14 @@
     "url": "git://github.com/patternfly/angular-patternfly.git"
   },
   "dependencies": {
-    "angular": "1.2.25 - 1.3.*",
-    "patternfly": "~2.0.0",
-    "c3": "~0.4.10"
+    "angular": "1.2.25 - 1.3.*"
   },
   "devDependencies": {
     "angular-animate": "1.2.25 - 1.3.*",
     "angular-mocks": "1.2.25 - 1.3.*",
     "angular-sanitize": "1.2.25 - 1.3.*",
     "angular-touch": "1.2.25 - 1.3.*",
-    "angular-route": "1.2.25 - 1.3.*"
+    "angular-route": "1.2.25 - 1.3.*",
+    "patternfly": "~2.0.0"
   }
 }


dtaylor113:
LGTM

jeff-phillips-18:
LGTM

bower.json
```diff
@@ -32,15 +32,14 @@
     "url": "git://github.com/patternfly/angular-patternfly.git"
   },
   "dependencies": {
-    "angular": "1.2.25 - 1.3.*",
-    "patternfly": "~2.0.0",
-    "c3": "~0.4.10"
```
waldenraines: c3 isn't needed as a dependency?  Thought it was used in that chart stuff?

erundle: It is but it comes in with base patternfly so I don't think we want it to get pulled twice.

erundle: As an FYI there is an effort going on to make the patternfly reference dependencies true bower dependencies so we might need to update this at a later date but knowing what we know right now I think this is the way to go.

waldenraines: > It is but it comes in with base patternfly so I don't think we want it to get pulled twice.

It won't be downloaded twice, bower handles that.  What happens if patternfly ever drops the dependency of c3?  It's unlikely to happen, especially without warning, but I think it's a good practice to include as dependencies libraries that you are using directly.

erundle: Since base patternfly isn't set up to use bower properly it will come down twice. They are checking in the libraries directly to the components directory. So we will get it once from the components directory and a second time from this bower dependency include. We don't have dependencies for the other libraries (datepicker) so I would suggest we add all of them or just let PF reference handle it.

waldenraines: Ahh, I see, thanks for the explaination.

waldenraines:
LGTM

skateman:
:+1: :+1: :tada:

PR #43, opened 2015-08-25
Upgrade to Angular 1.3 Only

erundle:
Changing the project to support angular 1.3+. This required updating ngDocs and fixing the example code since angular did not like the previous "function" definition. 

I made this change since we are looking in the future to add angular-bootstrap 1.3 to this project and angular 1.3 is a required. 

Side Note: I also fixed a previous existing bug with warning/danger icons not showing up for notifications. This was discovered during the test of this update.


dist/angular-patternfly.js
```diff
@@ -302,7 +304,7 @@ angular.module('patternfly.card').directive('pfCard', function () {
       $scope.updateAvailable();
       $scope.chartConfig.data.columns = [["Used",$scope.used],["Available",$scope.available]];
    };
- }
+  }]);
```
waldenraines: Nitpick:  these closing brackets and the parenthesis don't line up with the `angular.module` above.  This is also visible in the docs examples.

dtaylor113:
LGTM

waldenraines:
Looks good to me minus the aforementioned nitpick.

jeff-phillips-18:
+1

dist/angular-patternfly.js
erundle: fixed

waldenraines:
LGTM :+1:

PR #46, opened 2015-08-26
Fixes #34: add ng-annotate to build process.

### LINKED ISSUE #34: Use ng-annotate
We should be using [ng-annotate](https://github.com/olov/ng-annotate) via [grunt-ng-annotate](https://github.com/mzgol/grunt-ng-annotate) to avoid verbose function declarations when using dependency injection.

It turns this:

```javascript
angular.module("MyMod").controller("MyCtrl", function($scope, $timeout) {
});
```

into this

```javascript
angular.module("MyMod").controller("MyCtrl", ["$scope", "$timeout", function($scope, $timeout) {
}]);
```
automatically at build time so minificaiton is safe.

waldenraines:
Using ng-annotate in the build process so the verbose, array style,
function declarations are no longer needed.  Instead they are added
automatically during the build process.

Also fix the tabbing on some of the examples in the comments.


erundle:
I've ran into issues before with ng-annotate not working with controllers embedded inside a directive. Any idea it that is still a limitation? Otherwise it LGTM.

waldenraines:
> I've ran into issues before with ng-annotate not working with controllers embedded inside a directive. Any idea it that is still a limitation?

Are you talking about this https://github.com/olov/ng-annotate/issues/62?  If so I can upgrade the version of ng-annotate we are using in order to address this issue.

erundle:
Yes please upgrade it. I have found that it leads to much confusion when that one use case doesn't work.

waldenraines:
@erundle updated.

waldenraines:
Are we good to merge this?

PR #47, opened 2015-08-26
Adding AggStatus Card

erundle:
This adds a directive for the patternfly reference impl aggregate status card. An alternate layout is also provided to meet other project needs.

Note this PR increases the devDependency for PF to 2.1 since that is the version that the reference impl css lives in.


jeff-phillips-18:
LGTM

dist/styles/angular-patternfly.css
```diff
@@ -24,9 +52,6 @@
   color: #333;
 }
 
-.utilization-chart-pf {
-}
-
```
waldenraines: :+1:

src/card/aggregate-status/aggregate-status-card.directive.js
```diff
@@ -0,0 +1,93 @@
...
+ */
+
+angular.module( 'patternfly.card' ).directive('pfAggregateStatusCard', function () {
+  'use strict';
+  return {
+    restrict: 'A',
+    scope: {
+      status: '=',
+      showTopBorder: '@?',
+      altLayout: '@?'
```
waldenraines: If you use `=?` here you won't have to check against string true below.  Plus that would allow someone to use a variable for these properties.

src/card/aggregate-status/aggregate-status-card.html
```diff
@@ -0,0 +1,26 @@
+<div class="card-pf card-pf-aggregate-status" ng-class="{'card-pf-accented': showTopBorder === 'true', 'card-pf-aggregate-status-alt': altLayout==='true'}">
```
waldenraines: See note above about `true` vs `'true'`.

src/card/aggregate-status/aggregate-status-card.directive.js
erundle: We aren't intending that they should be able to change them after load. It is more of a UI layout decision based so for right now I don't think we need to support the variable cause and I would like to keep double binding to a minimum. If we decide that we need to support more of a soft UI case where the data dictates the page layout that might change but right now this isn't a requirement for us.

waldenraines: Ah I see.  What about converting the values to boolean in the directive itself with a comment?  To me, having the string comparisons in the template is not as straightforward and could even be mistaken as unintended.

erundle: Fixed. Converted to boolean in the directive link function.

waldenraines:
LGTM pending rebase.

PR #50, opened 2015-08-27
Minor styling updates to utilization charts per UXD review

jeff-phillips-18:


waldenraines:
> Also re-adds dist/angular-patternfly.js

I know this sounds silly but can we do this as a separate PR?  I'm happy to open one to re-add this file.

waldenraines:
> I know this sounds silly but can we do this as a separate PR? I'm happy to open one to re-add this file.

Actually see https://github.com/patternfly/angular-patternfly/pull/52

jeff-phillips-18:
I will resubmit

PR #51, opened 2015-08-27
Update version number to 2.1.0

erundle:


erundle:
Somehow it looks like we lost the dist files on my last rebase so I added those back. Master is basically broken until this goes in.

PR #54, opened 2015-08-28
Minor styling updates to utilization charts per UXD review

jeff-phillips-18:


src/charts/utilization/utilization-chart.html
```diff
@@ -1,7 +1,7 @@
 <div class="utilization-chart-pf">
-    <span class="title">{{config.title}}</span>
+    <span class="h3">{{config.title}}</span>
```
waldenraines: Shouldn't we just use a `<h1>` here?

src/charts/utilization/utilization-chart.html
```diff
@@ -1,7 +1,7 @@
...
     <div class="current-values">
-        <div class="available-count pull-left">
+        <div class="available-count h1 pull-left">
```
waldenraines: Shouldn't we just use an `<h1>` here?

src/charts/utilization/utilization-chart.html
jeff-phillips-18: Doing so would cause the .content h3 style setting to take over and change the line-height and font-size

src/charts/utilization/utilization-chart.html
jeff-phillips-18: Doing so would cause the .content h1 style setting to take over and change the line-height and font-size

src/charts/utilization/utilization-chart.html
waldenraines: It's arguable that's a bug in patternfly.  I would expect `.content .h3` to be equivalent to `.content h3`.  What if we override the font-size and line-height properties using `.utilization-chart-pf`?

jeff-phillips-18: Yeah, I thought of doing that, but I wanted to inherit from patternfly styles so if h3 changes in patternfly then this changes as well.

waldenraines: Right, but I feel like we're relying on a bug in patternfly to get the desired styling.  What will happen if they fix `.content .h3` to be the same as `.content h3`.  To me it makes more sense to explicitly override.

erundle: I think that is an ngDocs thing ".content" not patternfly so we need to stop that from affecting the examples since it is messing with our styles. But I agree with @waldenraines we either need to use the h3 or a class name that is relevent to the content not the tag.

jeff-phillips-18: I made the suggested change and I have added ngDoc only overrides for the .contents h# specifications to make them consistent with PF to give an accurate look in the ngDocs. I added one spec override for the font-weight for the utilization card h3 title as it varies from PF h3.

misc/bootstrap-overrides.css
```diff
@@ -5,3 +5,47 @@
 .form-horizontal .col-sm-2.control-label {
```
erundle: Umm so I know I named this originally but can we change this to ngdocs-overrides.css

erundle: Minus the file rename this LGTM

jeff-phillips-18: done

erundle: @jeff-phillips-18 thanks!! Should make things a bit clearer to why that file exists if someone sees it in the web inspector

waldenraines:
LGTM

Issue #49, opened 2015-08-27
Add Unit Tests for pfTranscludeDirective


jeff-phillips-18:
We need to add unit tests for the pfTransclude directive

PR #55, opened 2015-08-28
Add Unit Tests for Transclude Directive

jeff-phillips-18:
Fix for issue: https://github.com/patternfly/angular-patternfly/issues/49

### DIFF
--- a/test/utils/specUtils.js
+++ b/test/utils/specUtils.js
@@ -8,3 +8,4 @@ function eventFire(el, etype){
     el.dispatchEvent(evObj);
   }
 }
+
--- a/test/utils/transclude-directive.spec.js
+++ b/test/utils/transclude-directive.spec.js
@@ -1,0 +1,108 @@
+angular.module('patternfly.pfTranscludeTestHelper',['patternfly.utils'])
+  .controller( 'TestUtilCtrl', function($scope) {
+  })
+  .config(function($provide){
+    $provide.decorator('ngTranscludeDirective', ['$delegate', function($delegate) {
+      // Remove the original directive
+      $delegate.shift();
+      return $delegate;
+    }]);
+  })
+  .directive( 'transcludeSibling', function() {
+    return {
+      restrict: 'EAC',
+      transclude: true,
+      scope: {},
+      template:
+      '<div>' +
+      '  <input id="directiveId" value="{{$id}}">' +
+      '  <div pf-transclude></div>' +
+      '</div>'
+    }
+  })
+  .directive( 'transcludeParent', function() {
+    return {
+      restrict: 'EAC',
+      transclude: true,
+      scope: {},
+      template:
+      '<div>' +
+      '  <input id="directiveId" value="{{$id}}">' +
+      '  <div pf-transclude="parent"></div>' +
+      '</div>'
+    }
+  })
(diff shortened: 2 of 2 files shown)


test/utils/transclude-directive.spec.js
```diff
@@ -0,0 +1,61 @@
...
+    var markup = '<div ng-controller="TestUtilCtrl" >' +
+      '  <input id="outerId" value="{{$id}}">{{$id}}</input>' +
+      '  <div ' + transludeDirective + ' class="pf-transclude-example">' +
+      '    <input id="childId" value="{{$id}}">' +
+      '    <input id="parentId" value="{{$parent.$id}}">' +
+      '  </div>' +
+      '</div';
+    var element = $compile(angular.element(markup))(scope);
+    scope.$apply();
+    scope.$digest();
```
erundle: Why apply and digest (apply should call digest but digest is generally used in tests)

test/utils/transclude-directive.spec.js
```diff
@@ -0,0 +1,61 @@
+describe('Directive:  pfTransclude', function () {
+  var $scope;
+  var $compile;
+  var outerId;
+  var directiveId;
+  var childId;
+  var parentId;
```
erundle: Should these be comma separated instead for consistency?

The rest LGTM

jeff-phillips-18: Its a preference thing and I have a strong preference for not declaring multiple variables on a single line.

test/utils/transclude-directive.spec.js
jeff-phillips-18: Left over from trying to get things digested and working correctly. Removed the $apply call.

test/utils/transclude-directive.spec.js
erundle: OK. I am fine with this.

waldenraines: > Its a preference thing and I have a strong preference for not declaring multiple variables on a single line.

I tend to also prefer this style after using both.  I used to be in the "comma" camp but I now tend like this way better.

test/utils/specUtils.js
```diff
@@ -8,3 +8,51 @@ function eventFire(el, etype){
     el.dispatchEvent(evObj);
   }
 }
+
+angular.module('patternfly.testutils',['patternfly.utils'])
```
waldenraines: Wondering about the placement of these here versus in the test itself.  Will these ever be used outside of the context of the transclude directive test?

jeff-phillips-18: I found it a nice way to keep it separated and easy to be pulled in.  I could have added it to the spec file, but they aren't really test specs.

jeff-phillips-18: On second thought, I guess I agree... I will move it.

waldenraines: What about putting this into a different file within the utils directory?  That way it's contained since this file seems to be for common test utils.

waldenraines: Either way, thanks :)

jeff-phillips-18:
Fixed the last issue and squashed.

waldenraines:
@jeff-phillips-18 thanks!

PR #56, opened 2015-09-01
Donut Chart: Fixed ngdoc example and added more unit tests

dtaylor113:


src/charts/donut/donut-pct-chart-directive.js
```diff
@@ -302,6 +310,29 @@ angular.module('patternfly.charts').directive('pfDonutPctChart', function (c3Cha
           };
         };
 
+        $scope.getCenterLabelText = function (scope) {
```
jeff-phillips-18: No need to pass scope in here, use $scope inside the function

src/charts/donut/donut-pct-chart-directive.js
```diff
@@ -302,6 +310,29 @@ angular.module('patternfly.charts').directive('pfDonutPctChart', function (c3Cha
...
+          var centerLabelText;
+
+          // default to 'used' info.
+          centerLabelText = { bigText: scope.data.used,
+                              smText:  scope.config.units + ' Used' };
+
+          if (scope.config.centerLabelFn) {
+            centerLabelText.bigText = scope.config.centerLabelFn(scope);
```
jeff-phillips-18: Why would the centerLabelFn function need this scope?

src/charts/donut/donut-pct-chart-directive.js
```diff
@@ -316,34 +347,25 @@ angular.module('patternfly.charts').directive('pfDonutPctChart', function (c3Cha
...
-          } else if (scope.centerLabel === 'none') {
-            donutChartTitle.innerHTML = '';
+          if (!donutChartTitle) {
+            return;
+          }
+
+          centerLabelText = scope.getCenterLabelText(scope);
+
+          if (scope.config.centerLabelFn || scope.centerLabel === 'none') {
+            donutChartTitle.innerHTML = centerLabelText.bigText;
```
jeff-phillips-18: Seems like duplication of effort a bit. You could set innterHTML to centerLabelFn() and not even call getCenterLabelText. If 'none'  then using '' for both should be fine.

src/charts/donut/donut-pct-chart-directive.js
dtaylor113: So user can display scope.available, scope.total, scope.used, etc...

src/charts/donut/donut-pct-chart-directive.js
dtaylor113: I want 'getCenterLabelText' to return all possible label values (for unit testing).  If using a function or 'none', then we set the entire innerHTML to whatever centerLabelText.bigText has been set to; any other label format will divide the data into bigText and smText, to be used in the HTML spans specified.

src/charts/donut/donut-pct-chart-directive.js
jeff-phillips-18: that would mean we could never change things in this scope for fear of how it could be used outside. The total, used, and available (calculated at times) are all set on the data object which is passed in along with the function to call, therefore that data should already be accessible to them.

src/charts/donut/donut-pct-chart-directive.js
dtaylor113: Fixed

src/charts/donut/donut-pct-chart-directive.js
jeff-phillips-18: That makes both the getCenterLabelText function and the code outside of it know how to handle the Fn return value and how to handle none. Setting to None means the getCenterLabelText function MUST return "" for bigText and outside must set innerHTML to bigText not put it within the tspan. Harder to read and harder to be sure none is being handled correctly. When using the Fn, smallText is still set to scope.config.units + ' Used' so the outside needs to know not to use smallText.

src/charts/donut/donut-pct-chart-directive.js
dtaylor113: Ok, forgot about the two-way binding. Fixed

src/charts/donut/donut-pct-chart-directive.js
dtaylor113: If I move centerLabelFn outside of getCenterLabelText, and into just the link function, I can't unit test it

jeff-phillips-18: I'm not sure the ability to unit test makes it valid

dtaylor113: Ok, fixed the logic, made it easier to understand.  I'm ok with massaging the code in order to get more code coverage.

jeff-phillips-18:
LGTM

dist/angular-patternfly.js
```diff
@@ -611,6 +613,8 @@ angular.module('patternfly.card').directive('pfCard', function () {
             'total': '1000'
         };
 
+       $scope.availLabel = "available";
```
erundle: spelling

src/charts/donut/donut-pct-chart-directive.js
```diff
@@ -96,19 +96,19 @@
...
          center-label =<br> 'used'
        </div>
        <div class="col-md-3">
-         <div pf-donut-pct-chart config="availConfig" data="availData" center-label="available"></div>
+         <div pf-donut-pct-chart config="availConfig" data="availData" center-label="availLabel"></div>
          center-label =<br> 'available'
        </div>
        <div class="col-md-3">
-         <div pf-donut-pct-chart config="pctConfig" data="pctData" center-label="percent"></div>
+         <div pf-donut-pct-chart config="pctConfig" data="pctData" center-label="pctLabel"></div>
```
erundle: maybe spell this out percentageLabel?

erundle: Otherwise LGTM

dist/angular-patternfly.js
dtaylor113: lol..'availLabel' is actually correct, consistent with 'usedLabel', 'pctLabel', and 'noLabel'.

PR #57, opened 2015-09-02
Fix #37 - Adding support for angular 1.4.x

### LINKED ISSUE #37: Any Plans to support semver Angular 1.4.x?

erundle:
Adding support for angular 1.4.x and also changing angular to just be a devDependency since products seem the be specifying their own version anyway and having the listed dependency is causing problems with projects who are trying to consume angular-patternfly via rails-assets since they can't override the version like you can do with bower resolutions.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -31,15 +31,13 @@
     "type": "git",
     "url": "git://github.com/patternfly/angular-patternfly.git"
   },
-  "dependencies": {
-    "angular": "1.3.*"
-  },
   "devDependencies": {
-    "angular-animate": "1.3.*",
-    "angular-mocks": "1.3.*",
-    "angular-sanitize": "1.3.*",
-    "angular-touch": "1.3.*",
-    "angular-route": "1.3.*",
+    "angular": "1.3.0 - 1.4.*",
+    "angular-animate": "1.3.0 - 1.4.*",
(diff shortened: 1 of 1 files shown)


waldenraines:
I get the following when trying to `bower install`:

```
Unable to find a suitable version for angular, please choose one:
    1) angular#1.3.18 which resolved to 1.3.18 and is required by angular-animate#1.3.18
    2) angular#1.4.5 which resolved to 1.4.5 and is required by angular-animate#1.4.5, angular-mocks#1.4.5, angular-route#1.4.5, angular-sanitize#1.4.5, angular-touch#1.4.5
    3) angular#1.3.0 - 1.4.* which resolved to 1.4.5 and is required by angular-patternfly
```

Do we want to add a [resolutions](https://github.com/bower/bower.json-spec#resolutions) section to `bower.json`?

Note that this only happens when `bower install`ing on an existing set of dependencies (i.e. an upgrade).

erundle:
To me the resolutions section would belong in the project that is making the request not in angular-patternfly. Atleast that is how I have done it in the past. This is one of the reasons I am trying to remove away from listing angular as an actual dependency, so projects aren't required to have the resolution section.

dtaylor113:
LGTM

waldenraines:
@erundle makes sense.

PR #59, opened 2015-09-02
Do not allow both click selection and checkbox selection in same data…

jeff-phillips-18:
… list


test/views/datalist/data-list.spec.js
```diff
@@ -179,6 +182,10 @@ describe('Directive:  pfDataList', function () {
       return item.uuid === '2';
     };
 
+    // allow item selection
+    $scope.listConfig.selectItems = true;
+    $scope.listConfig.showSelectBox = false;
+
     $scope.listConfig.checkDisabled = checkDisabled;
```
dtaylor113: Would it be good to test an error case and that the exception is thrown by specifing:
$scope.listConfig.selectItems = true;
$scope.listConfig.showSelectBox = true;

dtaylor113:
Couldn't a click selection just check the checkbox if shown?

src/views/datalist/data-list-directive.js
```diff
@@ -214,12 +214,22 @@ angular.module('patternfly.views').directive('pfDataList', [
...
         }
       ],
 
       link: function (scope, element, attrs) {
         attrs.$observe('config', function () {
           scope.config = $.extend(true, angular.copy(scope.defaultConfig), scope.config);
+          if (scope.config.selectItems && scope.config.showSelectBox) {
+            throw new Error('pfDataList - ' +
+            'Illegal use of pfDataList directive! ' +
+            'Cannot allow both select box and click selection in the same data list.');
```
waldenraines: These two occurrences of throwing this error should be DRYed.

waldenraines:
Nitpick but could you reword the commit message so that it's < 50 characters as it's a [best practice] (https://github.com/blog/926-shiny-new-commit-styles).

jeff-phillips-18:
@dtaylor113 no, the checkboxes are used when the click event has different purpose. Typically for navigation/drill down.

src/views/datalist/data-list-directive.js
jeff-phillips-18: Actually, I need to remove this second occurrence. There is no way to catch the error so it's not really helping to have it thrown.

dtaylor113:
Ok, LGTM

test/views/datalist/data-list.spec.js
jeff-phillips-18: Added the test

jeff-phillips-18:
@waldenraines Updated the commit message

waldenraines:
@jeff-phillips-18 thanks!

PR #60, opened 2015-09-02
Add pfDataTiles directive, with ngDocs/Examples and Unit Tests

jeff-phillips-18:


dist/styles/angular-patternfly.css
```diff
@@ -171,3 +171,95 @@
...
+  position: relative;
+  text-align: center;
+}
+
+.tiles-view-pf .tile-check-box {
+  left: 10px;
+  position: absolute;
+  top: 8px;
+  width: 20px;
+  z-index: 1000;
```
waldenraines: Why 1000?  Why not 1 or 2 or even 3?

jeff-phillips-18: The idea is only to keep the checkbox on top of the content so that clicking the checkbox does not call selectItem which could be used by the outer code to do something unintended by the user (such as drill down). I will change it to 3.

src/views/datatiles/data-tiles.html
```diff
@@ -0,0 +1,14 @@
+<div class="tiles-view-pf">
+  <div class="tile" ng-repeat="item in items"
+       style="width:{{config.width}}px; height:{{config.height}}px;"
```
waldenraines: What about setting the width and height in the directive itself?

waldenraines:
`data-tiles-directive.js` should be named `data-tiles.directive.js` as per the convention.

src/views/datatiles/data-tiles.html
jeff-phillips-18: What is the aversion to doing it here?

jeff-phillips-18:
@waldenraines The convention is inconsistent. I see both -directive.js and .directive.js

dist/angular-patternfly.js
```diff
@@ -2767,6 +2767,345 @@ angular.module('patternfly.views').directive('pfDataList', [
...
+             <input type="radio" ng-model="selectType" value="none" ng-change="updateSelectionType()">None</input>
+           </label>
+         </div>
+       </form>
+     </div>
+     <div class="col-md-12">
+       <form role="form">
+         <div class="form-group">
+           <label class="checkbox-inline">
+             <input type="checkbox" ng-model="config.dblClick">Double Click</input>
```
waldenraines: It may make sense to `ng-hide` this (or `ng-disabled`) if "Tile" isn't selected above as it's somewhat confusing that you can check "Double Click" and there is no difference if "Checkbox" or "None" is selected.

dist/angular-patternfly.js
```diff
@@ -2767,6 +2767,345 @@ angular.module('patternfly.views').directive('pfDataList', [
...
+             <input type="checkbox" ng-model="showDisabled">Show Disabled Rows</input>
+           </label>
+         </div>
+       </form>
+     </div>
+     <div class="col-md-12">
+       <label style="font-weight:normal;vertical-align:center;">Tile Width: </label>
+       <input style="height:25px; width:60px;" type="number" ng-model="config.width"></input>
+       <label style="font-weight:normal;vertical-align:center;margin-left:20px;">Tile Height: </label>
+       <input style="height:25px; width:60px;" type="number" ng-model="config.height"></input>
```
waldenraines: Why are all of these inline styles necessary?  Can we somehow use the already available styling or put these into a class?

dist/angular-patternfly.js
jeff-phillips-18: Did you mean double click or multi-select? I just realized that with multi-select but don't believe double click changes based on the selection mode.

dist/angular-patternfly.js
jeff-phillips-18: I can do that, just wasn't that concerned about example code.

dist/angular-patternfly.js
jeff-phillips-18: Ah yes, there is no selection difference on a double click unless in select tile mode. I will update both, thanks.

src/views/datatiles/data-tiles-directive.js
```diff
@@ -0,0 +1,339 @@
...
+      restrict: 'A',
+      scope: {
+        config: '=?',
+        items: '=',
+        eventId: '@id'
+      },
+      transclude: true,
+      templateUrl: 'views/datatiles/data-tiles.html',
+      controller: ['$scope',
+        function ($scope) {
```
dtaylor113: Thought we don't need '['$scope'' anymore due to ng-annotate?

waldenraines: :+1:

test/views/datatiles/data-tiles.spec.js
```diff
@@ -0,0 +1,224 @@
...
+
+  // load the controller's module
+  beforeEach(function () {
+    module('patternfly.views', 'patternfly.utils', 'views/datatiles/data-tiles.html');
+  });
+
+  beforeEach(inject(function (_$compile_, _$rootScope_, _$timeout_) {
+    $compile = _$compile_;
+    $scope = _$rootScope_;
+    $timeout = _$timeout_;
```
dtaylor113: I don't think you need $timeout

dtaylor113:
LGTM

waldenraines:
> @waldenraines The convention is inconsistent. I see both -directive.js and .directive.js

Those are misses in reviews.  We started out with `.directive.js` because that came from our work on katello.  That being said, I'm definitely open to changing it.
 
> I personally don't like using the two periods, ex 'foo-bar.directive.js', think we should change them all to 'foo-bar-directive.js', unless there is a reason -which their might be, think I remember karma needing that naming convention? @dtaylor

Karma shouldn't care what they are named, and if it does we should fix that.  I think the `blah.directive.js` is nice because it provides a "sub-type" and, to me, is more readable because it makes the sub-type distinct from the name of the component.  I believe the original naming convention came from this [AngularJS style guide](https://github.com/johnpapa/angular-styleguide#naming).  That being said, John Papa's style guide isn't authoritative nor is it the only style guide in existence.

jeff-phillips-18:
@waldenraines I'll leave it be for now and if someone is concerned they can file an issue to address it across the board.

dist/styles/angular-patternfly.css
jeff-phillips-18: done

dist/angular-patternfly.js
jeff-phillips-18: Done

dist/angular-patternfly.js
jeff-phillips-18: Done

src/views/datatiles/data-tiles-directive.js
jeff-phillips-18: Done

test/views/datatiles/data-tiles.spec.js
jeff-phillips-18: Removed.

jeff-phillips-18:
I will squash once I get another LGTM

dist/styles/angular-patternfly.css
```diff
@@ -171,3 +171,95 @@
...
+  z-index: 3;
+}
+
+.tiles-view-pf .tile-content {
+  height: 100%;
+  left: 0px;
+  margin: 10px;
+  padding-right: 40px;
+  position: absolute;
+  top: 0px;
```
erundle: Why is position absolute needed here?

dist/styles/angular-patternfly.css
```diff
@@ -171,3 +171,95 @@
...
+  border: none;
+  opacity: 0;
+  transition: opacity 0s .2s, opacity .2s linear;
+}
+.tiles-view-pf .tile .more .fa {
+  font-size: 16px;
+  margin-right: 5px;
+}
+.tiles-view-pf .tile:hover .more {
+  opacity: 1;
```
erundle: What are we going to do about mobile?

src/views/datatiles/data-tiles-directive.js
```diff
@@ -0,0 +1,349 @@
...
+       vertical-align:middle;
+       padding-left: 10px;
+     }
+     .dimension-input {
+       height:25px !important;
+       width:60px !important;
+     }
+     .events-label {
+       font-weight:normal;
+     }
```
erundle: FYI I have a PR out to put the demo.css in its own file.

src/views/datatiles/data-tiles-directive.js
```diff
@@ -0,0 +1,349 @@
...
+        config: '=?',
+        items: '=',
+        eventId: '@id'
+      },
+      transclude: true,
+      templateUrl: 'views/datatiles/data-tiles.html',
+      controller: function ($scope) {
+        $scope.defaultConfig = {
+          width: '260',
+          height: '290',
```
erundle: Why are we not just setting this in angular-pf CSS and the products can override if necessary?

src/views/datatiles/data-tiles-directive.js
```diff
@@ -0,0 +1,339 @@
...
+            }
+
+            if (selectionChanged && scope.config.onSelect) {
+              scope.config.onSelect(item, e);
+            }
+            if (selectionChanged && scope.config.onSelectionChange) {
+              scope.config.onSelectionChange(scope.config.selectedItems, e);
+            }
+          }
+          if (scope.config.onClick) {
```
waldenraines: Not sure if it's worth saving the lines but you could default all of these event handlers to `function () {}` and then just call them.  I suppose the check is safer but, then again, it doesn't check that they are actually callable.

dist/styles/angular-patternfly.css
erundle: Actually is this even being used?

dist/styles/angular-patternfly.css
jeff-phillips-18: Because I want the checkbox to be on top of the tile content.

dist/styles/angular-patternfly.css
jeff-phillips-18: Mobile in NYI for this

src/views/datatiles/data-tiles-directive.js
jeff-phillips-18: I thought it would be better if it were calculateable by the callers

src/views/datatiles/data-tiles-directive.js
jeff-phillips-18: I think this is safe enough.

dist/styles/angular-patternfly.css
jeff-phillips-18: Actually, it is not. Left over from the SSUI implementation. I will remove

src/views/datatiles/data-tiles.html
waldenraines: It seems cleaner to me and doesn't result in inline styles in case someone wanted to override the style.  I'm not sure if setting of height and width of the tile is best handled this way (as opposed to just setting it with CSS) but I'm not really clear on the requirements for this component.

src/views/datatiles/data-tiles-directive.js
jeff-phillips-18: moved to CSS

src/views/datatiles/data-tiles.html
jeff-phillips-18: Moved to CSS

dist/styles/angular-patternfly.css
erundle: I think the issue is with your top margin. You probably don't want to use position absolute here. If you remove all the styles and just have .tiles-view-pf .tile-content { margin: 2px 40px 10px 10px; overflow: auto; } it should get you what you want.

jeff-phillips-18: Fixed. Thanks!

PR #62, opened 2015-09-02
Fixes #48: Creating demo.css file

### LINKED ISSUE #48: Create a demo.css file
Create a demo.css file so that we can put demo css into it. This file should be for showcase purposes only and not get distributed via bower.

erundle:
Cleaning up demo css

### DIFF
--- a/src/views/datalist/data-list-directive.js
+++ b/src/views/datalist/data-list-directive.js
@@ -28,16 +28,6 @@
  * @example
 <example module="patternfly.views" deps="patternfly.utils">
   <file name="index.html">
-    <style>
-      hr {
-        display: block;
-        height: 10px;
-        border: 0;
-        border-top: 1px solid #525252;
-        margin: 1em 0;
-        padding: 0;
-      }
-    </style>
     <div ng-controller="ViewCtrl" class="row" style="display:inline-block; width: 100%;">
       <div class="col-md-12">
         <div pf-data-list id="exampleDataList" config="config" items="items">
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -121,7 +121,7 @@ module.exports = function (grunt) {
             'lib/patternfly/dist/js/patternfly.js'],
           html5Mode: false,
           styles: ['lib/patternfly/dist/css/patternfly.css', 'lib/patternfly/dist/css/patternfly-additions.css',
-                   'misc/ngdocs-overrides.css', 'dist/styles/angular-patternfly.css']
+                   'misc/ngdocs-overrides.css', 'dist/styles/angular-patternfly.css', 'misc/demo.css']
         },
         all: ['src/**/*.js']
       },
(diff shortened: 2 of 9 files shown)


src/views/datalist/data-list-directive.js
```diff
@@ -28,16 +28,6 @@
...
-      hr {
-        display: block;
-        height: 10px;
-        border: 0;
-        border-top: 1px solid #525252;
-        margin: 1em 0;
-        padding: 0;
-      }
-    </style>
     <div ng-controller="ViewCtrl" class="row" style="display:inline-block; width: 100%;">
```
dtaylor113: This HR style is also used in the donut, sparkline, and utilization charts

dtaylor113:
LGTM.  Would be great if we could figure out how to get all the styles to show up when one clicks on 'Edit in Plunkr'

src/views/datalist/data-list-directive.js
waldenraines: > This HR style is also used in the donut, sparkline, and utilization charts

i.e. it needs to be removed from there also.

waldenraines:
> Would be great if we could figure out how to get all the styles to show up when one clicks on 'Edit in Plunkr'

If we were to put the styles into their own `<file>` section they would.  The downside of that obviously is that you end up having to repeat similar styles across directives.  Part of me prefers this even though it violated DRY.

src/views/datalist/data-list-directive.js
erundle: Done

jeff-phillips-18:
LGTM

waldenraines:
> If we were to put the styles into their own <file> section they would. The downside of that obviously is that you end up having to repeat similar styles across directives. Part of me prefers this even though it violates DRY.

@erundle thoughts on this?

erundle:
I personally hate style tags since I feel like it promotes bad practices which is why I created the CSS file but if we don't want to take this route I am OK with it.

waldenraines:
> I personally hate style tags since I feel like it promotes bad practices which is why I created the CSS file but if we don't want to take this route I am OK with it.

No, I mean putting the css for each component in `<file>` tags as normal css so they are displayed in plunkr.

dtaylor113:
I don't think it is worth re-creating all those styles in <file> just for plunkr.  Everything works in plunkr, it just isn't styled correctly.

waldenraines:
> I don't think it is worth re-creating all those styles in <file> just for plunkr. Everything works in plunkr, it just isn't styled correctly.

Yeah, you're probably right that DRY > styles in plunkr.  I created an issue for this (#65) in case we want to try to get these styles in plunk in the future.

@erundle can you prepend "Fixes #48: " to your commit message so the issue will close upon merge and so we have a link between the two?

erundle:
Commit message updated.

PR #63, opened 2015-09-03
Created pfTrendsChart directive, ngdocs and unit tests

dtaylor113:


jeff-phillips-18:
This is really a card. It should it be in the cards module and renamed appropriately.

src/charts/trends/trends-chart.directive.js
```diff
@@ -0,0 +1,121 @@
...
+             <input type="checkbox" ng-model="custShowYAxis">Y Axis</input>
+           </label>
+         </div>
+       </form>
+     </div>
+     <div class="col-md-4">
+       <form role="form" >
+         <div class="form-group">
+           <label>Chart Height</label></br>
+           <input style="height:25px; width:60px;" type="number" ng-model="custChartHeight"></input>
```
jeff-phillips-18: move inline styles to css file (coming soon) or to a style block.

erundle:
I don't think we want to add this to the cards module. That puts a dependency on cards that c3 is included which then add affects the base angular patternfly module.

I would prefer to see the card wrapper around this module be removed and just create a trends directive that could be used inside or outside a card.

src/charts/trends/trends-chart.directive.js
```diff
@@ -0,0 +1,121 @@
...
+ * <li>.yData  - Array, Y Values for the data points, first element must be the name of the data
+ * </ul>
+ *
+ * @param {int=} chartHeight   height of the sparkline chart
+ * @param {boolean=} showXAxis override sparkline config settings for showing the X Axis
+ * @param {boolean=} showYAxis override sparkline config settings for showing the Y Axis
+ * @example
+ <example module="patternfly.charts">
+ <file name="index.html">
+  <div ng-controller="ChartCtrl" class="row" style="display:inline-block; width: 100%;">
```
erundle: Is this CSS needed?

src/charts/trends/trends-chart.directive.js
```diff
@@ -0,0 +1,121 @@
...
+   </div>
+   </div>
+  </div>
+ </file>
+
+ <file name="script.js">
+ angular.module( 'patternfly.charts' ).controller( 'ChartCtrl', function( $scope ) {
+
+       $scope.config = {
+         'chartId': 'exampleTrendsChart',
```
erundle: I personally don't care how this is spaced but it should probably be consistent :)

test/charts/trends/trends-chart.spec.js
```diff
@@ -0,0 +1,58 @@
+describe('Directive: pfTrendsChart', function() {
+  var $scope, $compile, element, isolateScope;
+
+  beforeEach(module(
+    'patternfly.charts',
+    'charts/trends/trends-chart.html',
+    'card/basic/card.html',
+    'charts/sparkline/sparkline-chart.html'
```
erundle: Is sparkline needed here?

src/charts/trends/trends-chart.directive.js
dtaylor113: Is what CSS needed?  ..the <ul>?

waldenraines:
> Ugh...wish I saw this before I changed everything ;-(
> Let me digest this suggestion
> -DT

You could always just remove the third commit using `git rebase` to get back to where you were.

dtaylor113:
I don't think we want to add this to the cards module. That puts a dependency on cards that c3 is   
  included which then add affects the base angular patternfly module.

I suggest we leave it as a chart for now as we don't want to put the c3 dependency on cards.

  I would prefer to see the card wrapper around this module be removed and just create a trends 
  directive that could be used inside or outside a card.

I need some more information on how you envision this.
Thanks,
- Dave

erundle:
I would remove the outer card wrapper and just make this a trends directive. If this is really needed as a card then theoretically you would have a second directive that injected and wrapped the preexisting trends directive.

dtaylor113:
Removed the outer card wrapper from the trends directive.
Also fixed indentation in ngdocs.

src/charts/trends/trends-chart.directive.js
dtaylor113: Fixed

src/charts/charts.module.js
```diff
@@ -5,5 +5,5 @@
  *   Charts module for patternfly. Must Include d3.js and c3.js to use
  *
  */
-angular.module('patternfly.charts', []);
+angular.module('patternfly.charts', ['patternfly.card']);
```
erundle: Can this dependency be removed now?

src/charts/trends/trends-chart.directive.js
```diff
@@ -0,0 +1,121 @@
+/**
+ * @ngdoc directive
+ * @name patternfly.charts.directive:pfTrendsChart
+ *
+ * @description
+ *   Directive for rendering a trend chart. The trend chart combines overall data with a pfCard and a
+ *   pfSparklineChart.
```
erundle: Wording update needed.

src/charts/charts.module.js
dtaylor113: Removed

src/charts/trends/trends-chart.directive.js
dtaylor113: Fixed

dtaylor113:
Ok, I believe I have addressed all issues.  Can this be merged?
Thanks,
- Dave

erundle:
LGTM

PR #64, opened 2015-09-03
Addresses issue Donut Chart - color change request #106

dtaylor113:


dist/angular-patternfly.js
```diff
@@ -698,6 +697,7 @@ angular.module('patternfly.charts').directive('pfDonutPctChart', ["c3ChartDefaul
           var color = '#0088CE';
 
           if (thresholds) {
+            color = '#3f9c35';
```
waldenraines: I know that these colors were already in here but I wonder if it makes more sense to add a class (`.threshold-error` and `.threshold-warning`, for example) instead of setting these colors directly.  That would allow the projects that consume angular-patternfly to override these colors if necessary.

dtaylor113: That is a good idea. However, that would require a bunch of rework to
implement as it is coded to return a hex color value.  This is probably
better handled as an enhancement request.  Can you create an issue for
this? - thanks

On Thu, Sep 3, 2015 at 12:25 PM, Walden Raines <notifications@github.com>
wrote:

> In dist/angular-patternfly.js
> <https://github.com/patternfly/angular-patternfly/pull/64#discussion_r38665528>
> :
>
> > @@ -698,6 +697,7 @@ angular.module('patternfly.charts').directive('pfDonutPctChart', ["c3ChartDefaul
> >            var color = '#0088CE';
> >
> >            if (thresholds) {
> > +            color = '#3f9c35';
>
> I know that these colors were already in here but I wonder if it makes
> more sense to add a class (.threshold-error and .threshold-warning, for
> example) instead of setting these colors directly. That would allow the
> projects that consume angular-patternfly to override these colors if
> necessary.
>
> —
> <https://github.com/patternfly/angular-patternfly/pull/64/files#r38665528>
> .
>



-- 
Dave Taylor
Principle Software Engineer
(978) 392 - 1016

jeff-phillips-18:
LGTM

dist/angular-patternfly.js
waldenraines: Yeah, that's fair: #66.

waldenraines:
@dtaylor113 what is the #106 in reference to?

erundle:
+1 to using the CSS class but LGTM for now :)

jeff-phillips-18:
#106 is a patternfly/patternfly issue that also needs to be addressed here

PR #69, opened 2015-09-03
Fix tile borders based on visual design feedback.

jeff-phillips-18:


erundle:
LGTM....merging since this is pretty straight forward.

PR #70, opened 2015-09-04
Fixes #68: Removing c3 from function definition

### LINKED ISSUE #68: 2.1.1 requires that c3/d3 script includes be added to the base project for patternfly module to work
We only want to require c3/d3 if patternfly.charts is being used. We need to remove this dependency.

erundle:


jeff-phillips-18:
LGTM

jeff-phillips-18:
Merging since this is very straightforward and fixes a build issue.

PR #71, opened 2015-09-04
Add pfDataToolbar directive with ngDocs/Example and Unit Tests

jeff-phillips-18:


src/views/toolbar/data-toolbar-directive.js
```diff
@@ -0,0 +1,248 @@
...
+          return $scope.config.viewsConfig && ($scope.config.viewsConfig.currentView === viewId);
+        };
+
+        $scope.checkViewDisabled = function (view) {
+          return $scope.config.viewsConfig.checkViewDisabled && $scope.config.viewsConfig.checkViewDisabled(view);
+        };
+      },
+      link: function (scope, element, attrs) {
+        scope.$watch('config', function () {
+          if (scope.config && scope.config.viewsConfig && scope.config.viewsConfig.views) {
```
dtaylor113: Should this be moved to the controller?

jeff-phillips-18: No, watching and observing have to be done post compile (ie in the link function)

dtaylor113: Thought we could 'watch' in controller:, and 'observe' in the link:

jeff-phillips-18: I've always kept them in the link.

dtaylor113:
I think the pf-selector's "Address" is to low in the box:
apf-toolbar

From PF:
pf-toolbar-orig

jeff-phillips-18:
@dtaylor113 Yeah, this is only in the example though. It's fine when used in an application. I just couldn't get the right demo.css settings to fix it here.

dtaylor113:
Weird list item stacking when selecting between "Birth Month" and "Name".  The input box to the right of the pf-selector momentarily appears vertically stacked before collapsing into the single placeholder.  Is this just in the demo?

jeff-phillips-18:
Yes, again it's in the example only.

src/views/toolbar/data-toolbar-directive.js
```diff
@@ -0,0 +1,248 @@
...
+      $scope.listConfig = {
+        selectionMatchProp: 'name',
+        checkDisabled: false
+      };
+    }
+  ]);
+  </file>
+</example>
+ */
+angular.module('patternfly.views').directive('pfDataToolbar',
```
dtaylor113: I don't think a 'toolbar' should be under 'views', I know it consumes the 'views' mechanism, but it should probably go into a general 'components' directive.  For now it's fine.

dtaylor113:
LGTM

jeff-phillips-18:
@waldenraines Do you have a moment to review this?

waldenraines:
@jeff-phillips-18 I do but too late :frowning:

PR #73, opened 2015-09-08
This fixes #61 and fixes #72

### LINKED ISSUE #61: Missing info. in Sparkline chart ng-docs
Please add a description of 'total' to the chartData ng-doc 

chartData	  object  the data to be shown as an area chart
.xData - Array, X values for the data points, first element must be the name of the data
.yData - Array, Y Values for the data points, first element must be the name of the data
**.total - (optional)  The Total amount, used when determining percentages**

### LINKED ISSUE #72: If jQuery is not being used in application , than pfDonutPctChart directive having issue with $.extend function
If jQuery is not being used in application , than in pfDonutPctChart directive having issue with $.extend function , this is jQuery function , and here jquery is not being used , so either  $.extend wherever is used that should be not there. or we should have our own custom javascript extend function , that will merge two object and return in to the first object .

dtaylor113:
Addresses Issues 
#61 "Missing info. in Sparkline chart ng-docs"
#72 "If jQuery is not being used in application , than pfDonutPctChart directive having issue with $.extend function"


waldenraines:
I personally think it's better to do one PR per issue.  Also if you use "Fixes #<issue number>" in your commit it will close the issue automatically on merge.  See [closing issues via commit messages](https://help.github.com/articles/closing-issues-via-commit-messages/).

test/charts/sparkline/sparkline-chart.spec.js
```diff
@@ -105,7 +105,7 @@ describe('Directive: pfSparklineChart', function() {
 
     expect(isolateScope.config.data.x).toBe("dates");
     expect(isolateScope.config.data.columns.length).toBe(2);
-    expect(isolateScope.config.data.columns[0][1]).toBe($scope.data.xData[1]);
+    expect(isolateScope.config.data.columns[0][1].toString()).toBe($scope.data.xData[1].toString());
```
waldenraines: Why is the `toString()` necessary after these changes?

waldenraines:
> I personally think it's better to do one PR per issue.

Disregard this, I didn't realize the issues were so closely related.

waldenraines:
Also, the commit summary itself should be under [72 characters] (https://github.com/blog/926-shiny-new-commit-styles) with extra info in the extended description, if necessary.  Otherwise, LGTM.

test/charts/sparkline/sparkline-chart.spec.js
dtaylor113: I believe the unit test was checking that the actual javascript objects were the same, after the change:
+     scope.config = angular.merge({}, scope.defaultConfig, scope.config);
-      scope.config = $.extend(true, angular.copy(scope.defaultConfig), scope.config);
They don't appear to be, even though their contents are the same.  To me the angular.merge and the $.extend cmds should result in new javascript objects, so I'm not sure why/what is different between the commands.  In other words, I wouldn't think the 'toString()' would be neccesary.

erundle:
LGTM but I would rebase and update the commit message as Walden suggested.

test/charts/sparkline/sparkline-chart.spec.js
waldenraines: Thank makes sense, thanks for the explanation.

waldenraines:
@dtaylor113 I don't think "addresses issue..." will close it.  See [the list](https://help.github.com/articles/closing-issues-via-commit-messages/#keywords-for-closing-issues) of acceptable keywords for closing issues.

jeff-phillips-18:
LGTM

waldenraines:
> 1 check was pending

:question:

jeff-phillips-18:
Didn't realize the check was still pending...

Issue #35, opened 2015-08-19
Support i18n and l10n


waldenraines:
We have several English strings in angular-patternfly and should support internationalization and localization.

I've used [angular-gettext](https://github.com/rubenv/angular-gettext) in the past but I think it would be better to provide files per language if possible using something like [grunt-static-i18n](https://github.com/beck/grunt-static-i18n) perhaps?  Then the consumer of angular-patternfly could just include the js file for the language they need.

dtaylor113:
Hi, yeah, I was wondering about that.  We do have several hardcoded, customer facing, strings in angular-patternfly.  
+1 for this suggestion

dtaylor113:
We were considering a simpler approach that strings are passed in via a language object like this:

http://api.highcharts.com/highcharts#lang

We have a copy of the object with english defaults in the directive and then if the application passes us one as well we overlay what they gave us with what we have and those are the versions of the strings that we use. 

Keep it simple and let the products handle the translations and this way they can add/remove languages as they see fit.  I am going to recommend staying away from doing the actual translation since there are multiple translation services out there that teams are using

angular-translate
http://i18next.com/
angular-gettext etc

I would rather have them pick what they want instead and give us the strings then to deal with multiple ways of picking languages.

erundle:
I would prefer to have the consuming applications pass in the strings to the given directives than having angular-patternfly handle the translations directly. If we were to handle the translations we would need to check in translations for all possible languages and the parent application would need to tell us what language to use dynamically at runtime (Need to be able to handle user preferences or browser settings). Since our various consumers seem to be using different translation packages and our string are limited it seems like having the consuming application handle the translations would be the way to go. This also provides for greater customization ability if the applications want to change strings slightly.

waldenraines:
Those are good points, I have to keep reminding myself that this is a library of components and not a web application :smile: 

I agree that we should allow the users of angular-patternfly the ability to pass in their own strings.

Will we need some way to handle plurals?  A switch and replace is obviously fine for "1 item", "2 items" but it gets tricky with languages that support multiple plural forms.  I still think we should use a translation library in angular-patternfly so we are sure to handle these types of issues.  We would still let the user provide the actual translations though.  Thoughts?

jeff-phillips-18:
Unlikely this will be addressed in the current expected life span of Angular 1.

PR #74, opened 2015-09-08
Fixes: #35 support i18n

erundle:
Handles basic string replacement. Assumes products translate strings and pass in via lang object.


src/charts/donut/donut-pct-chart-directive.js
```diff
@@ -40,6 +40,14 @@
  * <li> 'percent'   - displays the Usage Percent of the Total amount in the center label
  * <li> 'none'      - does not display the center label
  * </ul>
+ *
+ * @param {object} lang translated i18n strings<br/>
```
dtaylor113: If this is an optional param, then it should be:

@param {object=} lang translated i18n strings<br/>

dist/angular-patternfly.js
```diff
@@ -771,10 +796,10 @@ angular.module('patternfly.charts').directive('pfDonutPctChart', ["c3ChartDefaul
             centerLabelText.smText = '';
           } else if ($scope.centerLabel === 'available') {
             centerLabelText.bigText = $scope.data.available;
-            centerLabelText.smText = $scope.config.units + ' Available';
+            centerLabelText.smText = $scope.config.units + ' ' + $scope.lang.available;
           } else if ($scope.centerLabel === 'percent') {
             centerLabelText.bigText = Math.round($scope.data.used / $scope.data.total * 100.0) + '%';
-            centerLabelText.smText = 'of ' + $scope.data.total + ' ' + $scope.config.units;
+            centerLabelText.smText = $scope.lang.of + ' ' + $scope.data.total + ' ' + $scope.config.units;
```
erundle: String concat is bad since we need to handle the ability to reorder strings in say Japanese. I need to see what we can do there but I wanted to get some feedback on this approach

src/charts/donut/donut-pct-chart-directive.js
```diff
@@ -215,6 +239,7 @@ angular.module('patternfly.charts').directive('pfDonutPctChart', function (c3Cha
     scope: {
       config: '=',
       data: '=',
+      lang: '=',
```
dtaylor113: If this is optional, it should be:
lang: '=?',

dtaylor113:
Is this just a PR to get the coding pattern approved, and then it will be applied to the other charts?

dtaylor113:
Other than those issues, LGTM

erundle:
Yes I wanted to show the concept and if we think it makes sense I will apply this to other directives.

waldenraines:
I personally think we should use the strings themselves as keys and pass them into a translation service (or a directive that uses the service in the case of html).   

We could use [angular-gettext](https://angular-gettext.rocketeer.be/) for this.  The user would just have to provide the `translation.js` file ([example of the grunt generated angular-gettext file](https://github.com/Katello/bastion/blob/master/app/assets/javascripts/bastion/i18n/translations.js)), which populates a map between the keys (the actual words) and the translations.  This translation service replaces the strings with the user-provided translations based on `$locale`.

I don't think we want to force the user to set every single string they want translated via a javascript object in the code that uses the angular-patternfly component.  This quickly becomes cumbersome:

 ```javascript 
$scope.lang = {
  available: 'Disponible',
  notAvailable: 'No disponible',
  yes: 'si',
  no: 'no'  
};
```
Instead we should encourage users to use .po files (pretty much the standard for i18n) and generate a javascript dictionary from that.  This approach also has the added benefit of falling back to English in the case that a translation is not found instead of resolving to undefined or resulting in an error.

Also, how would the implementation here support plurals?  What about languages with more than one plural form?  Using an existing translation library handles these issues for you.

**Comment by erundle** [comment]
I am not completely against using existing translation but the parent application will need to tell us what language to use. We also need to make sure we can bootstrap multiple translation libraries (I have seen this be an issue in angular applications in the past where we actually had to merge translation files since config was done as a singleton). As far as .po files vs. json when it comes to client side libraries JSON seems to be the new norm:
https://angular-translate.github.io/
http://i18next.com/

**Comment by erundle** [comment]
Also the beautiful part of not doing the actual translation is products can add languages based on their needs so we don't need to provide translations for every language out there and if a product wants to easily change a string value they can. If they want to use "Free" vs. "Available" vs. "Not Used" vs. "Whatever" they are free to customize things from a terminology perspective as they see fit.

**Comment by jeff-phillips-18** [comment]
I too believe the approach here is too cumbersome.

Could we use something like I18Next and require the application to set the locale? We could use I18Next for all of our strings and provide the appropriate translation files. If the application wanted to use some other library they could . Strings passed to our directives could be pre-translated or they could be keys.

Granted it would be easier to not have to provide translations. I just think requiring applications to pass in lang objects for every string we want to present is a bit overbearing and if we change how we present strings or which strings we present then the applications would need to change.

**Comment by waldenraines** [comment]
> We also need to make sure we can bootstrap multiple translation libraries 

That's a good point.  I'm not opposed to writing something ourselves but I do think that the translations should be done via a service/directive instead of with plain JS objects.

> As far as .po files vs. json when it comes to client side libraries JSON seems to be the new norm:
> https://angular-translate.github.io/
> http://i18next.com/

i18next supports [gettext](http://i18next.com/pages/ext_i18next-conv.html) and I'd argue that angular-translate is a bad implementation.  In addition to not supporting .po files it also requires made up IDs for the translation dictionary, and it uses filters instead of directives which is not as performant.

We could just write a simple service that reads in a JSON file (to be provided by the consuming project via whatever way they want) and then looks up values in that file based on `$locale`.

```json
{
    "es": {
        "yes": "sí",
        "no": "no"
    },
    "fr": {
        "yes": "oui",
        "no": "non"
    }
}
```

```html
<p i18n>Yes</p>
```

```javascript
var yes = i18n('yes');
```

Although we would still have to consider how to handle plurals...

waldenraines:
> Also the beautiful part of not doing the actual translation is products can add languages based on their needs so we don't need to provide translations for every language out there and if a product wants to easily change a string value they can. If they want to use "Free" vs. "Available" vs. "Not Used" vs. "Whatever" they are free to customize things from a terminology perspective as they see fit.

I agree, and I was not suggesting that we should do our own translations, just that we should create an easier way for the consumer to provide the translations.

erundle:
FYI this handles plurals https://docs.angularjs.org/api/ng/directive/ngPluralize

waldenraines:
> FYI this handles plurals https://docs.angularjs.org/api/ng/directive/ngPluralize

Sort of, you still have to mark the string for translation and then override the localization rules, check out the first paragraph on that page:

>ngPluralize is a directive that displays messages according to en-US localization rules. These rules are bundled with angular.js, but can be overridden (see Angular i18n dev guide). You configure ngPluralize directive by specifying the mappings between plural categories and the strings to be displayed.

erundle:
Please let me know if anyone has an objection to using this. I started to write my own provider etc to wrap i18next and then I realized that was silly. I think i18next is our best bet so I was thinking of adding this as well: https://github.com/i18next/ng-i18next. I have this plugged in and working now but I wanted to see if anyone had any major push back before I went further. Pending everyone thinks that makes sense I will close this PR and open up a new one once I am ready.

jeff-phillips-18:
I agree with that approach.

waldenraines:
> I think i18next is our best bet

I agree since it allows for both JSON and gettext (looks like there is also [grunt task](https://github.com/aaronlecompte/grunt-i18next-conv) for converting from gettext to i18next's JSON format).

Also, it seems that ng-i18next provides a filter which we can inject for translations within our JS code.  This approach looks good to me from a quick glance, thanks for researching.

dtaylor113:
LGTM,
Thanks,
- Dave

On Wed, Sep 9, 2015 at 12:58 PM, Eleni Rundle <notifications@github.com>
wrote:

> Please let me know if anyone has an objection to using this. I started to
> write my own provider etc to wrap i18next and then I realized that was
> silly. I think i18next is our best bet so I was thinking of adding this as
> well: https://github.com/i18next/ng-i18next. I have this plugged in and
> working now but I wanted to see if anyone had any major push back before I
> went further. Pending everyone thinks that makes sense I will close this PR
> and open up a new one once I am ready.
>
> —
> <https://github.com/patternfly/angular-patternfly/pull/74#issuecomment-138974403>
> .
>



-- 
Dave Taylor
Principle Software Engineer
(978) 392 - 1016

PR #67, opened 2015-09-03
Add pfSimpleFilter directive with ngDocs/Example and Unit Tests (PF 2.2 Dependent)

jeff-phillips-18:
Adds a simple filter bar directive.

PR #76, opened 2015-09-09
Add pfSimpleSort directive with unit tests and ng-docs

jeff-phillips-18:


src/sort/simple-sort-directive.js
```diff
@@ -0,0 +1,239 @@
...
+ */
+angular.module('patternfly.sort').directive('pfSimpleSort',
+  function ($document) {
+    'use strict';
+    return {
+      restrict: 'A',
+      scope: {
+        config: '='
+      },
+      transclude: false,
```
waldenraines: Isn't this false by default?

src/sort/sort.module.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * @name  patternfly card
+ *
+ * @description
+ *   Filters module for patternfly.
```
waldenraines: Needs to be updated after copy/paste.

src/sort/simple-sort-directive.js
```diff
@@ -0,0 +1,239 @@
...
+ * <ul style='list-style-type: none'>
+ * <li>.fields          - (Array) List of sortable fields containing:
+ * <ul style='list-style-type: none'>
+ * <li>.id          - (String) Unique Id for the sort field
+ * <li>.title       - (String) The title to display for the sort field
+ * <li>.sortType    - (String) The sort type, 'alpha' or 'numeric'
+ * </ul>
+ * <li>.sortId   - (Object) Id of the current sort field
+ * <li>.sortDirection - (String) The current sort direction 'asc' or 'desc'
+ * <li>.onSortChange - ( function(sortId, sortDirection ) Function to call when the current sort params change
```
waldenraines: These '.'s are showing up in the param documentation:

screen shot 2015-09-09 at 5 05 10 pm

waldenraines:
It would be nice if I could use the arrow keys and enter to select a filter.  This would be a good opportunity to incorporate the `onEnter` directive I mentioned [here](https://github.com/patternfly/angular-patternfly/pull/67#discussion_r38695363).

src/sort/simple-sort-directive.js
```diff
@@ -0,0 +1,239 @@
...
+          ],
+          onSortChange: sortChange
+        };
+      }
+    ]);
+  </file>
+</example>
+ */
+angular.module('patternfly.sort').directive('pfSimpleSort',
+  function ($document) {
```
waldenraines: `$document` doesn't appear to be used anywhere.

src/sort/simple-sort-directive.js
```diff
@@ -0,0 +1,239 @@
...
+            $scope.config.fields = [];
+          }
+
+          if ($scope.config.fields.length > 0) {
+            if ($scope.config.currentField === undefined) {
+              $scope.config.currentField = $scope.config.fields[0].id;
+              updated = true;
+            }
+            if ($scope.config.currentDirection === undefined) {
+              $scope.config.currentDirection = 'asc';
```
waldenraines: Thoughts on using a boolean for this instead of a string?

src/sort/simple-sort.html
```diff
@@ -0,0 +1,22 @@
...
+          </button>
+          <ul class="dropdown-menu">
+            <li ng-repeat="item in config.fields" ng-class="{'selected': item.id == config.currentField}">
+              <a class="sort-field" role="menuitem" tabindex="-1" ng-click="selectField(item)">
+                {{item.title}}
+              </a>
+            </li>
+          </ul>
+        </div>
+        <a><i class="sort-direction {{getSortIconClass()}}" ng-click="changeDirection()"></i></a>
```
waldenraines: Maybe use `ng-class` instead here?

src/sort/simple-sort-directive.js
```diff
@@ -0,0 +1,239 @@
...
+            $scope.config.onSortChange($scope.config.currentField, $scope.config.currentDirection);
+          }
+        };
+
+        $scope.changeDirection = function () {
+          if ($scope.config.currentDirection === 'asc') {
+            $scope.config.currentDirection = 'desc';
+          } else {
+            $scope.config.currentDirection = 'asc';
+          }
```
waldenraines: If you were to use a boolean then this simpy becomes `$scope.config.currentDirection = !$scope.config.currentDirection;`

src/sort/simple-sort-directive.js
```diff
@@ -0,0 +1,239 @@
...
+          });
+
+          if (updated === true && $scope.config.onSortChange) {
+            $scope.config.onSortChange($scope.config.currentField, $scope.config.currentDirection);
+          }
+        };
+
+        $scope.selectField = function (field) {
+          $scope.config.currentField = field.id;
+          $scope.config.currentSortField = field;
```
waldenraines: Why not just set `$scope.config.currentField` and then just reference the ID from there (ala `$scope.config.currentField.id`) when you need it?

src/sort/simple-sort-directive.js
jeff-phillips-18: Yes. I typically set it anyhow. I will remove though.

src/sort/simple-sort-directive.js
jeff-phillips-18: This is by design. This describes the config object so this is trying to show the fields for the object.

src/sort/simple-sort-directive.js
jeff-phillips-18: I've used 'asc' and 'desc' as this is what I've always seen in the past. Specifically with MongoDB and other DB APIs. Boolean would be easier for sure.

dtaylor113:
LGTM

jeff-phillips-18:
Addressed the open issues. Let me know and I will squash before merge.

dist/angular-patternfly.js
```diff
@@ -2469,6 +2478,233 @@ angular.module('patternfly.select', []).directive('pfSelect', ["$timeout", funct
...
+        };
+
+        $scope.getSortIconClass = function () {
+          var iconClass;
+
+          if ($scope.config.currentField.sortType === 'alpha') {
+            if ($scope.config.isAscending) {
+              iconClass = 'fa fa-sort-alpha-asc';
+            } else {
+              iconClass = 'fa fa-sort-alpha-desc';
```
waldenraines: What about using ng-class for these classes rather than having them in JS code?

src/sort/simple-sort-directive.js
```diff
@@ -0,0 +1,227 @@
...
+        };
+
+        $scope.getSortIconClass = function () {
+          var iconClass;
+
+          if ($scope.config.currentField.sortType === 'alpha') {
+            if ($scope.config.isAscending) {
+              iconClass = 'fa fa-sort-alpha-asc';
+            } else {
+              iconClass = 'fa fa-sort-alpha-desc';
```
waldenraines: What about using ng-class for these classes rather than having them in JS code?

src/sort/simple-sort-directive.js
```diff
@@ -0,0 +1,227 @@
...
+
+          if ($scope.config.onSortChange) {
+            $scope.config.onSortChange($scope.config.currentField, $scope.config.currentDirection);
+          }
+        };
+
+        $scope.getSortIconClass = function () {
+          var iconClass;
+
+          if ($scope.config.currentField.sortType === 'alpha') {
```
waldenraines: Do we envision types of sorts other than 'alpha' and 'numeric'?  Wondering if this ought to be a boolean as well.

src/sort/simple-sort-directive.js
jeff-phillips-18: There are 4 variations all which would check almost the same things. I thought it clearer to provide a function for this logic.

src/sort/simple-sort-directive.js
jeff-phillips-18: I could see additional sort types being requested in the future

waldenraines: Do you have an example?  I couldn't think of one but I suppose I could see a consumer of angular-patternfly wanting to provide their own in the future so I'm fine with keeping this as is.

src/sort/simple-sort-directive.js
waldenraines: That's a good point and this way we can test it.  Speaking of, I don't see any tests for this function.

jeff-phillips-18: The tests check for the correct icon class after selecting different sort fields and directions.

src/sort/simple-sort-directive.js
jeff-phillips-18: Amount is one for sure. There is already Font Awesome icons for sort-amount-asc and sort-amount-desc. Patternfly has not yet endorsed these however.

src/sort/simple-sort-directive.js
waldenraines: But we don't want to test this explicitly?  I expected a test for it since it's on the `$scope` and thus part of the API.

jeff-phillips-18: Done.

src/sort/simple-sort-directive.js
erundle: Yea I am not a fan of this. I think we are going to run into issues with translation for this approach so I think we should just go with a defacto sort icon until this is worked out

jeff-phillips-18: These icons are what the sort spec provided as part of the PF toolbar spec. This meets the PF standard that is being created.

erundle: This doesn't translate. We need to rethink this approach.

erundle: http://fortawesome.github.io/Font-Awesome/icon/sort-alpha-asc/

erundle: The same goes for numbers. http://prek-8.com/1stgrade/lessonplan_chinaNumber.html

jeff-phillips-18: OK. Can we add an issue to address i18n support for these icons and move forward?

jeff-phillips-18:
Issue added for i18n support for icons in the direction button.
All other issues are addressed.
Squashed commits.

src/sort/simple-sort-directive.js
erundle: https://github.com/patternfly/angular-patternfly/issues/79

waldenraines:
LGTM

styles/angular-patternfly.css
```diff
@@ -286,3 +286,24 @@
   background-image: none;
   color: #999999;
 }
+
+.simple-sort .sort-direction {
+  font-size: 22px;
+  color: #222222;
```
erundle: in the future just use #222 if everything is repeating.

src/sort/simple-sort-directive.js
```diff
@@ -0,0 +1,227 @@
...
+            } else {
+              iconClass = 'fa fa-sort-alpha-desc';
+            }
+          } else {
+            if ($scope.config.isAscending) {
+              iconClass = 'fa fa-sort-numeric-asc';
+            } else {
+              iconClass = 'fa fa-sort-numeric-desc';
+            }
+          }
```
erundle: Shouldn't we have a default that isn't numeric? Allow them to specify "alphabetic", "numeric" and if not get the default. If we aren't specifying the sort function and we are allowing them to pass in whatever columns they want I would think we would need to support at a basic level other possible values (Dates for example). Besides that this LGTM

jeff-phillips-18: Agreed, good catch.

PR #77, opened 2015-09-10
Fixes: #35 support i18n

erundle:
Adding i18next translation framework and ng-i18next for added angular support. Angular patternfly will only provide english translations to date.


dtaylor113:
LGTM, Nice job, it will be great to remove those hardcoded english strings from our code :smiley:

src/charts/donut/donut-pct-chart-directive.js
```diff
@@ -305,7 +307,7 @@ angular.module('patternfly.charts').directive('pfDonutPctChart', function (c3Cha
 
           // default to 'used' info.
           centerLabelText = { bigText: $scope.data.used,
-                              smText:  $scope.config.units + ' Used' };
+                              smText:  $i18next('pfDonutPctChart.used', { units: $scope.config.units }) };
```
waldenraines: Why not just use the words themselves as the keys in the hash?  Is the thought that people will want to alter these individually per component?

erundle: Yes they may potentially want to change it per chart but doing it this way also gives translators the context of what they are translating. Without the directive as a prefix they may have issues figuring out where the string is actually being used.

waldenraines: Hmm, I suppose it's a trade off.  As the user I'd have to specify 'available' and 'used' several times if I used more than one type of chart (for example).  And it also makes it more difficult for a developer to follow where text comes from as there is an extra level of abstraction.

I wonder if we could add the context in the string extraction process.  Gettext is nice for this sort of thing because it tells you the filename (and line) that the string comes from so it provides context that way.

jeff-phillips-18:
LGTM, I'm OK with the format to allow specification by component.

erundle:
@waldenraines can we move forward with this. I agree this is a tradeoff and not necessarily ideal that we may duplicate text but this approach does provide the appropriate context and also provides flexibility.

waldenraines:
> @waldenraines can we move forward with this. I agree this is a tradeoff and not necessarily ideal that we may duplicate text but this approach does provide the appropriate context and also provides flexibility.

@erundle can you provide an example of how a consumer of angular-patternfly would provide translations?

erundle:
They will need to provide this: https://github.com/erundle/angular-patternfly/blob/master/i18n/demo-i18n.js

Which contains the path to their translation files. If they only have english they will likely need to include the path to our json file/copy it to whatever locations they want.

waldenraines:
> If they only have english they will likely need to include the path to our json file/copy it to whatever locations they want.

Will it use the default English strings without action on the consumer's part?  In other words, this is always required regardless of whether or not they want to change/translate the strings?

Also, it looks like this is required per language, correct?

erundle:
Right now we are not shipping this code so the consuming application needs to add it (Hence to why it is in demo). I don't know how to get around this which is why I didn't really want to add a translation framework to begin with. We either need to be all inclusive (We have all the translations etc and we can bootstrap ourselves) or we need to pass of the bootstrapping to the consuming application. If you have a better idea on how to handle this please let me know.

waldenraines:
@erundle I think this approach would be sound if we were able to provide default text.  I am surprised that i18next doesn't support supplying a default language and then allowing overriding of that language.  It looks like in the demo that i18next supports a fallback language, is there a way to set that as our default strings?

I don't have a better idea on how to handle this without being able to set aside some time to look into the specifics though.

src/charts/utilization/utilization-chart.html
```diff
@@ -9,7 +9,7 @@ <h1 class="available-count pull-left">
                 <span>{{currentText}}</span>
             </div>
             <div>
-                <span> of {{chartData.total}} {{config.units}}</span>
+                <span>{{'pfUtilizationChart.percentOf' | i18next:{units: config.units, value: chartData.total} }}</span>
```
waldenraines: Doesn't ng-i18next provide a directive to do this?  Filters are supposed to be lightweight.

erundle: I didn't see a directive that does value replacement. Only options but maybe I missed something.

erundle:
Fallback language means that the language set by the user/system doesn't match any of the resource files we still need to configure the i18next provider and tell if where to look for resources. I can't easily see how we can configure it in source code and then also allow it to be configured in the consuming product.

jeff-phillips-18:
I'm OK with requiring consumers to config the i18n settings since there is no good way to automatically config because we won't know where the default translation file will live in the distribution. We could try to configure the settings if the application has not, but we would be guessing at the file location.

However, we may want to do this in a full release version (3.0) and not in a point release since it will be important for consumers to read the release notes.

erundle:
After talking with @jeff-phillips-18 jeff- I added defaultValue to the strings. If a provider is not configured like in demo-i18n.js it just fallback to the strings that are in the code as the defaults. If the provider is configured it uses those instead. What I don't love about this approach is it forces us to duplicate strings and it makes the definitions verbose. Thoughts?

https://github.com/erundle/angular-patternfly/commit/c35df466d711779bc78e1c60060e3bda2742620c

ehelms:
Sorry if I missed this in the comments, but as a user how do I extract these strings to provide to my translators? Also, having worked on a project that had a time figuring out and getting translations juts right, I am curious why a non-native angular translation framework (such as ngettext or angular-translate) was chosen over one that requires wrapping? I know that part of the 1.4+ and Angular2 work is to provide a core framework solution to this issue, so being closer to that world would seem to also make transition easier in the future.

erundle:
Hey @ehelms we have json files that keep the extracted strings:
https://github.com/erundle/angular-patternfly/blob/master/i18n/en/patternfly-resources.json

This gets copied to our dist directory so you would likely in your build process copy that file to a location that you prefer and then configure a provider based on the your path and application language settings:
https://github.com/erundle/angular-patternfly/blob/master/i18n/demo-i18n.js

I personally wanted to stay away from something this heavy and just pass in a 'lang' object to each directive that contained the already translated strings but others didn't want to go that route:
https://github.com/patternfly/angular-patternfly/pull/74

As far as the library choice angular-translate does not work well in a toolkit since it is a singleton and does not provide namespacing and I believe ng-gettext has the same limitation.

ehelms:
My brain is not seeing what you are thinking, do you mind explaining to me what is it about the singleton nature that makes it difficult?

erundle:
I previously had to do some hackery when I had multiple applications/modules that were both using the angular-translate. It wanted to pick up the translations of the parent application only and it ignored the translations of the inner module (this case angular-patternfly) since the translation service has already been configured at the application level. I believe we ended up combining the resource files to get around this issue which wasn't ideal. Maybe this is no issue but was what I encountered at the time.

waldenraines:
> I personally wanted to stay away from something this heavy and just pass in a 'lang' object to each directive that contained the already translated strings but others didn't want to go that route:

:fire: :ear: 

To me this approach still makes the most sense:

 1. We wrap our strings (ideally using normal strings as keys, not object properties) in a translate directive/filter that extracts them when running `grunt i18n:extract` (either into a `.pot` file or something else)
 2. Translations are performed by the consumer (although we could provide some translations ourselves and save a lot of duplicated effort for consuming projects)
 3. Translations are added to the application via an [angular service](https://github.com/Katello/bastion/blob/master/app/assets/javascripts/bastion/i18n/translations.js) such as `gettextCatalog` in this example.
 4. The language is set via [`$locale`](https://docs.angularjs.org/api/ng/service/$locale)
 5. Translations are displayed and will be automatically updated if the `$locale` changes

I still think that [angular-gettext](https://github.com/rubenv/angular-gettext) is a good way to do this.  Note that this approach would also allow overriding the default english strings by adding an appropriate translation entry for `en`.

jeff-phillips-18:
Does anyone want to make an attempt at what @waldenraines proposes or should we go with this approach? My feeling is that this approach is sound and that @erundle has addressed all the issues and the requirements are met.

I will merge unless I hear someone wanting to provide an alternate approach.

waldenraines:
> I will merge unless I hear someone wanting to provide an alternate approach.

Is there a rush to get this merged?  

I'll try to make some time to make an attempt today or tomorrow.

erundle:
@waldenraines no rush so go for it. Hopefully you can come up with something better. My only concern is being able to have multiple translation files for a given language so whatever framework we pick needs to support that.

ehelms:
I think angular-gettext would support what you are needing. In Katello, we have multiple code bases that supply strings to the same gettextCatalog using the 'setStrings' method. This method, as far as I can tell looking at the code itself, will add strings to a language in an additive way if there are multiple declarations.

erundle:
Closing. Will revisit at a later date

PR #80, opened 2015-09-10
Tagging for 2.2.1. SimpleFilter bug fix

erundle:
This is needed for manageIQ so they can consume the filter fix.


waldenraines:
LGTM

PR #82, opened 2015-09-11
Trend Chart Card updates

dtaylor113:
- Added ability to change card title from actual value '76 GB' to percentage value '30% of 250 GB'
- Added ability to change layout from 'large' card to 'small' card
- Updated Gruntfile.js: ngdocs:watch to rebuild upon html and css changes


Gruntfile.js
```diff
@@ -199,8 +199,8 @@ module.exports = function (grunt) {
           files: ['Gruntfile.js'],
           tasks: ['eslint']
         },
-        js: {
-          files: ['Gruntfile.js', 'src/**/*.js'],
+        all: {
+          files: ['Gruntfile.js', 'src/**/*.js', 'src/**/*.html', 'styles/**/*.css'],
           tasks: ['build']
```
waldenraines: How do you feel about making this a separate PR since it is distinct from the actual feature work.

src/charts/trends/trends-chart.directive.js
```diff
@@ -51,15 +53,29 @@
...
+             </select>
            </div>
          </form>
        </div>
-       <div class="col-md-4">
+       <div class="col-md-3">
+         <form role="form" >
+           <div class="form-group">
+             <label>Layout</label></br>
+             <select pf-select style="height:25px; width:120px;" ng-model="layout" id="layout">
```
waldenraines: Can we get these styles into demo.css since they are repeated above?

waldenraines: Either that or include a stylesheet via `<file>`.

Gruntfile.js
dtaylor113: I feel like this is a simple, non-intrusive, change and doesn't really warrant a separate PR ;-)

src/charts/trends/trends-chart.directive.js
```diff
@@ -114,7 +139,34 @@ angular.module('patternfly.charts').directive('pfTrendsChart',
         showYAxis: '=?'
       },
       replace: true,
-      templateUrl: 'charts/trends/trends-chart.html'
+      templateUrl: 'charts/trends/trends-chart.html',
+      controller: ['$scope',
+        function ($scope) {
```
waldenraines: This style is no longer necessary.

src/charts/trends/trends-chart.directive.js
```diff
@@ -114,7 +139,34 @@ angular.module('patternfly.charts').directive('pfTrendsChart',
         showYAxis: '=?'
       },
       replace: true,
-      templateUrl: 'charts/trends/trends-chart.html'
+      templateUrl: 'charts/trends/trends-chart.html',
+      controller: ['$scope',
```
waldenraines: This style is no longer necessary.

src/charts/trends/trends-chart.directive.js
```diff
@@ -114,7 +139,34 @@ angular.module('patternfly.charts').directive('pfTrendsChart',
...
+        function ($scope) {
+          $scope.getPercentageValue = function () {
+            return Math.round($scope.getLatestValue() / $scope.chartData.total * 100.0);
+          };
+          $scope.getLatestValue = function () {
+            return $scope.chartData.yData[$scope.chartData.yData.length - 1];
+          };
+          $scope.getChartHeight = function () {
+            var retValue = 60;
```
waldenraines: Thoughts about defining these as constants?

```javascript
const SMALL = 30, LARGE = 60;
```

src/charts/trends/trends-chart.directive.js
dtaylor113: What style?  Can you please elaborate?  -Thanks

waldenraines: Sorry, the annotation style of dependency injection.

src/charts/trends/trends-chart.html
```diff
@@ -1,7 +1,20 @@
...
+<span>
+  <div ng-class="{'trend-card-large-pf': showLargeCardLayout,'trend-card-small-pf': showSmallCardLayout}">
+
+    <span class="trend-header-pf" ng-if="config.title">{{config.title}}</span>
+
+    <span ng-if="showActualValue">
+      <span class="trend-title-big-pf">{{getLatestValue()}}</span>
+      <span class="trend-title-small-pf">{{config.units}}</span>
+    </span>
+    <span ng-if="showPercentageValue">
```
waldenraines: You could use [`ng-switch`](https://docs.angularjs.org/api/ng/directive/ngSwitch) here instead, especially if we plan on adding additional modes.

test/charts/trends/trends-chart.spec.js
```diff
@@ -55,4 +53,28 @@ describe('Directive: pfTrendsChart', function() {
...
+
+  it("should show large or small trend card layouts", function() {
+    // by default, should show a large card
+    var trendCard = angular.element(element.find('.trend-card-large-pf'));
+    expect(trendCard.size()).toBe(1);
+
+    $scope.config.layout = 'small';
+    $scope.$digest();
+    trendCard = angular.element(element.find('.trend-card-small-pf'));
+    expect(trendCard.size()).toBe(1);
```
waldenraines: Is it worth ensuring that the opposite card isn't being shown here, perhaps by using [`hasClass()`](https://api.jquery.com/hasclass/)?

test/charts/trends/trends-chart.spec.js
```diff
@@ -38,11 +38,9 @@ describe('Directive: pfTrendsChart', function() {
...
 
   var compileChart = function (markup, scope) {
-    element = $compile(angular.element(markup))(scope);
-    scope.$apply();
-    isolateScope = element.isolateScope();
-
-    return element;
+    var el = $compile(markup)(scope);
+    scope.$digest();
+    return el;
```
jeff-phillips-18: Why not return angular.element(el) here so that you don't have to do this each time in the calling functions?

src/charts/trends/trends-chart.directive.js
dtaylor113: When I try:
        const SMALL = 30, LARGE = 60;
...
I get:
src/charts/trends/trends-chart.directive.js
  144:10  error  Unexpected token const
✖ 1 problem (1 error, 0 warnings)
Warning: Task "eslint:target" failed. Use --force to continue.
Aborted due to warnings.

src/charts/trends/trends-chart.html
dtaylor113: Hi, great tip, thanks!  I'm not going to have anymore 'title format' modes, but in my next PR I'm going to introduce two more 'layout' modes ('compact' and 'flat'), ok if I introduce the ng-switch at that time?

src/charts/trends/trends-chart.directive.js
dtaylor113: Fixed

test/charts/trends/trends-chart.spec.js
dtaylor113: ok, added test(s)

test/charts/trends/trends-chart.spec.js
dtaylor113: fixed

dtaylor113:
I will squash PR issue commits before merging.

jeff-phillips-18:
LGTM

waldenraines:
> @waldenraines how do you feel about this practice? In a previous position we did a commit per raised issue to make code reviews easier to follow.

When you say "issue" do you mean a github issue?  If so, I prefer a PR per issue unless the issues are very closely related.

In the case of this PR these updates are from a UX review of the component, correct?  Not sure if this would be too much overhead but it could be good to create a github issue with the things to be changed and then use a commit message format like this:

```
Fixes #12345:  trend chart card updates

This commit enhances the trend chart by doing the following:

1. Added ability to change card title from actual value '76 GB' to percentage value '30% of 250 GB'
2. Added ability to change layout from 'large' card to 'small' card
3. Updated Gruntfile.js: ngdocs:watch to rebuild upon html and css changes
```

Even without creating the github issue I tend to prefer one commit with a list of things fixed in cases like this PR because the changes are all very closely related (except for maybe the gruntfile change, which should probably remain it's own separate commit).  However, I don't feel too strongly about this and am open to other approaches.

dtaylor113:
Hi, I meant one commit per PR comment/conversation.  For example, for this PR I submitted three more commits to address comments/issues raised for this particular PR:

 @dtaylor113	removed hardcoded styles in demo	a26e75a
 @dtaylor113	Removed annotation style dependency injection	6e188b6
 @dtaylor113	Updated unit test	 4bc23db

I feel doing it this way helps the code reviewer find the code which apples to their raised comment.
Before merging, I will squash these commits into the main one so as not to mess up the history with unnecessary details.

erundle:
LGTM

src/charts/trends/trends-chart.directive.js
dtaylor113: @waldenraines any idea why eslint doesn't like constants?

waldenraines: Looks like we need to enabled `blockBindings`.

https://github.com/peol/brackets-eslint/issues/10

waldenraines: See also https://github.com/patternfly/angular-patternfly/pull/86

waldenraines:
> I feel doing it this way helps the code reviewer find the code which apples to their raised comment.
> Before merging, I will squash these commits into the main one so as not to mess up the history with unnecessary details.

That's a good point, I'm okay with doing it this way.

src/charts/trends/trends-chart.directive.js
dtaylor113: Hi, I tried using the 'const' keyword, but our PhantomJS/Karma Unit Test browers throws the following error:
```
PhantomJS 1.9.8 (Linux) ERROR
  SyntaxError: Parse error
  at /home/dtaylor/Dev/repos/angular-patternfly/src/charts/trends/trends-chart.directive.js:132

Warning: Task "karma:unit" failed. Use --force to continue.

Aborted due to warnings.
~/Dev/repos/angular-patternfly>node /home/dtaylor/Dev/repos/angular-patternfly/src/charts/trends/trends-chart.directive.js

/home/dtaylor/Dev/repos/angular-patternfly/src/charts/trends/trends-chart.directive.js:132
    const SMALL = 30, LARGE = 60;
    ^^^^^
SyntaxError: Use of const in strict mode.
    at Module._compile (module.js:439:25)
```
Which lead me to read about how '...brower support is heavily fractionary'.  So, I just used a 'var' until I can figure out how to implement 'const'.

dtaylor113:
@waldenraines if you are happy with the changes, please merge.
ty,
- Dave

PR #83, opened 2015-09-14
Change pfSimpleFilter directive to use two sub-component directives

jeff-phillips-18:
pfSimpleFilter directive needs to be split up so that the components for the fields area, and the components for the current results and active filters area can be placed separately on the toolbar or where ever an application would like them. Currently, this one directive takes up the entire toolbar area and won't allow (easily) for additional toolbar components.


src/filters/simple-filter-fields.html
```diff
@@ -0,0 +1,32 @@
...
+            </li>
+          </ul>
+        </div>
+        <input class="form-control" type="{{currentField.filterType}}" ng-model="config.currentValue"
+               placeholder="{{currentField.placeholder}}"
+               ng-keypress="onValueKeyPress($event)"
+               ng-if="currentField.filterType !== 'select'"/>
+        <select pf-select class="form-control filter-select" id="currentValue"
+                ng-model="config.currentValue"
+                ng-options="o as o for o in currentField.filterValues"
```
waldenraines: Is the "o as o" part needed?  I would have thought the `for o in currentField.filterValues` was sufficient.

jeff-phillips-18: It needs to have the select and the value but not the as. I agree this would be more readable if simply o was not used. I will update.

src/views/toolbar/data-toolbar-directive.js
```diff
@@ -231,7 +231,33 @@ angular.module('patternfly.views').directive('pfDataToolbar',
         $scope.checkViewDisabled = function (view) {
           return $scope.config.viewsConfig.checkViewDisabled && $scope.config.viewsConfig.checkViewDisabled(view);
         };
+
+        $scope.filterExists = function (filter) {
+          var found = false;
+          $scope.config.filterConfig.appliedFilters.forEach(function (nextFilter) {
```
waldenraines: Seems like we could use [`_.where`](https://lodash.com/docs#where) here to simplify this.

jeff-phillips-18: Yep. Thanks!

grunt-ngdocs-index.tmpl
```diff
@@ -31,7 +31,7 @@
       }
 
       addTag('base', {href: baseUrl});
-      addTag('link', {rel: 'stylesheet', href: 'css/prettify.css', type: 'text/css'});
+      // addTag('link', {rel: 'stylesheet', href: 'css/prettify.css', type: 'text/css'});
```
waldenraines: Is this change intentional?

jeff-phillips-18: Yes, sorry I squashed a commit that explained it. The css/prettify.css changes some sizing causing the filter components to not line up as they would if just using patternfly.

waldenraines: Should we just remove this instead of commenting it out then?

jeff-phillips-18: I left it incase we noticed that parts of the page (not patternfly components) didn't look right, we would have the hint as to where settings might be.

waldenraines: In that case I think we should add a comment above as to why it is commented out and still here.

jeff-phillips-18: done

dtaylor113:
LGTM

PR #84, opened 2015-09-14
Add ngdocs template file, remove overrides css file

jeff-phillips-18:
By modifying the default ngdocs template file to remove the inclusion of bootstrap css file we no longer need the overrides css file. The ngdocs default template was using an old version of bootstrap which was causing the conflicts.


waldenraines:
LGTM :+1:

dtaylor113:
LGTM

PR #87, opened 2015-09-15
Enabling CI

erundle:
Trying to get this repo hooked up to an jenkins build and deploying:

http://angularpf-erundle.rhcloud.com/

Karma tests are currently not executing so I am working on that.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -229,6 +229,12 @@ module.exports = function (grunt) {
       grunt.task.run(['clean', 'lint', 'test', 'ngtemplates', 'concat', 'ngAnnotate', 'uglify:build', 'cssmin', 'copy', 'ngdocs', 'clean:templates']);
     });
 
+    // Runs all the tasks of build with the exception of tests
+    grunt.registerTask('deploy', 'Prepares the project for deployment. Does not run unit tests', function () {
+      var concatSrc = 'src/**/*.js';
+      grunt.task.run(['clean', 'lint', 'ngtemplates', 'concat', 'ngAnnotate', 'uglify:build', 'cssmin', 'copy', 'ngdocs', 'clean:templates']);
+    });
+
     grunt.registerTask('default', ['build']);
     grunt.registerTask('ngdocs:view', ['build', 'connect:docs', 'watch']);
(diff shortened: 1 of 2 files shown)


Gruntfile.js
```diff
@@ -229,6 +229,12 @@ module.exports = function (grunt) {
       grunt.task.run(['clean', 'lint', 'test', 'ngtemplates', 'concat', 'ngAnnotate', 'uglify:build', 'cssmin', 'copy', 'ngdocs', 'clean:templates']);
     });
 
+    // Runs all the tasks of build with the exception of tests
+    grunt.registerTask('deploy', 'Create bootstrap build files', function () {
```
waldenraines: This task description should be updated.

server.js
```diff
@@ -1,34 +1,27 @@
 'use strict';
 
-var fs =require('fs');		//for image upload file handling
+var fs =require('fs');    //for image upload file handling
```
waldenraines: Is this even needed?

package.json
```diff
@@ -10,6 +10,7 @@
     "grunt": "0.4.5",
     "grunt-angular-templates": "0.5.7",
     "grunt-available-tasks": "0.4.3",
+    "grunt-cli" : "~0.1.13",
```
waldenraines: Is this required because we cannot install `grunt-cli` globally on the openshift box?  Is there a way to install this by configuring openshift appropriately?

erundle: I will dig on it a bit more but yes I added this to get around the permission issue and I am using that version to actually execute the grunt build (hence to why unit tests are failing). Right now I am using a node cartridge on the openshift online version since I can't get a jenkins instance running on the itos version. I opened a support ticket. The mean stack cartridges seem to be having issues as well.

dtaylor113:
LGTM

Gruntfile.js
erundle: Updated

server.js
erundle: Removed. Not needed

package.json
erundle: Was able to configure jenkins to get around this so this is removed

jeff-phillips-18:
LGTM

PR #88, opened 2015-09-15
This Fixes #85  'angular.merge is not a function .controller'

### LINKED ISSUE #85: angular.merge is not a function .controller
Downloaded the latest, saw the above error when employing pfSimpleFilter / pfDataToolbar.  Working on the issue presently, will provide additional information as the issue becomes clearer.

**more info**
the below commit does away with angular extend in lieu of merge, this is presently not supported by angular > 1.4  (ie 1.3.18)
https://github.com/patternfly/angular-patternfly/commit/be9c015fd4aadeeef48ad195df44838179e02967

dtaylor113:


waldenraines:
Need some tests for `pfUtils`.

waldenraines:
Also, why don't we just use [`angular.extend()`](https://docs.angularjs.org/api/ng/function/angular.extend)

jeff-phillips-18:
The extendDeep copy does not clone the second argument's (defaultConfig) fields when copying to the dst object. This causes the defaultConfig to be modified if the dest object is modified for fields that are objects or arrays.

dtaylor113:
"Also, why don't we just use angular.extend() instead of angular.merge()?"

angular.extend() doesn't do a deep copy.

dtaylor113:
"Need some tests for pfUtils."
There chart and view objects which use pfUtils do have unit tests which verify pfUtils is working correctly.

waldenraines:
> There chart and view objects which use pfUtils do have unit tests which verify pfUtils is working correctly.
No they don't, they verify that `angular.merge()` is working correctly (or in the case of 1.3, `_.merge`).  If we're going to keep this approach it needs to be tested, especially this new `extendDeep()` function.

waldenraines:
> angular.extend() doesn't do a deep copy.

Then what about `_.extend()`?  Since we're requiring underscore can't we just use that and get rid of the need for `pfUtils`?

dtaylor113:
Underscore extend is a shallow copy, we need a deep copy.  pfUtils is need for those customers who are < angular-js 1.4 and are using underscore (not lodash).  In this case, we need to make our own 'extendDeep' function.

src/utils/utils.js
```diff
@@ -0,0 +1,48 @@
+(function () {
+  'use strict';
+
+  angular.module('patternfly.utils').constant('pfUtils', {
+    combine: function (source1, source2) {
+      var retValue;
```
erundle: Would prefer to see this called merge but otherwise LGTM

waldenraines:
> Underscore extend is a shallow copy, we need a deep copy. pfUtils is need for those customers who are < angular-js 1.4 and are using underscore (not lodash). In this case, we need to make our own 'extendDeep' function.

How long will we support < angular 1.4?  Should we file an issue to get rid of pfUtils at some point in the future?  I don't think we want to support this indefinitely and it introduces extra dependencies that will not be needed once people upgrade.

dtaylor113:
Hi, I believe my latest push addressed all issues.
If so, @jeff-phillips-18 please merge
-thanks

dtaylor113:
Hi @jeff-phillips-18, latest push address you last comment  -thanks

PR #89, opened 2015-09-16
Progress chart

erundle:
Directive to create a basic progress chart


src/charts/progress/progress-chart.directive.js
```diff
@@ -0,0 +1,72 @@
...
+ * <ul style='list-style-type: none'>
+ * <li>.used   - number representing the amount used
+ * <li>.total  - number representing the total amount
+ * </ul>
+ *
+ * @param {object} config the chart configuration options<br/>
+ * <ul style='list-style-type: none'>
+ * <li>.title  - the chart title
+ * <li>.units  - units to be displayed on the chart
+ * </ul>
```
jeff-phillips-18: I think a config object is overkill here. Can we just use title and units parameters?

erundle: Sure, will make that change.

test/charts/progress/progress.spec.js
```diff
@@ -0,0 +1,49 @@
...
+    expect(progressBar).toBe("50%");
+
+    title = angular.element(element).find('.progress-description').html();
+    expect(title).toBe("CPU Usage");
+
+    subTitle = angular.element(element).find('.progress-bar span').html();
+    expect(subTitle).toBe("8 of 16 MB");
+  });
+
+});
```
jeff-phillips-18: Could we add a test to verify updating the data, title or units updates the displayed components? Silly and easy, but it adds check.

test/charts/progress/progress.spec.js
```diff
@@ -0,0 +1,49 @@
...
+
+  beforeEach(module(
+    'patternfly.charts',
+    'charts/progress/progress-chart.html'
+  ));
+
+  beforeEach(inject(function(_$compile_, _$rootScope_, _$timeout_) {
+    $compile = _$compile_;
+    $scope = _$rootScope_;
+    $timeout = _$timeout_;
```
jeff-phillips-18: I don't see $timeout being used

serenamarie125:
Please hold off on this PR.  We need design and high fidelity first, and then evaluate if implementation matches.

PR #90, opened 2015-09-16
Add sort and actions to pfDataToolbar directive

jeff-phillips-18:
Added pfSimpleSort to the data toolbar
Added actions to the data toolbar
Updated toolbar styling based on latest Toolbar additions to Patternfly
    
Currently using a specific Patternfly commit as a dependency, this will need to be updated when patternfly/patternfly releases next and before we create a patternfly/angular-patternfly release


bower.json
```diff
@@ -40,6 +40,6 @@
     "angular-route": "1.3.0 - 1.4.*",
     "angular-bootstrap": "0.13.x",
     "lodash": "3.x",
-    "patternfly": "2.2.0"
+    "patternfly": "https://github.com/patternfly/patternfly.git#41b81ba31dba9b451dc5b57b17a37b8d5d37d3af"
```
waldenraines: Why this change?

jeff-phillips-18: The current toolbar styling in Patternfly has not yet been released. As I put in the PR notes, we will need to update this once PF releases and before we release the next version of angular-patternfly

waldenraines: Sorry I missed that in the notes.

dtaylor113:
LGTM

src/views/toolbar/data-toolbar.html
```diff
@@ -1,15 +1,51 @@
...
+                  ng-disabled="action.isDisabled === true">
+            {{action.name}}
+          </button>
+          <div class="dropdown btn-group" ng-if="config.actionsConfig.moreActions && config.actionsConfig.moreActions.length > 0">
+            <button type="button" class="btn btn-link dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+              <span class="fa fa-ellipsis-v"></span>
+            </button>
+            <ul class="dropdown-menu">
+              <li ng-repeat="action in config.actionsConfig.moreActions"
+                  role="{{action.isSeparator === true ? 'separator' : ''}}"
```
waldenraines: Nitpick but shouldn't the role be something like `menuitem`?

waldenraines:
LGTM pending rebase.

src/views/toolbar/data-toolbar.html
jeff-phillips-18: fixed

PR #91, opened 2015-09-16
Trend chart new 'compact' and 'inline' layouts

dtaylor113:
Compact Layout:

image

Flat Layout:

image


jeff-phillips-18:
LGTM

waldenraines:
Selecting "actual" doesn't seem to work for me, I always see percentage:

screen shot 2015-09-16 at 3 21 29 pm

waldenraines:
Is the layout name "flat" open to changes?  I don't see what is flat about this layout.  Maybe "inline"?

dtaylor113:
@erundle please merge this when/if you are happy with it.
Thanks,
 - Dave

waldenraines:
> Actual .vs Percent currently only work for 'large', 'small', and 'compact' layouts. Currently, 'percentage' format is the only one officially approved for the 'flat' layout, but this may change in the future.

Then we should disabled this option when flat is selected.  Otherwise it is quite confusing.

dtaylor113:
Ok, Actual .vs Percent is hidden when layout === 'inline'.
Please merge if happy with this PR.
Thanks,
- Dave

PR #92, opened 2015-09-16
Improving look and feel of documentation

erundle:


Gruntfile.js
```diff
@@ -106,6 +106,7 @@ module.exports = function (grunt) {
         options: {
           title: 'Angular-Patternfly Documentation',
           dest: 'dist/docs',
+          imageLink: 'https://rawgit.com/patternfly/patternfly/master/dist/img/logo-alt.svg',
```
erundle: Anyone care that I am linking out to this or do you want me to add it to the project?

erundle:
@jeff-phillips-18 anything else that got messed up? What about search?

dtaylor113:
Looks much better, however I still see titles getting truncated:
image

But not sure if we can make the font any smaller.

jeff-phillips-18:
The search before had a border radius not square.
I'd like to see the search and breadcrumbs be part of the fixed navbar area rather than scroll off. It's not how it used to work, just how it ought to work.

Gruntfile.js
dtaylor113: Hmm...I don't see this logo anywhere when I do 'grunt ngdocs:view'

erundle:
We can't move the search up. But I can look at making it fixed if that is what you really want. PF also doesn't really use border radius so I would not recommend adding that. As far as the truncation the page should now be fluid so pending you have enough room we it should render fine.

dtaylor113:
So in your setup those titles are not truncated?  They are truncated in my
laptop and external monitor which is set to a high resolution.

On Wed, Sep 16, 2015 at 3:53 PM, Eleni Rundle <notifications@github.com>
wrote:

> We can't move the search up. But I can look at making it fixed if that is
> what you really want. PF also doesn't really use border radius so I would
> not recommend adding that. As far as the truncation the page should now be
> fluid so pending you have enough room we it should render fine.
>
> —
> <https://github.com/patternfly/angular-patternfly/pull/92#issuecomment-140868274>
> .
>



-- 
Dave Taylor
Principle Software Engineer
(978) 392 - 1016

erundle:
I pulled down an old branch. The search and breadcrumb were never fixed. They always scrolled with the entire page. @dtaylor113 if you had your server running already stop it and do ngdocs:view again

jeff-phillips-18:
Yeah, like I said, it's now how it used to work, just how it ought to work.

dtaylor113:
Hi, I did, I still see patternfly.filte..., patternfly.fitle..., and patternfly.notif...  -the rest are fine.

@jeff-phillips-18 noticed that the ng-docs for pfSimpleFilter has 'patternfly.***fit***lers'  typo

erundle:
screen shot 2015-09-16 at 4 07 05 pm

dtaylor113:
My bad, I was still on 'master' branch, switched to 'ngdocs' and everything looks great now.

dtaylor113:
@erundle could you please change ng-docs for pfSimpleFilter to 'patternfly.filters', thanks

erundle:
Typo fixed and changes squashed

dtaylor113:
LGTM

Gruntfile.js
waldenraines: > Hmm...I don't see this logo anywhere when I do 'grunt ngdocs:view'

Top left corner.



> Anyone care that I am linking out to this or do you want me to add it to the project?

I think we should probably add it unless we are fairly confident that it's not going to be renamed, etc.

waldenraines:
> Search worked for me.
> 
> Dave

The search seems to mostly work for me.  Here are a couple of test cases:

 - "pfDataList", "dataList", "pfData" - no results
 - "pf-data-list", "data-list", "pf-data" - results

It seems like it's not searching on the directive name but only on the contents of the documentation?

waldenraines:
The breadcrumbs don't work, not sure if that was the case before.

waldenraines:
Overall this is a very nice change, good work @erundle!

jeff-phillips-18:
Breadcrumbs are working fine for me, but kind of useless as there is no doc for the modules or overall API Documentation.

jeff-phillips-18:
LGTM

erundle:
Logo added locally to the project and changes squashed. Please go ahead and merge. 

Yes the issue with the breadcrumbs is that we have no documentation on the module level so that is something we may want to address in the future :). Search is and has always been going against the entire document not just the module name. So if we want I can add that as an issue to address at a later date.

waldenraines:
> So if we want I can add that as an issue to address at a later date.

If you don't mind that would be good.

Otherwise LGTM

erundle:
Logged: https://github.com/patternfly/angular-patternfly/issues/93

PR #94, opened 2015-09-17
Refactored pfTrendsChart to pfTrendChart 

dtaylor113:
Directive will only display a single trend chart.  Trend charts can be stacked within a pfCard to render multiple trends.


dtaylor113:
@erundle @jeff-phillips-18 @waldenraines please review.
This is a change in a directive name, which I know ManageIQ is using, how do I update the release notes?

jeff-phillips-18:
I don't mind calling charting the trend of a single data point "trends". Could be daily trends, weekly trends, monthly trends show for a single data point in a single chart.

I'm not so sure this warrants refactoring since it is already released. We would need to deprecate pfTrendsChart not simply remove it.

dtaylor113:
Ok, keeping the 'Trends' name

PR #95, opened 2015-09-21
Add ability to have actions pulldown menu in pfDataList directive

jeff-phillips-18:
This adds a dropdown menu button to the far right of the items in the data list if actions are provided to the pfDataList directive.


src/views/datalist/data-list-directive.js
```diff
@@ -227,6 +279,80 @@ angular.module('patternfly.views').directive('pfDataList', function (pfUtils) {
...
+
+            if (action.isVisible !== false) {
+              actionHtml += '<li role=\'' + (action.isSeparator === true ? 'separator' : 'menuitem') + '\' ' +
+              'class="' + (action.isSeparator === true ? 'divider ' : '') + (action.isDisabled === true ? 'disabled' : '') + '"> ';
+              if (action.isSeparator !== true) {
+                actionHtml += '<a href="#" ' +
+                '   title="' + action.title + '" ' +
+                '   ng-click="handleAction(\'' + action.name + '\')"> ' +
+                ' ' + action.name +
+                '</a>';
```
waldenraines: Constructing HTML using javascript is difficult to read and maintain and is somewhat of an AngularJS anti-pattern.  We should be using angular-bootstrap's [dropdown directive](https://angular-ui.github.io/bootstrap/#/dropdown), wrapping it with modifications if desired.

src/views/datalist/data-list-directive.js
```diff
@@ -227,6 +279,80 @@ angular.module('patternfly.views').directive('pfDataList', function (pfUtils) {
...
+          };
+
+          $scope.getDropDownHtml = function () {
+            var dropDownHtml = '<ul class="dropdown-menu">';
+
+            $scope.actions.forEach(function (action) {
+              dropDownHtml += $scope.getActionHtml(action);
+            });
+
+            dropDownHtml += '</ul>';
```
waldenraines: See above note

src/views/datalist/data-list-directive.js
```diff
@@ -227,6 +279,80 @@ angular.module('patternfly.views').directive('pfDataList', function (pfUtils) {
             'Illegal use of pfDataList directive! ' +
             'Cannot allow both select box and click selection in the same data list.');
           }
+
+          if (!$scope.listId) {
+            $scope.listId = 'pf-data-list';
```
waldenraines: What happens if there are more than one of these on a page?

src/views/datalist/data-list-directive.js
jeff-phillips-18: The problem with that is that it would create the dropdown for every item in the list. I did this so that the dropdown menu was only created when the item was selected. If there are 1000's of rows the performance would suffer creating all these menus.

src/views/datalist/data-list-directive.js
waldenraines: See comment about constructing HTML in javascript below.  If we were to use angular-bootstrap's [dropdown directive](https://angular-ui.github.io/bootstrap/#/dropdown) then I think the need for `$scope.listId` is removed.

jeff-phillips-18: If there are more than one on a page the app would need to pass in the ID via the listId parameter.

styles/angular-patternfly.css
```diff
@@ -172,6 +172,43 @@
...
+}
+
+.data-list-pf .list-menu .btn-link:hover {
+  color: #0099d3;
+}
+
+.data-list-pf .list-menu.disabled .btn-link,
+.data-list-pf .list-menu.disabled .btn-link:hover {
+  color: #999999;
+  cursor: not-allowed;
```
waldenraines: This isn't applied by [`.dropdown-menu > .disabled`](http://getbootstrap.com/components/#dropdowns-disabled)?

jeff-phillips-18: This is for the link button not the menu item.

src/views/datalist/data-list-directive.js
waldenraines: > The problem with that is that it would create the dropdown for every item in the list.

What about using `ng-if` and the aforementioned directive?

src/views/datalist/data-list-directive.js
waldenraines: I think this needs to be documented at the very least.

styles/angular-patternfly.css
waldenraines: :+1: my mistake.

waldenraines:
Should we use absolutely position the dropdown?  See the attached screenshot.

screen shot 2015-09-21 at 11 45 53 am

src/views/datalist/data-list-directive.js
jeff-phillips-18: That's a good idea I hadn't thought of. I will do that and see how it goes.

jeff-phillips-18: That worked well and simplified the unit tests greatly. Thanks!

src/views/datalist/data-list-directive.js
jeff-phillips-18: Removed this since it is no longer necessary.

dist/angular-patternfly.js
```diff
@@ -3216,15 +3217,22 @@ angular.module('patternfly.validation', []).directive('pfValidation', ["$timeout
...
+ *     <li>.actionFn - (function(action)) Function to invoke when the action selected
+ *     <li>.isVisible - (Boolean) set to true to disable the action
+ *     <li>.isDisabled - (Boolean) set to true to disable the action
+ *     <li>.isSeparator - (Boolean) set to true if this is a placeholder for a separator rather than an action
+ *   </ul>
+ * @param {function (action, item))} updateActionForItemFn function(action, item) Used to update an action based on the current item
  * @example
 <example module="patternfly.views" deps="patternfly.utils">
   <file name="index.html">
     <div ng-controller="ViewCtrl" class="row" style="display:inline-block; width: 100%;">
```
dtaylor113: Think we have a style for this now

dist/angular-patternfly.js
```diff
@@ -3380,20 +3388,63 @@ angular.module('patternfly.validation', []).directive('pfValidation', ["$timeout
...
+          }
+        ];
       }
     ]);
   </file>
 </example>
  */
-angular.module('patternfly.views').directive('pfDataList', ["pfUtils", function (pfUtils) {
+angular.module('patternfly.views').directive('pfDataList',
+  ["$compile", "pfUtils", function ($compile, pfUtils) {
```
dtaylor113: ng-annotate?

jeff-phillips-18: This is in dist, post ng-annotate.

dtaylor113:
LGTM

waldenraines:
> Should we use absolutely position the dropdown? See the dropdown barely visible at the bottom of the attached screenshot.

@jeff-phillips-18 thoughts on this?

jeff-phillips-18:
@waldenraines Yeah, I agree and I'm working on it.

dist/styles/angular-patternfly.css
```diff
@@ -172,6 +172,43 @@
...
 
+.data-list-pf .list-menu {
+  bottom: 0px;
+  position: absolute;
+  right: -20px;
+  top: 0px;
+}
+
+.data-list-pf .list-menu .btn-link {
+  color: #222222;
```
erundle: perfect chance to save some bytes #222 :)

jeff-phillips-18:
@waldenraines I've got it working. The only issue I'm seeing is that the dropdown menu will stay put if you scroll the window while it is up. That is, if I click the link button to see the menu, then I use the wheelie mouse to scroll the page, the dropdown menu does not move with the item it was displayed for. Any thoughts? Acceptable?

erundle:
This solution is going to have problems if the menu is rendered near the bottom of the page. 
screen shot 2015-09-21 at 2 35 33 pm

erundle:
lol nm looks like @waldenraines is already on it.

erundle:
We need to have logic that figures out if it is close to the bottom of the page and if so have it show up and not down.

waldenraines:
> The only issue I'm seeing is that the dropdown menu will stay put if you scroll the window while it is up. That is, if I click the link button to see the menu, then I use the wheelie mouse to scroll the page, the dropdown menu does not move with the item it was displayed for. Any thoughts? Acceptable?

This isn't yet committed is it?  I was going to test this out to see it in action but I still see the same behavior.  

If I understand correctly, I wonder if we should just close any open dropdowns on scroll?

jeff-phillips-18:
I have not yet pushed. I will so you can see what I mean.

dist/styles/angular-patternfly.css
```diff
@@ -172,6 +172,46 @@
...
+  color: #999999;
+  cursor: not-allowed;
+}
+
+.data-list-pf .list-menu .dropdown-menu {
+  position: fixed;
+}
+
+.data-list-pf .list-menu ul {
+  cursor: pointer;
```
erundle: instead of ul use .dropdown-menu it is more performant

src/views/datalist/data-list-directive.js
```diff
@@ -186,25 +194,104 @@
...
+            var top = buttonRect.top + buttonRect.height;
+            var left = buttonRect.left + buttonRect.width - menuRect.width;
+            var docHeight = $window.innerHeight;
+
+            if (top + menuRect.height > docHeight) {
+              top = docHeight - menuRect.height;
+            }
+
+            dropMenu.style.top = top + "px";
+            dropMenu.style.left = left + "px";
```
erundle: based on where you are on the page you may need to set dropMenu.style.bottom instead

waldenraines:
@jeff-phillips-18 yeah I think closing the menu on scroll would be the best bet here.  I am open to other suggestions though.

erundle:
So what happens if the data list is wider than the space that is available. Does the page scroll, datalist wrapper? Do we just lose the actions menu?

jeff-phillips-18:
Yeah, I think closing the menu on scroll is the best option as well. Just added a commit for that.

I will squash commits when ready before merge.

src/views/datalist/data-list-directive.js
```diff
@@ -186,25 +212,115 @@
...
+          }
+        ];
       }
     ]);
   </file>
 </example>
  */
-angular.module('patternfly.views').directive('pfDataList', function (pfUtils) {
+angular.module('patternfly.views').directive('pfDataList',
+  function ($compile, $timeout, $window, pfUtils) {
```
waldenraines: `$compile` is no longer used.

test/views/datalist/data-list.spec.js
```diff
@@ -1,8 +1,9 @@
 describe('Directive:  pfDataList', function () {
   var $scope;
   var $compile;
```
waldenraines: `$compile` is no longer used.

waldenraines:
> So what happens if the data list is wider than the space that is available. Does the page scroll, datalist wrapper? Do we just lose the actions menu?

@jeff-phillips-18 what about @erundle's concern?

test/views/datalist/data-list.spec.js
jeff-phillips-18: It is still used in the compileHTML function

jeff-phillips-18:
@erundle @waldenraines Data list is designed to not scroll horizontally. That is the pattern's design.

test/views/datalist/data-list.spec.js
waldenraines: My mistake, disregard.

src/views/datalist/data-list-directive.js
waldenraines: This one still isn't used.

jeff-phillips-18: Fixed

jeff-phillips-18:
Squashed.

src/views/datalist/data-list.html
```diff
@@ -1,17 +1,31 @@
...
-                 ng-dblclick="dblClick($event, item)">
-            </div>
-            <div class="list-check-box" ng-if="config.showSelectBox" style="padding-top: {{(config.rowHeight - 32) / 2}}px;">
-                <input type="checkbox" value="item.selected" ng-model="item.selected" ng-disabled="checkDisabled(item)" ng-change="checkBoxChange(item)"/>
-            </div>
-        </div>
+  <div class="list-group-item"
+     ng-repeat="item in items track by $index"
+     ng-class="{'pf-selectable': selectItems, 'active': isSelected(item), 'disabled': checkDisabled(item)}">
+    <div class="row list-row">
```
erundle: FYI I am looking at this directive in self-service and we are using class="row" too much. Everytime that is used we get margin: 0 -20px. This is causing scrollbars. Please only use the class of row if it is directly inside a container or container-fluid div

jeff-phillips-18: In this particular case, row is needed as we use the margin areas for the checkbox and actions menu.

erundle: Ok well this margin pulls things closer to the edge of the page. The way you have things currently structured is you are making this element wider than the containing parent div. Maybe your PR will address this but if not we need to look at restructuring somethings. Just putting overflow-x:hidden isn't the solution. We need to stop it from growing wider than its container to begin with.

erundle: If you add this: .data-list-pf .list-group-item { padding: 10px 0 }; and remove the row class from that line things seem to render better.

jeff-phillips-18: Are you doing something different in your parent container? Making this change causes the list example in ngdoc unreadable.

jeff-phillips-18: OK, I found the issue and removed the row class.

jeff-phillips-18:
@erundle Are you ok with that change?

erundle:
Yes will merge

PR #100, opened 2015-09-24
pfCard: added footer action bar and filter, and Fixed #99

### LINKED ISSUE #99: Ng-Docs 'docs.css' overriding/conflicting with PatternFly styles
For example, if I look at:
http://angular-patternfly.rhcloud.com/#/api/patternfly.card.directive:pfCard

Inspect "My Card Title", I see that '.content h2' selectors are overriding 'pf' selectors:

.content h2 {
  font-size: 24px;
  line-height: 36px;
}

.content h2, .content h3, .content h4, .content h5 {
  margin: 1em 0 5px;
}
.card-pf-title {
  ~~font-size: 16px;~~
  font-weight: 400;
  ~~margin: 20px 0;~~
  padding: 0;
}

dtaylor113:
image


src/card/basic/card.directive.js
```diff
@@ -3,26 +3,140 @@
...
  * @restrict A
  * @element ANY
- * @param {headTitle=} Title for the card - required
- * @param {subTitle=} Subtitle for the card - optional
- * @param {showTopBorder=} Show Top Border, true shows top border, false (default) hides top border - optional
- *
+ * @param {string} headTitle Title for the card
+ * @param {string=} subTitle Sub-Title for the card
+ * @param {boolean=} showTopBorder Show/Hide the blue top border. True shows top border, false (default) hides top border
+ * @param {boolean=} showBottomBorder Show/Hide the bottom grey line between the title and sub-title.
```
jeff-phillips-18: Might want to call this showTitlesSeparator or something. I expected this to mean show a border at the bottom of the card.

src/card/basic/card.directive.js
```diff
@@ -35,7 +149,32 @@ angular.module('patternfly.card').directive('pfCard', function () {
...
+      if ($scope.filter && !$scope.currentFilter) {
+        if ($scope.filter.defaultFilter) {
+          $scope.currentFilter = $scope.filter.filters[$scope.filter.defaultFilter];
+        } else {
+          $scope.currentFilter = $scope.filter.filters[0];
+        }
+      }
+
+      $scope.footerCallBackFn = function () {
+        $scope.footerCallBackResult = $scope.footer.callBackFn();
```
jeff-phillips-18: I know this won't get called if $scope.footer.callBackFn is not defined, but you might want to check just in case.

src/card/basic/card.html
```diff
@@ -1,9 +1,41 @@
...
+        <a ng-if="footer.callBackFn && !footer.href" ng-click="footerCallBackFn()">
+          <span class="{{footer.iconClass}} card-pf-footer-text" ng-if="footer.iconClass"></span>
+          <span class="card-pf-footer-text" ng-if="footer.text">{{footer.text}}</span>
+        </a>
+        <span ng-if="!footer.href && !footer.callBackFn">
+          <span class="{{footer.iconClass}} card-pf-footer-text" ng-if="footer.iconClass"></span>
+          <span class="card-pf-footer-text" ng-if="footer.text">{{footer.text}}</span>
+        </span>
+      </div>
+      <div class="card-pf-cell text-right" ng-if="filter.filters && (!filter.position || filter.position === 'footer')">
```
jeff-phillips-18: Should this check "filter" before "filter.filters"?

jeff-phillips-18:
LGTM

src/card/basic/card.directive.js
```diff
@@ -3,26 +3,140 @@
...
+ *                    {label:'Today', value:'today'}]</pre>
+ * <li>.defaultFilter - integer, 0 based index into the filters array
+ * <li>.callBackFn - user defined function to call when a filter is selected
+ * </ul>
  * @description
  * Directive for easily displaying a card with html content
  *
  * @example
- <example module="patternfly.card">
+ <example module="test">
```
waldenraines: test?

src/card/basic/card.html
```diff
@@ -1,9 +1,41 @@
 <div ng-class="showTopBorder === 'true' ? 'card-pf card-pf-accented' : 'card-pf'">
-  <div class="card-pf-heading">
-    <h2 class="card-pf-title">{{headTitle}}</h2>
+  <div ng-class="!showTitlesSeparator || showTitlesSeparator === 'true' ? 'card-pf-heading' : 'card-pf-heading-no-bottom'">
+    <div class="card-pf-table">
+      <div class="card-pf-cell">
+        <h2 class="card-pf-title" ng-class="{'card-pf-filter-header': showFilterInHeader()}">{{headTitle}}</h2>
+      </div>
+      <div class="card-pf-cell text-right" style="padding-right: 0px" ng-if="showFilterInHeader()">
```
waldenraines: Can we get this inline style out of here?

src/charts/trends/trends-chart.directive.js
```diff
@@ -82,10 +82,17 @@
...
+       <div pf-card head-title="Cluster Utilization" show-top-border="true"
+            footer="footerConfig" filter="filterConfig">
+         <div pf-trends-chart config="config2" chart-data="data"></div>
+       </div>
+     </div>
+    </div>
  </file>
  <file name="script.js">
- angular.module( 'patternfly.charts' ).controller( 'ChartCtrl', function( $scope ) {
+ angular.module( 'test', ['patternfly.charts', 'patternfly.card'] ).controller( 'ChartCtrl', function( $scope ) {
```
waldenraines: Why the change to the test module?

src/card/basic/card.directive.js
dtaylor113: just a dummy name for the ngDoc example

src/charts/trends/trends-chart.directive.js
dtaylor113: For the ngDoc examples, whenever two directives are needed, they need to be listed in the module dependencies.  In this case, I wrapped a trends chart inside a card.

src/card/basic/card.html
dtaylor113: Ok, removed, wasn't needed

dtaylor113:
@waldenraines, ok to merge?

src/card/basic/card.directive.js
waldenraines: Why do all of the other examples use the module name of the directive?  Is there a reason this couldn't be used here?

src/card/basic/card.directive.js
```diff
@@ -3,26 +3,140 @@
...
+     <div pf-card head-title="Card Title" sub-title="Card Subtitle" show-top-border="true"
+          footer="footerConfig" filter="filterConfig" style="width: 50%">
+       [Card Contents] <button>Click Me</button> Footer with Link & Timeframe filter
+     </div>
 
-    <div pf-card head-title="Card With Top Border" sub-title="My card subtitle" show-top-border="true">
-      <button>Click Me</button>
-    </div>
+     <div pf-card head-title="Performance" sub-title="Last 30 Days" show-top-border="false"
+          show-titles-separator="false" style="width: 65%" footer="actionBarConfig">
```
waldenraines: Should we move this to `demo.css`?

src/card/basic/card.directive.js
```diff
@@ -3,26 +3,140 @@
...
-    <div pf-card head-title="My Card Title" sub-title="My card subtitle">
-      <button>Click Me</button>
-    </div>
+   <div ng-controller="ChartCtrl">
+     <div pf-card head-title="Card Title" sub-title="Card Subtitle" filter="filterConfigHeader" style="width: 50%">
+       [Card Contents] <button>Click Me</button> Timeframe filter in header
+     </div>
+
+     <div pf-card head-title="Card Title" sub-title="Card Subtitle" show-top-border="true"
+          footer="footerConfig" filter="filterConfig" style="width: 50%">
```
waldenraines: Should we move this inline style to `demo.css`?

waldenraines:
Clicking "add new cluster" here doesn't do anything, not sure if that's desired:

screen shot 2015-09-25 at 10 15 48 am

src/card/basic/card.directive.js
dtaylor113: I thought it would be confusing since this ngDoc example isn't just using the one directive, but two.  For ngDoc examples, the module name can be anything, it's just for the ngDoc, it's like a third party app.   'test' probably isn't the best name either, I'll change it to 'demo' so it's clearer.

src/card/basic/card.directive.js
dtaylor113: No, I think that would be overkill, it's not a standard, nor do I think that particular width would be reusable by other demo examples.

src/card/basic/card.directive.js
dtaylor113: Same reason as above, it's just for this ngDoc example.  Each example may have different layouts, widths, etc..

waldenraines:
I am not a fan of the way this footer is declared with a config object.  In my opinion the footer should be it's own directive and be transclude the content of the directive.  Something like this:

```html
<div pf-card head-title="Performance" sub-title="Last 30 Days" show-top-border="false" show-titles-separator="false">
  <div pf-trends-chart config="configVirtual" chart-data="dataVirtual"></div>
  <div pf-trends-chart config="configPhysical" chart-data="dataPhysical"></div>
  <div pf-trends-chart config="configMemory" chart-data="dataMemory"></div>
  <div pf-card-footer>
    <i class="some-class"></i>
    <span>Whatever I want</span>
  </div>
</div>
```

The way we're doing it now is an anti-pattern and is not flexible.  I believe the footer should be created declaratively instead of using some arbitrary configuration object.  What happens when someone wants to put custom HTML into a footer instead of just text?

src/card/basic/card.directive.js
waldenraines: :+1: sounds good, thanks.

src/charts/trends/trends-chart.directive.js
dtaylor113: Changed module name from 'test' to 'demo'

dtaylor113:
@waldenraines  Hi, does my request to file a separate issues for the customization make sense?  If so, can you please merge this -thanks

waldenraines:
> @waldenraines Hi, does my request to file a separate issues for the customization make sense? If so, can you please merge this -thanks

I never saw that request because you edited my comment instead of posting a new one.  In the future, you can highlight text and hit the "r" button to quote like I am doing above.

See https://github.com/patternfly/angular-patternfly/issues/105 for the requested issue.

waldenraines:
I still feel as if we could accomplish what has been designed more elegantly by using my suggestion but since we will already have to update the header similarly I am fine with you merging this.

dtaylor113:
> I never saw that request because you edited my comment instead of posting a new one (and thus no email notification). In the future, you can highlight text and hit the "r" key to quote like I am doing above.

Ah..good to know, was wondering about that.  sometimes I see "Add a Note Inline" button, but other times not.   Anyway, if I have addressed all your issues, can you please merge?  -thanks

waldenraines:
> sometimes I see "Add a Note Inline" button, but other times not.

Yeah, GH has the "conversation view" only for comments on code and not for general comments.  Would be nice if they had it for both.

PR #101, opened 2015-09-24
Update example code for c3-chart directive

jeff-phillips-18:
A consumer had an issue with the c3-chart directive’s example code.
1. There was an extra setting of donut options in the data field.
2. Updated the config styling to make it easier to see the fields.


dtaylor113:
LGTM

PR #104, opened 2015-09-25
Remove href="#" from html A tags to avoid unwanted redirections

jeff-phillips-18:
Using href="#" causes a redirect when  html5mode is on.
Using href="" causes a redirect when html5mode is off.
Removing it completely doesn't cause a redirect in either case.


waldenraines:
Will the `<a>`s get `cursor:pointer;` without the `href`?

dtaylor113:
LGTM

jeff-phillips-18:
Ugh, @waldenraines you are correct, we lost the pointer. I will fix it.

jeff-phillips-18:
Add CSS to put back missing pointer cursors

styles/angular-patternfly.css
```diff
@@ -413,3 +413,11 @@
   border-color: #0076b7 !important;
   color: #fff !important;
 }
+
+.toolbar-pf-actions .toolbar-pf-view-selector a {
+  cursor: pointer;
+}
+
+.toolbar-pf-actions .dropdown-menu a {
+  cursor: pointer;
```
waldenraines: Thoughts about making an `<a>` always have `cursor: pointer;` not just within the toolbar?

jeff-phillips-18:
I'm hesitant to make that broad of a stroke.

waldenraines:
> I'm hesitant to make that broad of a stroke.

Yeah, I hear ya, but I wonder if there is any valid reason to use an `<a>` that is not a link.

I won't hold this up on this issue but it may be something to revisit if we find ourselves having to do this elsewhere.

waldenraines:
@jeff-phillips-18 this will need a rebase now because I just merged #100, sorry about that.

jeff-phillips-18:
Rebased and squashed

PR #113, opened 2015-09-30
Fixing documentation css

erundle:
Order matters :)


waldenraines:
Does the order matter because we have two rules with the same specificity?  Should we alter the specificity?  Do we need both `pf-docs.css` and `demo.css` or can they be combined?  When should each be used?

Sorry for all the questions.

erundle:
So order mattered since pf-docs was presently in the ngdocs/modules folder. We copied it to our project and deleted some CSS that conflicted with base patternfly. The demo.css file has some overrides to the stock ncdoc.css so when they added the new file and added it after the demo.css those overrides weren't being picked up. So I took all ngdoc "framework" styles out of demo.css and put them into pf-doc.css and added them where appropriate. So long story short pf-docs.css contains "ng-doc framework" stuff and demo.css contains any CSS that is related to our example code. If you think we should change the names of the files please let me know.

dtaylor113:
Yes, I was going to suggest renaming the file.  How about changing 'pf-doc' to 'ng-doc.css'
Thanks,
- Dave

dtaylor113:
Other than that, LGTM

waldenraines:
@erundle thanks for the explanation, LGTM.

erundle:
Squashed and renamed files to ng-docs.css and examples.css.

PR #117, opened 2015-10-01
Bump version to 2.3.1

jeff-phillips-18:


dtaylor113:
This makes angular-patternfly version different from patternfly base:

"patternfly": "~2.3.0"

If this is ok, then LGTM

erundle:
Yes we don't really have a choice. That is why the readme says you need to be on the same major and minor version. Bug fix versions can be different. Which is why I suggested "~2.3.0" so this project will also pick up and bug fixes in the base reference impl

PR #120, opened 2015-10-06
Example Pages: Charts & Cards

dtaylor113:
Angular-PatternFly Examples, similar to the test pages in PatternFly Ref. Impl.:

image

'grunt ngdocs:view'

Change url to:
http://127.0.0.1:8000/examples

- Good place to test and demo components and widgets
- Better place to develop than ng-docs commented out code
- Not cross-referencing or publicizing yet until flushed out some more
- Next steps would be to add Data List, Data Toolbar, Filtering, Sorting, etc..

### DIFF
--- a/misc/examples.css
+++ b/misc/examples.css
@@ -87,3 +87,38 @@ hr {
 .actions-label {
   margin-top: 5px;
 }
+
+.examples-container {
+  margin-left: 10px;
+  margin-right: 10px;
+}
+
+.examples-container .row {
+  margin-top: 10px;
+}
+
+.examples-title-lg {
+  font-size: 24px;
+  font-weight: 600;
+}
+
+.examples-title-sm {
+  margin-left: 10px;
+  font-size: 16px;
+  font-weight: 400;
+}
+
+.examples-container .row h4{
+  font-size: 18px;
+  font-weight: 400;
+  font-style: italic;
+}
+
+.examples-container hr {
+  display: block;
+  height: 10px;
+  border: 0;
+  border-top: 1px solid #525252;
+  margin: 1em 0;
+  padding: 0;
+}
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -62,6 +62,12 @@ module.exports = function (grunt) {
           dest: 'dist/docs',
           expand: true
         },
+        examples: {
+          cwd: 'examples/',
+          src: ['*.html', '*.js'],
+          dest: 'dist/docs/examples',
+          expand: true
+        },
         styles: {
           cwd: 'styles/',
           src: ['*.css', '!*.min.css'],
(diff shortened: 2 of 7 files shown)


jeff-phillips-18:
Is this going to be another requirement on new components going forward?
Is this going to be hosted similarly to the ng-docs viewer?
Will there be links between the two?

dtaylor113:
Is this going to be another requirement on new components going forward?
dt> I don't think as ridged as PF ref. impl.
Is this going to be hosted similarly to the ng-docs viewer?
dt> It is running on the same server as ng-docs
Will there be links between the two?
dt> I think eventually there will be.  This examples section still needs to be flushed out some more before that happens though.

This initial PR is just to get the infrastructure in place, I think this can grow into a showcase for our angular components, as well as serve as a much needed place to test/develop new components.

erundle:
So this seems to duplicate effort. I would prefer just to see this in the ngdocs and then the new site. I don't want to need to put components in 3 places.

dtaylor113:
I look at this as a companion to ngdocs, but I am tired of having to develop/test components in ngdocs commented out code  -would much rather have a test bed, like pf ref impl.  Eventually, we can cross-ref from ngdocs to examples, and visa versa.

waldenraines:
What if we had a grunt task to take these examples and put them into the documentation of a component?

I'm a big fan of having html in a .html file and javascript in a .js file and then creating the documentation from that.  It would also allow us to develop components without running `grunt build` which would be really nice in my opinion.

dtaylor113:
> What if we had a grunt task to take these examples and put them into the documentation of a component?
> 
> I'm a big fan of having html in a .html file and javascript in a .js file and then creating the documentation from that. It would also allow us to develop components without running grunt build which would be really nice in my opinion.

With the recent enhancement of adding livereload to to ngdocs grunt task, if you have the examples page(s) loaded and make a change, everything is rebuilt and reloaded for you  -which is really nice!

Not sure about a grunt task to take Examples and put them into ngdocs.  IMO, ngdocs and Examples serve two different goals, ngdocs is for developers to understand all the knobs and settings, whereas Examples is/can be a showcase to give a nice overview of all the angular components.   ngdocs is detailed and for developers, whereas Examples is a catalog so anyone can see what angular-patternfly offers -without having to see/understand all the underlying code.

dtaylor113:
Any movement on this?  For now we can just treat this as a dev only improvement to make it easier to code/test new directives.

jeff-phillips-18:
I'm good with this

misc/examples.css
```diff
@@ -87,3 +87,38 @@ hr {
...
+
+.examples-title-lg {
+  font-size: 24px;
+  font-weight: 600;
+}
+
+.examples-title-sm {
+  margin-left: 10px;
+  font-size: 16px;
+  font-weight: 200;
```
erundle: So I don't believe this font-weights map to anything. Just 300,400,600,700 and 800

dtaylor113: Fixed

jeff-phillips-18:
LGTM

PR #121, opened 2015-10-09
pfnotification service does not make data available to html template

AllenBW:
The related provider should be rewritten, no question about that. Until that time, this fix makes $rootScope.notification available to the pfNotificationList directive.

Previous behavior, where <pf-notification-list></pf-notification-list> was declared, notifications would not appear when the Notification (Notification.error for example) provider was invoked.

Present behavior (with this commit) <pf-notification-list></pf-notification-list> will now display the list of cued notifications no matter where it's declared.

### DIFF
--- a/src/notification/notification.js
+++ b/src/notification/notification.js
@@ -325,6 +325,11 @@ angular.module( 'patternfly.notification' ).directive('pfNotificationList', func
 
   return {
     restrict: 'E',
+    controller: NotificationListController,
     templateUrl: 'notification/notification-list.html'
   };
+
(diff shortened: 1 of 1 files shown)


erundle:
+1 on the rewrite. The notifications service should just be injected into the directive and rootscope should not be involved. As for the fix any reason why you didn't just use an anonymous controller function?

AllenBW:
readability primarily, out of habbit from the product i am most familiar
with at the moment, can do if desired

erundle:
OK LGTM

jeff-phillips-18:
LGTM

PR #123, opened 2015-10-12
Aggregate Status Card - Mini Layout

dtaylor113:
image


dtaylor113:
@erundle , @jeff-phillips-18 @rhamilto   when you get a chance ;-)

src/card/aggregate-status/aggregate-status-card.directive.js
```diff
@@ -18,7 +18,8 @@
  *   </ul>
  * </ul>
  * @param {boolean=} show-top-border Show/hide the top border, true shows top border, false (default) hides top border
- * @param {boolean=} alt-layout Display the aggregate status card in a 'alternate tall' layout.  false (default) displays normal layout, true displays tall layout
+ * @param {string=} layout 'tall' equals the depreciated 'alt-layout' param, 'mini' equals the mini layout.
```
jeff-phillips-18: A list of valid choices for layout would be helpful.

dtaylor113: ok, added list

src/card/aggregate-status/aggregate-status-card.html
```diff
@@ -24,3 +24,30 @@ <h2 class="card-pf-title">
...
+      <span class="card-pf-aggregate-status-count">{{status.count}}</span>
+      {{status.title}}
+    </a>
+    <span ng-if="!status.href">
+      <span class="card-pf-aggregate-status-count">{{status.count}}</span>
+      {{status.title}}
+    </span>
+  </h2>
+  <div class="card-pf-body">
+    <p class="card-pf-aggregate-status-notifications" ng-repeat="notification in status.notifications">
```
jeff-phillips-18: The repeater here is on the "card-pf-aggregate-status-notifications" rather than on "card-pf-aggregate-status-notification" as it is for the other layouts. 

Currently multiple notifications would stack vertically rather than horizontally. Horizontally would make more sense though I'm not sure we should show multiples for the mini layout anyhow. If not, then it should be prevented.

dtaylor113: Confirmed with Liz B. that we are not allowing multiple notifications in the min agg status layout.  Removed the repeater and changed the example java objects.

src/card/aggregate-status/aggregate-status-card.html
```diff
@@ -24,3 +24,30 @@ <h2 class="card-pf-title">
...
+  <h2 class="card-pf-title">
+    <span ng-if="status.iconClass" class="{{status.iconClass}}"></span>
+    <a ng-if="status.href" href="#{{status.href}}">
+      <span class="card-pf-aggregate-status-count">{{status.count}}</span>
+      {{status.title}}
+    </a>
+    <span ng-if="!status.href">
+      <span class="card-pf-aggregate-status-count">{{status.count}}</span>
+      {{status.title}}
+    </span>
```
jeff-phillips-18: Need to document the status parameter and add that when using the 'mini' layout that status is used rather than notifications.

dist/angular-patternfly.js
```diff
@@ -155,12 +160,24 @@ angular.module('patternfly.autofocus', []).directive('pfFocused', ["$timeout", f
...
-     <h4>Aggregate Status Card - Alternate Layout<h4>
-     <div pf-aggregate-status-card status="aggStatus" show-top-border="true" alt-layout="true"></div>
+     <h4>With Top Border<h4>
+     <div pf-aggregate-status-card status="status" show-top-border="true" style="width: 30%"></div>
+     <br/>
+     <h4>No Top Border<h4>
+     <div pf-aggregate-status-card status="status" style="width: 30%"></div>
+     <br/>
+
+     <h4>layout = "mini"<h4>
```
erundle: Any reason why we aren't using columns for this instead of setting widths?

src/card/aggregate-status/aggregate-status-card.html
dtaylor113: Added

dist/angular-patternfly.js
dtaylor113: Using column now

dtaylor113:
All comments addressed.  @erundle @jeff-phillips-18

jeff-phillips-18:
LGTM

dtaylor113:
Fixed example HTML headings

erundle:
LGTM

PR #124, opened 2015-10-12
Adding missing dist files for 2.3.2 fix. Notification service scope

erundle:


erundle:
@jeff-phillips-18 @dtaylor113 these were missing for the 2.3.2 tag

PR #125, opened 2015-10-13
bower main is not pointing to correct css files location

erundle:
This prevents services like wireDep from working properly


erundle:
@jeff-phillips-18 please review/merge

jeff-phillips-18:
LGTM

PR #126, opened 2015-10-13
Moving angular-patternfly to use patternfly 2.4.0

dtaylor113:
Moving angular-patternfly to use pf 2.4.0.
Needed for #123 

@erundle @jeff-phillips-18 please review and merge if ok with this. -thanks!


erundle:
LGTM

PR #131, opened 2015-10-20
Utilization Bar Chart: directive, ng-doc, and unit-tests

dtaylor113:
image

image


dist/angular-patternfly.js
```diff
@@ -1599,6 +1599,159 @@ angular.module('patternfly.charts').directive('pfTrendsChart', function () {
...
+ * <li>.total  - number representing the total amount
+ * </ul>
+ *
+ * @param {object=} chart-title The title displayed on the left-hand side of the chart
+ * @param {object=} layout Various alternative layouts the utilization bar chart may have:<br/>
+ * <ul style='list-style-type: none'>
+ * <li>.type - The type of layout to use.  Valid values are 'regular' (default) displays the standard chart layout,
+ * and 'inline' displays a smaller, inline layout.</li>
+ * <li>.titleLabelWidth - Width of the left-hand title label when using 'inline' layout. Example values are "120px", "20%", "10em", etc..</li>
+ * <li>.usedLabelWidth - Width of the right-hand used label when using 'inline' layout. Example values are "120px", "20%", "10em", etc..</li>
```
waldenraines: Why not just allow the user to set these with css and make the layout parameter a simple string?

dist/angular-patternfly.js
```diff
@@ -1599,6 +1599,159 @@ angular.module('patternfly.charts').directive('pfTrendsChart', function () {
...
+ * </ul>
+ * @param {string=} used-label-format The format of the used label on the right side of the bar chart. Values may be:<br/>
+ * <ul style='list-style-type: none'>
+ * <li>'actual' - (default) displays the standard label of '(n) of (m) (units) Used'.
+ * <li>'percent' - displays a percentage label of '(n)% Used'.</li>
+ * </ul>
+ * @param {object=} units to be displayed on the chart. Examples: "GB", "MHz", "I/Ops", etc...
```
waldenraines: Why is this an object?

dist/angular-patternfly.js
```diff
@@ -1599,6 +1599,159 @@ angular.module('patternfly.charts').directive('pfTrendsChart', function () {
...
+ *   <li>When the used percentage has surpassed the error threshold, the indicator is is red.
+ *   </ul>
+ *
+ * @param {object} chartData the data to be shown in the utilization bar chart<br/>
+ * <ul style='list-style-type: none'>
+ * <li>.used   - number representing the amount used
+ * <li>.total  - number representing the total amount
+ * </ul>
+ *
+ * @param {object=} chart-title The title displayed on the left-hand side of the chart
```
waldenraines: Why is this an object instead of a string?

dist/angular-patternfly.js
dtaylor113: Because that would require users to figure out what css hierarchy to adjust, much simpler to just pass in values and we'll apply them appropriately in the template.

dist/angular-patternfly.js
```diff
@@ -1599,6 +1599,159 @@ angular.module('patternfly.charts').directive('pfTrendsChart', function () {
...
+ * <li>.titleLabelWidth - Width of the left-hand title label when using 'inline' layout. Example values are "120px", "20%", "10em", etc..</li>
+ * <li>.usedLabelWidth - Width of the right-hand used label when using 'inline' layout. Example values are "120px", "20%", "10em", etc..</li>
+ * </ul>
+ * @param {string=} used-label-format The format of the used label on the right side of the bar chart. Values may be:<br/>
+ * <ul style='list-style-type: none'>
+ * <li>'actual' - (default) displays the standard label of '(n) of (m) (units) Used'.
+ * <li>'percent' - displays a percentage label of '(n)% Used'.</li>
+ * </ul>
+ * @param {object=} units to be displayed on the chart. Examples: "GB", "MHz", "I/Ops", etc...
+ * @param {object=} thresholds warning and error percentage thresholds used to determine the Usage Percentage fill color
```
waldenraines: To me, setting `threshold-warning` and `threshold-error` declaratively when initializing the directive is more intuitive than this configuration object.

dist/angular-patternfly.js
dtaylor113: Because an end user could decide to dynamically switch the units displayed.  Perhaps not typical, but could be a usecase.

dist/angular-patternfly.js
dtaylor113: Same reason as above.  User/developer may want to dynamically change the title.  In general, anything that maybe dynamic should be an scope object rather than an attribute string.

dist/angular-patternfly.js
```diff
@@ -1599,6 +1599,159 @@ angular.module('patternfly.charts').directive('pfTrendsChart', function () {
...
+    restrict: 'A',
+    scope: {
+      chartData: '=',
+      chartTitle: '=',
+      units: '=',
+      thresholds: '=?',
+      usedLabelFormat: '@?',
+      layout: '=?'
+    },
+    replace: true,
```
waldenraines: This is [deprecated](https://docs.angularjs.org/api/ng/service/$compile#-replace-deprecated-will-be-removed-in-next-major-release-i-e-v2-0-).

src/charts/utilization-bar/utilization-bar-chart.directive.js
```diff
@@ -0,0 +1,153 @@
...
+    restrict: 'A',
+    scope: {
+      chartData: '=',
+      chartTitle: '=',
+      units: '=',
+      thresholds: '=?',
+      usedLabelFormat: '@?',
+      layout: '=?'
+    },
+    replace: true,
```
waldenraines: This is [deprecated](https://docs.angularjs.org/api/ng/service/$compile#-replace-deprecated-will-be-removed-in-next-major-release-i-e-v2-0-).

src/charts/utilization-bar/utilization-bar-chart.directive.js
```diff
@@ -0,0 +1,153 @@
...
+ * <li>.titleLabelWidth - Width of the left-hand title label when using 'inline' layout. Example values are "120px", "20%", "10em", etc..</li>
+ * <li>.usedLabelWidth - Width of the right-hand used label when using 'inline' layout. Example values are "120px", "20%", "10em", etc..</li>
+ * </ul>
+ * @param {string=} used-label-format The format of the used label on the right side of the bar chart. Values may be:<br/>
+ * <ul style='list-style-type: none'>
+ * <li>'actual' - (default) displays the standard label of '(n) of (m) (units) Used'.
+ * <li>'percent' - displays a percentage label of '(n)% Used'.</li>
+ * </ul>
+ * @param {object=} units to be displayed on the chart. Examples: "GB", "MHz", "I/Ops", etc...
+ * @param {object=} thresholds warning and error percentage thresholds used to determine the Usage Percentage fill color
```
waldenraines: To me, setting threshold-warning and threshold-error declaratively when initializing the directive is more intuitive than this configuration object.

src/charts/utilization-bar/utilization-bar-chart.directive.js
```diff
@@ -0,0 +1,153 @@
...
+ *   <li>When the used percentage has surpassed the error threshold, the indicator is is red.
+ *   </ul>
+ *
+ * @param {object} chartData the data to be shown in the utilization bar chart<br/>
+ * <ul style='list-style-type: none'>
+ * <li>.used   - number representing the amount used
+ * <li>.total  - number representing the total amount
+ * </ul>
+ *
+ * @param {object=} chart-title The title displayed on the left-hand side of the chart
```
waldenraines: Why is this an object instead of a string?

src/charts/utilization-bar/utilization-bar-chart.directive.js
```diff
@@ -0,0 +1,153 @@
...
+ * @param {object=} layout Various alternative layouts the utilization bar chart may have:<br/>
+ * <ul style='list-style-type: none'>
+ * <li>.type - The type of layout to use.  Valid values are 'regular' (default) displays the standard chart layout,
+ * and 'inline' displays a smaller, inline layout.</li>
+ * <li>.titleLabelWidth - Width of the left-hand title label when using 'inline' layout. Example values are "120px", "20%", "10em", etc..</li>
+ * <li>.usedLabelWidth - Width of the right-hand used label when using 'inline' layout. Example values are "120px", "20%", "10em", etc..</li>
```
waldenraines: Why not just allow the user to set these with css and make the layout parameter a simple string?

dist/angular-patternfly.js
waldenraines: I would argue that we should enhance this if it is ever needed as changing the units would involve more than just changing the string.

dist/angular-patternfly.js
waldenraines: This implementation causes me to have to mix concerns between behavior and styling.  Now I have to set display values in my controller that contains my other business logic.  

If we are going to allow setting of width and height via the directive, these values should be set directly when initiating the directive (i.e. `<div pf-utilization-bar-chart title-label-width="200px">`) rather than requiring these values to be in a user's controller (where they do not belong).

src/charts/utilization-bar/utilization-bar-chart.directive.js
```diff
@@ -0,0 +1,153 @@
...
+        if (typeof(newVal) !== 'undefined') {
+          //Calculate the percentage used
+          scope.chartData.percentageUsed = Math.round(100 * (scope.chartData.used / scope.chartData.total));
+
+          scope.isError = (scope.thresholds && scope.chartData.percentageUsed > scope.thresholds.error);
+          scope.isWarn  = (scope.thresholds &&
+                          (scope.chartData.percentageUsed > scope.thresholds.warning &&
+                           scope.chartData.percentageUsed < scope.thresholds.error));
+          scope.isOk    = (scope.thresholds &&
+                          (scope.chartData.percentageUsed < scope.thresholds.warning));
```
waldenraines: What happens here if `scope.thresholds` is falsey?

Should we just check `scope.thresholds` once rather than in each of these?

src/charts/utilization-bar/utilization-bar-chart.directive.js
```diff
@@ -0,0 +1,153 @@
...
+
+          //Animate in the chart load.
+          scope.animate = true;
+          $timeout(function () {
+            scope.animate = false;
+          }, 0);
```
waldenraines: Why is this `$timeout` needed?

src/charts/utilization-bar/utilization-bar-chart.html
```diff
@@ -0,0 +1,34 @@
+<div class="pf-utilization-bar-chart">
+  <span ng-if="!layout || layout.type==='regular'">
```
waldenraines: This would be more readable as `layout.type === 'regular'` (i.e. spaces).

src/charts/utilization-bar/utilization-bar-chart.directive.js
dtaylor113: These variables are used in the template's ng-class call,  if scope.thresholds is falsey than none of these status colors/classes will be used (red, orange, green) and the base class will color the bar, which would be blue.

src/charts/utilization-bar/utilization-bar-chart.directive.js
dtaylor113: You asked this question already, please see answer above.

src/charts/utilization-bar/utilization-bar-chart.html
```diff
@@ -0,0 +1,34 @@
...
+    <div ng-if="chartTitle" class="progress-description">{{chartTitle}}</div>
+    <div class="progress progress-label-top-right">
+      <div class="progress-bar" role="progressbar" ng-class="{'animate': animate,
+           'progress-bar-success': isOk, 'progress-bar-danger': isError, 'progress-bar-warning': isWarn}"
+           style="width:{{chartData.percentageUsed}}%" tooltip="{{chartData.percentageUsed}}% Used">
```
waldenraines: Could you use ui-bootstraps [progressbar](https://angular-ui.github.io/bootstrap/#/progressbar) here?

src/charts/utilization-bar/utilization-bar-chart.html
```diff
@@ -0,0 +1,34 @@
...
+        <span ng-if="!usedLabelFormat || usedLabelFormat==='actual'">{{chartData.used}} of {{chartData.total}} {{units}} Used</span>
+        <span ng-if="usedLabelFormat==='percent'">{{chartData.percentageUsed}}% Used</span>
```
waldenraines: Same comment above about spacing for these.

dist/angular-patternfly.js
dtaylor113: The wrinkle is that these widths only apply when the layout=inline, which is why I added them to the layout object.

src/charts/utilization-bar/utilization-bar-chart.html
```diff
@@ -0,0 +1,34 @@
...
+      </div>
+      <div class="progress-bar progress-bar-remaining" role="progressbar" aria-valuenow="5" aria-valuemin="0" aria-valuemax="100"
+           style="width: {{100-chartData.percentageUsed}}%" tooltip="{{100-chartData.percentageUsed}}% Available">
+      </div>
+    </div>
+  </span>
+  <span ng-if="layout && layout.type==='inline'">
```
waldenraines: Same comment above about spacing.

src/charts/utilization-bar/utilization-bar-chart.html
```diff
@@ -0,0 +1,34 @@
...
+    <div class="progress-container progress-description-left progress-label-right"
+         ng-style="{'padding-left':layout.titleLabelWidth, 'padding-right':layout.usedLabelWidth}">
+      <div ng-if="chartTitle" class="progress-description"  ng-style="{'max-width':layout.titleLabelWidth}">{{chartTitle}}</div>
+      <div class="progress">
+        <div class="progress-bar" role="progressbar" aria-valuenow="25" aria-valuemin="0" aria-valuemax="100"
+             ng-class="{'animate': animate, 'progress-bar-success': isOk, 'progress-bar-danger': isError, 'progress-bar-warning': isWarn}"
+             style="width: {{chartData.percentageUsed}}%" tooltip="{{chartData.percentageUsed}}% Used">
+          <span ng-if="!usedLabelFormat || usedLabelFormat==='actual'" ng-style="{'max-width':layout.usedLabelWidth}">{{chartData.used}} of {{chartData.total}} {{units}} Used</span>
+          <span ng-if="usedLabelFormat==='percent'" ng-style="{'max-width':layout.usedLabelWidth}">{{chartData.percentageUsed}}% Used</span>
```
waldenraines: Same comment above about spacing.

src/charts/utilization-bar/utilization-bar-chart.html
dtaylor113: I received the HTML/CSS from a RH-UXD designer and am required to match angular-patternfly to the base patternfly reflerence implementation HTML/CSS as close  as possible.

src/charts/utilization-bar/utilization-bar-chart.directive.js
waldenraines: That makes sense.  Can we then do the check once, i.e.:

```javascript
if (scope.thresholds) {
  scope.isError = scope.chartData.percentageUsed > scope.thresholds.error;
  scope.isWarn  = (scope.chartData.percentageUsed > scope.thresholds.warning) &&
    (scope.chartData.percentageUsed < scope.thresholds.error);
  scope.isOk = scope.chartData.percentageUsed < scope.thresholds.warning;
}
```

src/charts/utilization-bar/utilization-bar-chart.directive.js
waldenraines: Yeah, sorry, I realized I had commented on the dist version and tried to delete the comments and repost in the source but you were too quick :smile:

src/charts/utilization-bar/utilization-bar-chart.html
waldenraines: Hmm, okay.  I don't have insight into that but to me it seems like as long as it looks the same it should be fine (since these are directives consumed by other projects).  By using ui-bootstrap instead of reinventing it we decrease the amount of code we have to maintain.

src/charts/utilization-bar/utilization-bar-chart.directive.js
dtaylor113: It is a standard we established with the pfDonutChart

src/charts/utilization-bar/utilization-bar-chart.html
dtaylor113: Fixed

src/charts/utilization-bar/utilization-bar-chart.html
dtaylor113: Fixed

src/charts/utilization-bar/utilization-bar-chart.html
dtaylor113: Fixed

src/charts/utilization-bar/utilization-bar-chart.html
dtaylor113: Fixed

src/charts/utilization-bar/utilization-bar-chart.directive.js
dtaylor113: Fixed

src/charts/utilization-bar/utilization-bar-chart.directive.js
dtaylor113: Ok, that worked fine -thanks!

dist/angular-patternfly.js
jeff-phillips-18: I suggested this to Dave because using styling only will not allow for applications to adjust based on locale. While 85px may be fine for english text, the german text may require more and having the ellipses for the title seems silly if the bar is 400px wide (based on the app's layout).

dist/angular-patternfly.js
jeff-phillips-18: The use case would be for applications that change units based on value. That is, showing 1.5 GB instead of 1500 MB. Nothing else would change just the units.

waldenraines: That's true, and it would be up to the consumer to change the data accordingly.  Feel free to disregard the above comment then :+1:

src/charts/utilization-bar/utilization-bar-chart.directive.js
waldenraines: I'm not sure I would call it established, there is an [open discussion](https://www.redhat.com/mailman/private/patternfly-angular/2015-September/msg00001.html) on this topic (that I haven't had a chance to come back around to, unfortunately).

jeff-phillips-18:
LGTM

dist/angular-patternfly.js
erundle: So we shouldn't be messing with the label width. If they don't have room to render the text we either need to look at changing the design of this component or they need to use the top labels.

src/charts/utilization-bar/utilization-bar-chart.directive.js
erundle: This is to allow a CSS3 animation. Basically we need to add a class for a quick second which sets the width to 0 and then remove it. The timeout allows that to occur.

dist/angular-patternfly.js
jeff-phillips-18: I disagree. How can we setup styling that will always fit the text? Why would we disallow wider spacing for the labels?. Using the top labels takes up vertical space and if there is more horizontal space then vertical space available, the application would want to use the horizontal space.

serenamarie125:
I think using the default of 85, but allowing a mechanism for width to be changed by the application is the way to go.

dist/angular-patternfly.js
erundle: Right now we are truncating with an ellipsis which is what the defined behavior is. I mentioned the same concerns you mentioned when the initial spec was being created and this was what the team came up with.

We should not be doing inline CSS. If a product really wants the labels to be larger they should override the css on their own in a CSS file otherwise we are making it too easy for them to create bad designs (huge labels, small bars etc)

jeff-phillips-18: again, I18N comes into play here. there is no set width that would cover multiple languages. Flexibility should be favored as it makes the components usable in more situations. If the application wants bad designs they have many other mechanisms to do so. By making it flexible and usable by more applications, we are at least getting some levels of consistency.

src/charts/utilization-bar/utilization-bar-chart.html
```diff
@@ -0,0 +1,34 @@
...
+           style="width:{{chartData.percentageUsed}}%" tooltip="{{chartData.percentageUsed}}% Used">
+        <span ng-if="!usedLabelFormat || usedLabelFormat === 'actual'">{{chartData.used}} of {{chartData.total}} {{units}} Used</span>
+        <span ng-if="usedLabelFormat === 'percent'">{{chartData.percentageUsed}}% Used</span>
+      </div>
+      <div class="progress-bar progress-bar-remaining" role="progressbar" aria-valuenow="5" aria-valuemin="0" aria-valuemax="100"
+           style="width: {{100-chartData.percentageUsed}}%" tooltip="{{100-chartData.percentageUsed}}% Available">
+      </div>
+    </div>
+  </span>
+  <span ng-if="layout && layout.type === 'inline'">
```
erundle: I don't believe this last example is in the spec for this component?
screen shot 2015-10-21 at 12 34 18 pm

dist/angular-patternfly.js
jeff-phillips-18: In the Patternfly design, there isn't much choice (other than Flexbox) to set some number for a width and use the ellipsis when necessary. We have the choice of being better in Angular Patternfly and we should take advantage of that.

serenamarie125: +1 @jeff-phillips-18  Default should be 85, and the application should be able to modify it appropriately.

erundle: Yes @jeff-phillips-18 I very much understand the i18N concern which is why I raised an issue with the left and right aligned labels to begin with and also commented on the space allowed here: https://github.com/patternfly/patternfly/pull/143

src/charts/utilization-bar/utilization-bar-chart.html
dtaylor113: HI, it is defined here:
https://docs.google.com/document/d/11uZg3vPsw1K87R_QbroG_-ytPXOgWI6Hq849fyuGHns/edit
"Horizontal Configuration"

erundle: I merged in the CSS. Can you please update this PR to look at patternfly master.

dist/angular-patternfly.js
dtaylor113: So is the issue that this helper method makes it too easy for an end user to change widths if they need to?  Please note that the end user can still utilize css changes if they need/want to.

src/charts/utilization-bar/utilization-bar-chart.html
dtaylor113: Done

dist/angular-patternfly.js
dtaylor113: > +1 @jeff-phillips-18 Default should be 85, and the application should be able to modify it appropriately.

Unless I hear other concerns, I'm going to go with Serena's comment and consider this closed.

src/charts/utilization-bar/utilization-bar-chart.directive.js
dtaylor113: I'm concerned that this would bring the number of attributes for this directive up to 8, plus what if we wanted to introduce another threshold ("critical", etc..), then we'd need additional attributes.

waldenraines: How many of those attributes would be used in a standard case?

Another, cleaner way of grouping these directives instead of having a single directive with all of the attributes on it would be to have a "parent" directive with child directives for particular types of charts.

dtaylor113: Ok, as thresholds aren't likely to change, I don't think they should be in the controller.  I'll implement the attributes you suggest.

dtaylor113: Fixed

dtaylor113:
Hi, I believe I have addressed all issues. @waldenraines if happy, please merge

dtaylor113:
@waldenraines?

PR #134, opened 2015-10-23
Card: Adjusted styles to match pf ref. impl.

dtaylor113:
Only html/css changes to the header and footer areas to properly display the href and filter dropdown(s)

image


jeff-phillips-18:
LGTM

dtaylor113:
@erundle please review and merge if ok with changes -thanks!

src/card/basic/card-filter.html
```diff
@@ -2,7 +2,7 @@
   {{currentFilter.label}}
   <span class="caret"></span>
 </button>
-<ul class="dropdown-menu">
+<ul class="dropdown-menu dropdown-menu-right" role="menu">
```
erundle: We should probably be using the angular-ui directive here.

src/card/basic/card.html
```diff
@@ -1,42 +1,33 @@
...
-    <div class="card-pf-table">
-      <div class="card-pf-cell">
-        <h2 class="card-pf-title" ng-class="{'card-pf-filter-header': showFilterInHeader()}">{{headTitle}}</h2>
-      </div>
-      <div class="card-pf-cell text-right" ng-if="showFilterInHeader()">
-        <div class="dropdown btn-group">
-          <div ng-include="'card/basic/card-filter.html'"></div>
-        </div>
-      </div>
+  <div ng-if="showHeader()" ng-class="!showTitlesSeparator || showTitlesSeparator === 'true' ? 'card-pf-heading' : 'card-pf-heading-no-bottom'">
```
erundle: Can this just be replaced with a function call?

src/card/basic/card-filter.html
dtaylor113: I'm trying to use the exact html/css used in the base pf ref. impl.

src/card/basic/card.html
dtaylor113: Fixed

src/card/basic/card.directive.js
```diff
@@ -180,9 +180,17 @@ angular.module('patternfly.card').directive('pfCard', function () {
         return ($scope.headTitle || $scope.showFilterInHeader());
       };
 
+      $scope.getHeaderClass = function () {
+        return (!$scope.showTitlesSeparator || $scope.showTitlesSeparator === 'true' ? 'card-pf-heading' : 'card-pf-heading-no-bottom');
+      };
```
waldenraines: Thoughts on pulling this into the html and using `ng-class`?

dtaylor113: That logic was in the html template, I moved into to the controller function per @erundle comment above :smile:  HTML is using ng-class

erundle: So I think Dave made this change in response to my code comment. What I was looking for was a scope variable that was just to to true or false so we weren't checking $scope.showTitlesSeparator === 'true' in the HTML. I am fine with the ng-class being in the HTML

erundle: See https://github.com/patternfly/angular-patternfly/blob/master/src/card/aggregate-status/aggregate-status-card.directive.js#L142

dtaylor113: Ok, changed to match that coding pattern

src/card/basic/card.html
```diff
@@ -1,42 +1,33 @@
...
-      <div class="card-pf-cell">
-        <h2 class="card-pf-title" ng-class="{'card-pf-filter-header': showFilterInHeader()}">{{headTitle}}</h2>
-      </div>
-      <div class="card-pf-cell text-right" ng-if="showFilterInHeader()">
-        <div class="dropdown btn-group">
-          <div ng-include="'card/basic/card-filter.html'"></div>
-        </div>
-      </div>
+  <div ng-if="showHeader()" ng-class="{'card-pf-heading': shouldShowTitlesSeparator, 'card-pf-heading-no-bottom': !shouldShowTitlesSeparator}">
+    <div ng-if="showFilterInHeader()" class="dropdown card-pf-time-frame-filter">
```
erundle: This might be easier to read
ng-class="shouldShowTilesSeparator ? 'card-pf-heading' | 'card-pf-heading-no-bottom'"

dtaylor113: Fixed

erundle: Thanks. @waldenraines any other issues or can I merge?

waldenraines:
LGTM

PR #143, opened 2015-10-28
Update utilization chart available value when used or total changes

spadgett:
Update the text in the upper left corner of the utilization chart showing the current value when the data changes.

/cc @jwforres @erundle

PR #144, opened 2015-10-28
Update utilization chart available value when used or total changes

erundle:
This has already been merged into master. Adding to 2.3.x so that we can create a patch release

@spadgett


spadgett:
Hi, @erundle. It looks like your commit is different from #143. The change needs to be in utilization-chart-directive.js.

erundle:
@spadgett what is different about it?

spadgett:
The change is not in the correct file. It should be in utilization-chart-directive.js, not donut-pct-chart-directive.js.

erundle:
@spadgett nevermind it looks like we have the bug in two places and show be applied to both files

PR #146, opened 2015-10-29
Fix for PTNFLY-539 "AngJS Simple Filter - Attr. Selector should be…

dtaylor113:
…single select"

https://patternfly.atlassian.net/browse/PTNFLY-539


src/filters/simple-filter-directive.js
```diff
@@ -183,13 +183,23 @@ angular.module('patternfly.filters').directive('pfSimpleFilter', function () {
...
           id: field.id,
           title: field.title,
+          type: field.filterType,
           value: value
         };
         if (!$scope.filterExists(newFilter)) {
+
+          $scope.enforceSingleSelect(newFilter);
+
           $scope.config.appliedFilters.push(newFilter);
```
jeff-phillips-18: I wonder if we should modify the existing filter to use the new selection rather than removing the old and adding the new. This would keep the filter in the same position in the list of filters. This is really a modification of the current filter.

Thoughts?

src/filters/simple-filter-directive.js
```diff
@@ -183,13 +183,23 @@ angular.module('patternfly.filters').directive('pfSimpleFilter', function () {
         return foundFilter !== undefined;
       };
 
+      $scope.enforceSingleSelect = function (filter) {
+        if (filter.type === 'select') {
+          _.remove($scope.config.appliedFilters, {title: filter.title, type: filter.type});
+        }
```
jeff-phillips-18: I don't think we need type here. Any filters with that title should have the same type, otherwise the same title (field) would show up in the list of filters more than once.

src/filters/simple-filter-directive.js
dtaylor113: I was wondering if we should be consistent and always show the order that the filters were added.  Removing old, and adding new, makes the filter appear last in the active filters list, so user can see what they added/changed last.
But, it might be better, as Jeff suggests, to just change the existing filter and it keeps it's position in the active filters list.

cshinn: I think Dave's idea of doing a remove and add rather than a replace makes sense. The "jolt" in the filter list should make it more obvious that something has changed and that the old filter is no longer active.

serenamarie125: +1 to removing old and adding new.  It allows the user to see the most recent change as well

src/filters/simple-filter-directive.js
dtaylor113: Ok, removed title

dtaylor113:
@erundle please review and merge if ok with fix -thanks

src/filters/simple-filter-directive.js
```diff
@@ -183,13 +183,23 @@ angular.module('patternfly.filters').directive('pfSimpleFilter', function () {
...
       $scope.addFilter = function (field, value) {
         var newFilter = {
           id: field.id,
           title: field.title,
+          type: field.filterType,
           value: value
         };
         if (!$scope.filterExists(newFilter)) {
+
+          $scope.enforceSingleSelect(newFilter);
```
erundle: So I would change the order of this code. You want to make the check for if (filter.type === 'select')  here and then if true call enforceSingleSelect. Also should this be an option or do we want this as the default.

src/filters/simple-filter-directive.js
jeff-phillips-18: OK, by title you meant type  ;)

jeff-phillips-18:
LGTM

src/filters/simple-filter-directive.js
dtaylor113: I believe we want to do this as default

dtaylor113: changed order of code

src/filters/simple-filter-directive.js
```diff
@@ -183,13 +183,23 @@ angular.module('patternfly.filters').directive('pfSimpleFilter', function () {
...
         var newFilter = {
           id: field.id,
           title: field.title,
+          type: field.filterType,
           value: value
         };
         if (!$scope.filterExists(newFilter)) {
+
+          if (newFilter.type === 'select') {
+            $scope.enforceSingleSelect(newFilter);
```
erundle: Thank you. This is much more readable.

dtaylor113: no problem, merge please?

PR #148, opened 2015-11-02
Show Data Tile checkboxes only on hover or if selected

jeff-phillips-18:
According to patternfly specs (draft) checkboxes should only be shown
in data tiles when hovered over the tile or if selected.


dtaylor113:
LGTM

PR #149, opened 2015-11-03
Heatmap

erundle:
Initial version of heatmap component


src/charts/heatmap/heatmap-legend.directive.js
```diff
@@ -0,0 +1,25 @@
+angular.module('patternfly.charts').directive('pfHeatmapLegend',
+  function () {
+    'use strict';
+    return {
+      restrict: 'A',
+      scope: {
+        legend: '=',
+        legendColors: '='
+      },
+      replace: true,
```
waldenraines: This is deprecated.

src/charts/heatmap/heatmap.directive.js
```diff
@@ -0,0 +1,223 @@
...
+    scope: {
+      data: '=',
+      height: '=',
+      chartTitle: '=?',
+      legendLabels: '=?',
+      thresholds: '=?',
+      heatmapColorPattern: '=?',
+      clickAction: '=?'
+    },
+    replace: true,
```
waldenraines: This is deprecated.

dtaylor113:
Hi,
Heatmap Legend doesn't display properly in IE 9:

headmapinie

I only see this warning in the Console:
**tooltip-html-unsafe is now deprecated. Use tooltip-html or tooltip-template instead.** 

Then when I move the mouse over heatmap boxes I see a bunch of:
**bindHtmlUnsafe is now deprecated. Use ngBindHtml instead 
bindHtmlUnsafe is now deprecated. Use ngBindHtml instead 
bindHtmlUnsafe is now deprecated. Use ngBindHtml instead**

src/charts/heatmap/heatmap.directive.js
```diff
@@ -0,0 +1,223 @@
...
+        } else {
+          sx = x / px;
+        }
+
+        if (Math.floor(py * x / y) * py < n) {
+          sy = x / Math.ceil(x * py / y);
+        } else {
+          sy = y / py;
+        }
+        return Math.max(sx, sy);
```
jeff-phillips-18: I think we should have a maximum size. If there are only 4 blocks, it's not great to draw each of them really huge. This might be an enhancement request.

erundle: Yes minimum and max block size should both be in the next version

erundle:
@dtaylor113 I am trying to track down the HTML issue. tooltip-html doesn't work for me even with escaped HTML content and I don't believe tooltip-template isn't really an option for this use case. I don't know if it is an issue with using angular 1.4.x with UI bootstrap 0.13.x

erundle:
The IE issue with the legend was addressed. This has been rebased and is green.

dtaylor113:
Team decided to allow this to be merged with the tooltip console warnings.  When/if it becomes an issue we will re-address at that time.

PR #150, opened 2015-11-03
Adding tooltips for sort and filter fields

erundle:


dist/angular-patternfly.js
```diff
@@ -3205,6 +3205,26 @@ angular.module('patternfly.sort').directive('pfSimpleSort', function () {
...
+          }
+        } else {
+          if ($scope.config.isAscending) {
+            tooltip = 'Sort alphabetically, from A to Z';
+          } else {
+            tooltip = 'Sort alphabetically, from Z to A';
+          }
+        }
+
+        $scope.sortTooltip = tooltip;
```
waldenraines: What about putting these in the HTML?  Seems odd to have this sort of stuff in a controller.

src/sort/simple-sort-directive.js
```diff
@@ -154,6 +154,26 @@ angular.module('patternfly.sort').directive('pfSimpleSort', function () {
...
+          }
+        } else {
+          if ($scope.config.isAscending) {
+            tooltip = 'Sort alphabetically, from A to Z';
+          } else {
+            tooltip = 'Sort alphabetically, from Z to A';
+          }
+        }
+
+        $scope.sortTooltip = tooltip;
```
waldenraines: What about putting these in the HTML? Seems odd to have this sort of stuff in a controller.

erundle: I don't think we really want that many ng-ifs sprinkled in the HTML unless you have another idea. We have the icon logic in the controller so for consistency reasons I would suggest they stay together.

dtaylor113:
LGTM

src/sort/simple-sort-directive.js
```diff
@@ -154,6 +154,26 @@ angular.module('patternfly.sort').directive('pfSimpleSort', function () {
     },
     templateUrl: 'sort/simple-sort.html',
     controller: function ($scope) {
+      var setSortTooltip = function () {
```
jeff-phillips-18: This name is a bit confusing, more specifically this is the sort direction tooltip.

erundle: Changed to setSortDirectionTooltip and squashed

jeff-phillips-18:
LGTM

erundle:
@waldenraines can this be merged?

waldenraines:
@erundle I personally would like to see the tooltips in the HTML itself rather than the controller, what is the opposition to ng-ifs in the HTML?  I won't hold this PR up on this point but to me it's cleaner.

erundle:
I would prefer not to repeat this markup 4 times:

<button class="btn btn-link" type="button"  ng-click="changeDirection()" tooltip="{{ sortDirectionTooltip }}" tooltip-placement="bottom"><span class="sort-direction" ng-class="getSortIconClass()"></span></button>

Granted with having the ng-ifs in the HTML we can also get rid of the getSortIconClass method I just find this harder to read and more error prone but that is just me.

src/sort/simple-sort-directive.js
```diff
@@ -154,6 +154,26 @@ angular.module('patternfly.sort').directive('pfSimpleSort', function () {
     },
     templateUrl: 'sort/simple-sort.html',
     controller: function ($scope) {
+      var setSortDirectionTooltip = function () {
+        var tooltip = '';
+
+        if ($scope.config.currentField.sortType === 'numeric') {
+          if ($scope.config.isAscending) {
+            tooltip = 'Sort numerically, from least to greatest';
```
waldenraines: When I hover over a button, I expect the tooltip to tell me what pressing the button would do, not what the current state of the button is.  Here we are doing the latter and I just want to make sure that is intentional.

jeff-phillips-18: I have to agree with @waldenraines here. I'm not really sure why we are setting the tooltip to explain what the current state of the button means. Shouldn't the tooltip just be "Change the current sort direction" always?

waldenraines:
> I would prefer not to repeat this markup 4 times:

Me too.  If I were to design this I may have done so by wrapping the tooltip directive in a "directional tooltip" directive that handles (and hides away) this behavior.  I suppose it just seems odd to me to have logic for a tooltip inside of the sort controller but, like I said, I won't hold this PR up because of that.

Do you think we'll have to copy this logic elsewhere in the future?  If so then this approach would make sense.

src/sort/simple-sort-directive.js
erundle: I asked @serenamarie125 to comment since this text came from her

erundle:
I don't know if we will have more of this in the future but creating a directive for this seems like overkill for a handful of if blocks. If we were to take that approach we would want to do that same thing for the sort icon itself which means this simple-sort-directive.js is pretty much paired down to nothing. I could see taking that approach if the file size was large and these components were complex but right now that isn't the case. IMO having everything in one file makes it easier for the developer to understand what is going on with a code base of this size.

src/sort/simple-sort-directive.js
serenamarie125: Please pull the tooltips from the sort icons themselves, and only have tips on Sort by and Filter by.

erundle:
I talked with @serenamarie125 and we are going to pull the tooltip from the sort icon all together. I made the change and squashed the commit so this should be G2G.

PR #152, opened 2015-11-04
Fix for PTNFLY-616: AngJS Data List: In IE & FF Actions Menu doesn't display

dtaylor113:
- 'event.toElement' doesn't work in IE or FF, switched to 'event.target'
- '\n' doesn't cause a newline in textarea in IE or FF, switched to '\r\n'
- All fixes work in Chrome, IE, and FF


jeff-phillips-18:
LGTM

waldenraines:
LGTM

dtaylor113:
@erundle please merge if ok with this, thanks!

PR #154, opened 2015-11-06
pfSelect - fixed blank choice in select

thomasmckay:
If the ng-model does not contain a recognized value a blank entry will be displayed in the choices. I set the value to "orange" to be different than the hard-coded value of "apple".


jeff-phillips-18:
LGTM

waldenraines:
LGTM

PR #155, opened 2015-11-06
added optional port override

thomasmckay:
If port 8000 is already in use, this will allow an alternative to be specified.


jeff-phillips-18:
LGTM, simple change, merging

PR #157, opened 2015-11-09
Fix to allow setting of sparkline chart height in config.

jeff-phillips-18:
The config value was getting over-written by ‘undefined’ when the
chart-height value is not set.


dtaylor113:
LGTM

PR #159, opened 2015-11-10
Fix setting colors and legend labels. Allow not showing legend.

jeff-phillips-18:
Changing the default color pattern and labels for a heat map did not work correctly.
The heat map legend would not allow default color patterns or labels.
The heat map did not allow hiding the legend.


dtaylor113:
LGTM

PR #163, opened 2015-12-08
Distributing ngdocs for use in pf.org

dtaylor113:
Checking in the following additional 'dist/doc' directories:

dist/docs/css
dist/docs/js
dist/docs/grunt-scripts
dist/docs/partials

These artifacts will be utilized by pf.org v2 to show angular-patternfly ngdoc partials within the new pf.org site.


jeff-phillips-18:
LGTM

PR #164, opened 2015-12-18
Update to use the c3ChartsDefaults from Patternfly

jeff-phillips-18:
Updated bower.son to pull from #master, this needs to be updated on
release (as usual but just making in clear).


dtaylor113:
LGTM

PR #165, opened 2015-12-21
Add pfEmptyChart directive, use it in Heatmap, Trends, and Utilizatio…

jeff-phillips-18:
…n charts


dtaylor113:
Downloaded this branch, viewed ng-docs, reviewed code.  LGTM, although the "no data" icon and message doesn't look centered in the HeatMaps, but that could just be because there is no border to actually see the width of the "no data" heatmap.  Just looks odd to me.

jeff-phillips-18:
Yeah, it is centered. If you size the browser down it looks better. Not sure of a good way to fix this. Maybe it should just have a left padding rather than be centered. I'll try to get some opinions.

jeff-phillips-18:
Since heat maps (and most charts) will usually be in a card, I think it's best to leave the icon and text centered. I've updated the example to add a border around the heat maps to make it look better in this context.

dtaylor113:
Ok, sounds good.

dtaylor113:
@rhamilto are you ok with this PR?  Ready to merge?  -thanks!

rhamilto:
@dtaylor113 given my lack of Angular knowledge, I defer to you guys.

PR #170, opened 2016-01-04
Fixes to handle no chart data for data unavailable charts

jeff-phillips-18:
Also straighten up the inclusion of patternfly.js in ngdocs app.


dtaylor113:
LGTM

PR #176, opened 2016-01-20
Fixes 151 - Set bower dependencies

jeff-phillips-18:
Move dependencies that are required by consumers from
devDependencies to dependencies in bower.json.


dtaylor113:
LGTM

PR #184, opened 2016-02-01
bump to v2.10.0

dtaylor113:
**Release Notes**
Updated devDependency for patternfly to 2.10.0

**Enhancements**

DataList - Updated to use latest Patternfly styling and added support for Action Buttons
HeatMaps - Made HeaMap Height parameter optional and updated ngdoc
Bower - Moved dependencies that are required by consumers from devDependencies to dependencies in bower.json

**Bug Fixes**
TrendChart - Fixed showing trend charts details for inline layout

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "angular-patternfly",
-  "version": "2.9.0",
+  "version": "2.10.0",
   "authors": [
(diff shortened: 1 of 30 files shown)


jeff-phillips-18:
LGTM

PR #195, opened 2016-02-25
Fix for list view to show actions when any actions are supplied

jeff-phillips-18:
Previously the actions only showed if both menu and button actions were
supplied.


dtaylor113:
LGTM

Issue #193, opened 2016-02-24
Action button in pfToolbar has broken style


romanblanco:
It looks that there is wrong padding for button, In case I have just one Action button used in [pfToolbar](http://angular-patternfly.rhcloud.com/#/api/patternfly.toolbars.directive:pfToolbar):

code example:

```javascript
    vm.toolbarConfig = {
      actionsConfig: {
        primaryActions: [
          {
            name: 'Edit',
            title: 'Edit service dialog',
            actionFn: performAction
          }
        ]
      }
    };
```

**screenshots:**

_single action button:_
![screencapture-localhost-3001-self_service-dialogs-1456343129813](https://cloud.githubusercontent.com/assets/1187051/13298858/cd4ef4a8-db38-11e5-97d2-fc66ffddde1d.png)

_action button together with filters:_
![screencapture-localhost-3001-self_service-dialogs-1456343561548](https://cloud.githubusercontent.com/assets/1187051/13298860/ce78d1aa-db38-11e5-98ac-d167bd4d0793.png)

serenamarie125:
@romanblanco I'm assuming that you are talking about the fact that there is no bottom padding on the Edit button in your first screenshot.  Can you please confirm?

romanblanco:
@serenamarie125 yes, that is what I'm talking about :blush:

PR #197, opened 2016-02-26
Fix for #193, add a bottom margin to the toolbar when no filter results

jeff-phillips-18:


dtaylor113:
LGTM, and not just because it's Friday :smile:

dlabrecq:
Looks good

PR #199, opened 2016-03-01
Add Cards View example to the toolbar directive docs.

jeff-phillips-18:


dtaylor113:
Downloaded repo, ran pfToolbar example in Card View.  LGTM

PR #200, opened 2016-03-02
PTNFLY-1168 Code for controls should not be in angular ngdoc snippets

dlabrecq:
Control code replaced with straight HTML showing each chart variation


jeff-phillips-18:
The trends charts inline layout has x & y axis, it should not. I assume this is done to show one of the charts with them but without labels it's not something we should promote.

jeff-phillips-18:
The sparkline charts for Percentage Tooltip shows x & y axis and should not for the same reasons.

jeff-phillips-18:
Ditto for Utilization Bar chart for No Donut Center Label

jeff-phillips-18:
Shouldn't we also remove controls from the pfC3Chart and pfHeatMap pages?

jeff-phillips-18:
Is this limited to chart or should we also include:

pfInlineNotification
pfNotificationList
pfCardView
pfListView

dtaylor113:
Hi, sorry I didn't do a great job of identifying all the angular-patternfly ngdocs which need to be updated, tried to do it quick during the grooming meeting.

Please update the JIRA story below with all the directives we decide to update:
https://patternfly.atlassian.net/browse/PTNFLY-1168

Thanks,
- Dave

dtaylor113:
Downloaded repo, viewed the changed ngdocs so far.
Actually like the new layout for pfSparklineChart, nice to clearly see the different tooltip variations.

dtaylor113:
For pfTrendsChart, the Inline example has the x and y axis turned on.  If this is intentional, then perhaps update the title to something like "Inline Layout with X & Y axis shown"

dtaylor113:
Hi, how do we feel about adding HRs to some of the ngdocs:

**pfTrendsChart**

image

dlabrecq:
I like that better considering there are so many examples. I thought 
about doing that, but didn't because pfUtilizationBarChart didn't have it.

dlabrecq:
I thought we wanted an example with the x and y axis because the 
previous controls had a way to toggle this on/off. I'm planning to 
remove it per Jeff's earlier recommendation.

jeff-phillips-18:
I would remove the axis unless we want to add labels as well but then I would add an additional large layout chart with the axes since only that layout really supports them well.

dlabrecq:
Question about pfCardView. This page doesn't appear to have controls, like the others. It has areas for footer and filters. Each config has its own callback, which shows an alert popup. Should I remove the callbacks? As a developer, it seems like something I'd be interested in seeing here.

dlabrecq:
Question about pfNotification. Not sure what to do here as the notifications eventually disappear? It's possible that the notifications won't be visible by the time a user scrolls to the bottom of the page. Kind of need the control to generate more.

dlabrecq:
Question about pfList. Do we want to keep the events?

jeff-phillips-18:
I would leave the events and such as they show good examples. 

For the notifications, IMO pfInlineNotification should show all the variations with no controls, leave pfNotificationList and Notification as is.

PR #205, opened 2016-03-15
Add callback param for accessing C3Chart object

### LINKED ISSUE #204: Provide access to the chart object for C3 charts
When using C3 charts is often necessary to access the API for performing additional operations. For that, it should be possible to get access to the chart object that is generated by C3.

ammendonca:
Fixes #204


src/charts/c3/c3-chart.directive.js
```diff
@@ -69,17 +78,23 @@
     return {
       restrict: 'A',
       scope: {
-        config: '='
+        config: '=',
+        callback: '='
```
jeff-phillips-18: Can we name this as getChartCallback or generateChartCallback? In the off case where we need to add another callback.

ammendonca: Agree. Will use getChartCallback.

jeff-phillips-18:
Please add  getChartCallback to the ng-docs parameters list.

jeff-phillips-18:
LGTM

dtaylor113:
LGTM

PR #210, opened 2016-03-23
Angular Bootstrap Switch added as a dependency

romanblanco:
@jeff-phillips-18

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -38,6 +38,7 @@
     "angular-animate": "1.3.0 - 1.4.*",
     "angular-sanitize": "1.3.0 - 1.4.*",
     "angular-bootstrap": "0.13.x",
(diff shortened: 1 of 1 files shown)


dlabrecq:
LGTM

jeff-phillips-18:
Why do we need to add this dependency?

romanblanco:
[Bootstrap switch](http://www.bootstrap-switch.org/) itself seems to not working with AngularJS, at least to me.

jeff-phillips-18:
But unless angular-patternfly depends on it then we shouldn't add the dependency. If we had a component built on top of it then it would make sense. Projects using angular and the bootstrap switch can add the dependency.

romanblanco:
@jeff-phillips-18 thanks, closing this issue

PR #211, opened 2016-03-23
add line-chart wrapper

yaacov:
This pull request adds pfLineChart directive to angular-patternfly.
It adds patternfly's line-chart, https://www.patternfly.org/patterns/line-chart/

line-chart is useful to show trends of two or more values over the same time window, e.g. creation and deletion of container pods over the same time window. Unlike trend-chart and sparkline, line-chart can show multiple data lines using multiple colors.

screenshot from 2016-03-24 14-49-43


yaacov:
cc @alongoldboim @zeari @simon3z

src/charts/line/line-chart.directive.js
```diff
@@ -0,0 +1,287 @@
...
+            }
+          }
+        };
+
+        // Setup the default configuration
+        $scope.defaultConfig.point.r = 1;
+        $scope.defaultConfig.grid.y.show = false;
+        $scope.defaultConfig.color.pattern = ["#92D400", "#CC0000", "#0088CE", "#9F7FC3"];
+        $scope.defaultConfig.tooltip = $scope.lineTooltip();
+        $scope.defaultConfig.units = '';
```
jeff-phillips-18: This should be using the default line config provided by Patternfly.
We need to first update c3-charts-defaults to include the new patternfly default chart configs for line chart such as:

      getDefaultLineAxis: patternflyDefaults.getDefaultLineAxis,
      getDefaultLineGrid: patternflyDefaults.getDefaultLineGrid,
      getDefaultLineLegend: patternflyDefaults.getDefaultLineLegend,
      getDefaultLinePoint: patternflyDefaults.getDefaultLinePoint,
      getDefaultLineConfig: patternflyDefaults.getDefaultLineConfig,

The use the new c3-charts-defaults.getDefaultLineConfig here rather than creating the default config.

Note that this config is for multiple lines only and for a single line chart it should use the getDefaultSingleLineConfig Patternfly function (basically just the tooltip changes).

src/charts/line/line-chart.directive.js
```diff
@@ -0,0 +1,287 @@
...
+ *
+ * @param {object} config configuration settings for the line chart:<br/>
+ * <ul style='list-style-type: none'>
+ * <li>.chartId            - the ID of the container that the chart should bind to
+ * <li>.units              - unit label for values, ex: 'MHz','GB', etc..
+ * <li>.tooltipType        - (optional) set the type of tooltip, valid values:
+ * <ul style='list-style-type: none'>
+ * <li>'default'           - show the data point value and the data point name.
+ * <li>'valuePerDay'       - show the date and value for the data point.
+ * </ul>
```
jeff-phillips-18: Does this tooltip type make sense for a line chart wrapper? I believe these come from the sparkline chart wrapper which has a more specific purpose.

src/charts/line/line-chart.directive.js
```diff
@@ -0,0 +1,287 @@
...
+                return {
+                  top: top - height,
+                  left: Math.min(x, chartBox.width - width)
+                };
+              } catch (e) {
+              }
+            }
+          };
+        };
+
```
jeff-phillips-18: I believe we can drop these tooltip functions and just rely on the Patternfly default tooltip functionality.

yaacov:
Changes done:
1. Using the default line config provided by Patternfly.
2. Just rely on the Patternfly default tooltip functionality.
cc @jeff-phillips-18

dist/angular-patternfly.js
```diff
@@ -1465,6 +1466,210 @@ angular.module('patternfly.charts').directive('pfHeatmap', ["$compile", function
...
+            tick: {
+              format: function () {
+                return '';
+              }
+            }
+          }
+        };
+
+        // Setup the default configuration
+        $scope.defaultConfig.point.r = 1;
```
jeff-phillips-18: This is set by the default patternfly config to be 3 and shouldn't be overridden here.

src/charts/line/line-chart.directive.js
```diff
@@ -0,0 +1,204 @@
...
+            tick: {
+              format: function () {
+                return '';
+              }
+            }
+          }
+        };
+
+        // Setup the default configuration
+        $scope.defaultConfig.point.r = 1;
```
jeff-phillips-18: This is set by the default patternfly config to be 3 and shouldn't be overridden here.

yaacov:
> This is set by the default patternfly config to be 3 and shouldn't be overridden here.
@jeff-phillips-18 sorry :-) line removed.

simon3z:
The two shades of blue used in the screenshot are hard to distinguish :smile: ...anyway I gather that you can pick different colors for the two lines, which is a :+1: in terms of usability for me. (PS. blue and green would have been nicer as example).

jeff-phillips-18:
LGTM

yaacov:
> @simon3z: PS. blue and green would have been nicer as example

Changed tha demo to be blue and green :-) (defaults stay the same)

simon3z:
Thanks @yaacov as far as I am concerned LGTM (and @jeff-phillips-18 ack'ed this already). So I think it's ready to be merged. (cc @dtaylor113 )

src/charts/line/line-chart.directive.js
```diff
@@ -0,0 +1,203 @@
...
+        if ($scope.config.chartId) {
+          $scope.lineChartId = $scope.config.chartId + $scope.lineChartId;
+        }
+
+        /*
+         * Convert the config data to C3 Data
+         */
+        $scope.getLineData = function (chartData) {
+          var lineData  = {
+            type: 'area'
```
jeff-phillips-18: Sorry I didn't catch this before. The directive is line chart but this is actually creating an area chart. It should be renamed if the type is area or change type here to 'line'.

dist/angular-patternfly.js
```diff
@@ -1465,6 +1466,209 @@ angular.module('patternfly.charts').directive('pfHeatmap', ["$compile", function
...
+   </file>
+
+   <file name="script.js">
+     angular.module( 'patternfly.charts' ).controller( 'ChartCtrl', function( $scope ) {
+
+       $scope.config = {
+         chartId: 'exampleLine',
+         grid: {y: {show: false}},
+         point: {r: 1},
+         color: {pattern: ["#0033CF", "#00CF33"]}
```
jeff-phillips-18: Please choose colors from the patternfly color palette found at https://www.patternfly.org/styles/color-palette/

yaacov:
Changes done:
1. Default to line type. Optional Area type.
2. Demo use colors from the patternfly color palette, it does look better :-)
cc @jeff-phillips-18

jeff-phillips-18:
LGTM (again) :+1:

dtaylor113:
LGTM

PR #217, opened 2016-03-24
Add optional tooltip to line chart

yaacov:
The angular-patternfly line chart use the defualt patternfly tool-tip, this is useful in most cases.
It is also useful to override the default tool-tip in some cases.
This pull request add the option to override the default tool-tip.

screenshot from 2016-03-24 17-18-19_01


yaacov:
@simon3z @jeff-phillips-18 please review

jeff-phillips-18:
Is this needed? The caller can set config.tooltip how they like and it will get used.

yaacov:
@jeff-phillips-18 
> Is this needed? The caller can set config.tooltip how they like and it will get used.

The tooltip function position element is not nice code to leave for the user.
For example, the line:
`chartBox = document.querySelector('#' + $scope.lineChartId).getBoundingClientRect();`
gets very ugly in my config object :-(

jeff-phillips-18:
OK, so this really just takes a custom tooltip and positions it at the top of the chart. Would it be better to have a config option that tells us to position the tooltip at the top (default or otherwise)? And if so, just set the config.tooltip.position to this function which does so.

yaacov:
@jeff-phillips-18 
> Would it be better to have a config option that tells us to position the tooltip at the top (default or otherwise)?

Re-position of the tool-tip is a side-effect, I just want to set the string in the tool-tip.
Is there a way to set the string without changing the position ?

yaacov:
@jeff-phillips-18 after rebase i get `Error: pfUtils is not defined` on the demo `pfUtils.colorPalette.blue` line :-(
what am i doing wrong ?

jeff-phillips-18:
Sorry, this is an issue from my last PR to use the Patternfly Color Palette constants. Fix will be in shortly.

jeff-phillips-18:
I believe you really only want to set the tooltip function like I was suggesting before. No need to change the directive, just set:
config.tooltip: {
 contents : myTooltipFunction
}

yaacov:
@jeff-phillips-18 Thanks :+1: I did not know that :-)
closing this PR

PR #220, opened 2016-04-01
Split pfCard examples to be included separately in patternfly-toolkit.

dlabrecq:
@dtaylor113 and @jeff-phillips-18 please review.

### DIFF
--- a/dist/angular-patternfly.js
+++ b/dist/angular-patternfly.js
@@ -295,7 +295,7 @@ angular.module( 'patternfly.card' ).directive('pfAggregateStatusCard', function 
 });
 ;/**
  * @ngdoc directive
- * @name patternfly.card.directive:pfCard
+ * @name patternfly.card.directive:pfCard - Utilization
  * @restrict A
  * @element ANY
  * @param {string} headTitle Title for the card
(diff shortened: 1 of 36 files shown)


dtaylor113:
LGTM:  I would just suggest making the titles be:

pfCard: Timeframe Filters
pfCard: Trend Charts
pfCard: Multiple Utilization Bars

Either a colon ":" or perhaps a dahs "-".  -thanks

dlabrecq:
I can't do the colon ":", but dash "-" works fine. When colon is added, the "pfCard:" prefix isn't shown.

dtaylor113:
ok, sounds good

jeff-phillips-18:
Did you think about putting the examples in a different location from the actual directives? This would have less clutter in the actual code directories and make the examples easily distinguished from additional directives.

dlabrecq:
I actually tried placing examples under the doc name "patternfly.examples". In this scenario, the directive only contained one example (i.e., utilization), while "pfCard Trend" and "pfCard Timeframe" were split out. 

The new examples appeared as a block, somewhere in the middle of everything else. My thinking was it may be best to keep all the pfCard examples together? It could be confusing jumping around in order to find all examples? 

I can throw an example together if you would like to review that, first?

dlabrecq:
I created an alternate example #221 with pfCard examples under the "patternfly.examples" doc name. Please take a look and let me know what you think?

dtaylor113:
I think Jeff was just talking about moving the actual files to an 'examples' directory -leaving the ngdoc @names the same.   Just concerned about cluttering up the number of files in the actual directive directories

dlabrecq:
Ok, that's easy enough.

jeff-phillips-18:
@dtaylor113 exactly. The concern is more about cluttering the directory and not being clear which files contain examples vs the actual directives.

dlabrecq:
@jeff-phillips-18, I've moved the new files to their own examples directory.

dtaylor113:
LGTM

jeff-phillips-18:
LGTM

PR #222, opened 2016-04-12
Add optional image to card aggregate status

yaacov:
The provider icons are not provided as css class anymore. But they are available as svg files.
This PR makes it possible to optionally use images, in cases where icons are not available. 

It adds an optional `iconImage` parameter.

screenshot-2016-04-12-15-19-38


yaacov:
@simon3z @zeari 
@jeff-phillips-18 is this a good approach to fix the missing icons ?

jeff-phillips-18:
re: Is it ok to add kubernetes.svg and openshift.svg to the misc directory ?
  The OpenShift and Kubernetes images are available from Patternfly.

simon3z:
> re: Is it ok to add kubernetes.svg and openshift.svg to the misc directory ?
> The OpenShift and Kubernetes images are available from Patternfly.

cc @abonas

yaacov:
@jeff-phillips-18 :+1: 
now using this images.

p.s.
This PR has bad karma because in the html template image src is an angular var :-1:

abonas:
> re: Is it ok to add kubernetes.svg and openshift.svg to the misc directory ?
> The OpenShift and Kubernetes images are available from Patternfly.

@epwinchell can you please sync with @jeff-phillips-18 on this topic so there won't be 2 places to maintain the same svg icons?
cc @simon3z

jeff-phillips-18:
@epwinchell since you already have a good list of svg files I would continue to use yours as you are probably already maintaining them for other uses as well.  I was under the impression this PR was going to be adding the svgs which doesn't seem to be the case.

epwinchell:
@jeff-phillips-18 Sounds good

@yaacov @abonas @simon3z Our SVG vendor icons are up-to-date and available here: https://github.com/ManageIQ/manageiq/tree/master/app/assets/images/svg

yaacov:
@jeff-phillips-18 I just wanted nice icons for the demo, not for manageiq that has it's icons mechanism.
I need help with karma, it's complaining about the image source , being angular variable.
cc @epwinchell

dtaylor113:
@yaacov , you wrote:  "This PR has bad karma because in the html template, image src is an angular var :-( what to do ?", can you elaborate on what the issues is?
Thanks,
- Dave

yaacov:
@dtaylor113  the line:
`<image ng-if="status.iconImage" src="{{status.iconImage}}" alt="" class="card-pf-icon-image"></image>`
makes karma to say:
```
Running "karma:unit" (karma) task
WARN [web-server]: 404: /%7B%7Bstatus.iconImage%7D%7D
```
The preprocessor need to know about it ?

jeff-phillips-18:
@yaacov Change from src="{{  to ng-src="{{

yaacov:
@jeff-phillips-18 Thanks !
This PR has good karma now :-)

yaacov:
@jeff-phillips-18 @dtaylor113 please review

jeff-phillips-18:
Changes LGTM, but speaking of good karma... please add a unit test for the correct setting of the iconImage value.

yaacov:
@jeff-phillips-18 :+1: added a unit test "should set of the iconImage value".

jeff-phillips-18:
LGTM 👍

dtaylor113:
LGTM

PR #225, opened 2016-04-20
Add no-data to linechart

yaacov:
Add _"No Data Available"_ simbol to linechart if no data is available, without this PR it just render an empty div.

**With data**
screenshot-2016-04-20-10-02-50

**No data**
screenshot-2016-04-20-10-09-49


yaacov:
@jeff-phillips-18 @simon3z @zeari please review

dist/angular-patternfly.js
```diff
@@ -1670,6 +1683,7 @@ angular.module('patternfly.charts').directive('pfHeatmap', ["$compile", "$window
        }
 
        $scope.data = {
+         dataAvailable: true,
```
simon3z: @yaacov can't you just detect this by the fact that the data is missing?

yaacov: @simon3z this is how trend-chart and heat-maps works, so all the cards will behave the same.
If we do it this way, it will just work in manageiq without changing anything :-).
Doing it by checking the data will need to change the overview in manageIQ.

simon3z: @yaacov ok, it just seems odd to pass an extra flag to signal whether there's data or not.

In the long run you may want to think if it make sense to drop it everywhere (unless I missed a specific use-case).

yaacov: @simon3z the example will not work without it  :-)

simon3z: > @simon3z the example will not work without it :-)

@yaacov it doesn't sound like an explanation on why it's required. You're just making more evident that there's some code relying on an extra unneeded flag.

jeff-phillips-18: We added the flag in the angular patternfly directive so the application could differentiate between an empty dataset vs data unavailable.

zeari: > We added the flag in the angular patternfly directive so the application could differentiate between an empty dataset vs data unavailable.

@simon3z if you look into the [containers dashboard](https://github.com/zeari/manageiq/blob/396befd12a5ec4b71969cfb32c93a9235c31dcfa/app/assets/javascripts/controllers/container_dashboard/util/charts-mixin.js#L129-129) youll see that a `dataAvailable` flag is the current standard with these patternfly directives.

yaacov:
@jeff-phillips-18 please review

jeff-phillips-18:
LGTM

dtaylor113:
LGTM

PR #226, opened 2016-04-22
fix possible null exception

jstrachan:
without this fix I was getting `TypeError: Cannot set property 'bindto' of undefined` when creating a dynamic list of charts using ng-repeat


dtaylor113:
LGTM.  Simple null check

jeff-phillips-18:
LGTM

jstrachan:
thanks!

PR #227, opened 2016-04-28
Change sortId to currentField in sort directive

karelhala:
In documentatin was type, where it said to use sortId to select element from which items should be filtered, however currentField is the option which shuld be used.

No changes in code, just in usage documentation.


karelhala:
I am not sure whether dist folder should be pushed as well, but since its using bower I assume it's nice to keep changes in dist for upstream, right?

dist/angular-patternfly.js
```diff
@@ -775,10 +775,12 @@ angular.module('patternfly.card').directive('pfCard', function () {
...
-            chart = c3.generate(chartData);
-            if (scope.getChartCallback) {
-              scope.getChartCallback(chart);
+            if (chartData) {
+              chartData.bindto = '#' + attrs.id;
+              chart = c3.generate(chartData);
+              if (scope.getChartCallback) {
+                scope.getChartCallback(chart);
+              }
             }
```
jeff-phillips-18: Looks like an issue with a rebase here. This is removing the change from #226

jeff-phillips-18:
Yes, these dist files do need to be included.

dist/angular-patternfly.js
karelhala: Yeah, I figured that #226 did not included dist folder. I'll keep it in.

jeff-phillips-18:
LGTM 👍

dtaylor113:
LGTM

PR #231, opened 2016-05-05
Add capability to add app defined action buttons to the toolbar

jeff-phillips-18:
Currently, there is a limitation in the toolbar. One cannot add, say, a
dropdown button to the action buttons. This provides the capability for
an application to add buttons how they want to the toolbar action
buttons area.


dtaylor113:
ngDocs needs to be updated with this new config option.

dtaylor113:
The primary and more actions have a 'actionFn' callback.  Just curious, how would someone define a callback for included actions?

jeff-phillips-18:
The application can add ng-click or whatever they want to their HTML. This is a mechanism for the application to provide buttons however they would like to the button area of the toolbar.

jeff-phillips-18:
Added ngdoc

dtaylor113:
Ok, perhaps it would be good to have an example of ng-click and ng-change for the dropdown list.  Just to show how to do that with included actions since it is different from how the other action buttons are handled.

PR #232, opened 2016-05-06
Add capability to add app defined action buttons to the toolbar

jeff-phillips-18:
Currently, there is a limitation in the toolbar. One cannot add, say, a
dropown button to the action buttons. This provides the capability for
an application to add buttons how they want to the toolbar action
buttons area using a named transclusion. This feature is only available
to those using Angular 1.5 and later.

NG-Doc example does not use this feature since its not something we
need to show off and ng docs is not using angular 1.5 which is required
to use this feature.


dtaylor113:
Hi, not sure I agree with the ng-doc conclusion.  angular-patternfly supports ""angular": "1.3.0 - 1.5.*", so one 'could' use this feature if, as you say, they are using angular 1.5.   I still don't know how one is supposed to do a callback function for this new feature?   Having an example would help, even if we specify it is only supported for angular 1.5; perhaps not a working example, but documented in any case.

jeff-phillips-18:
This is simply adding a transclusion. Applications can use it as they see fit. Callbacks would work as any angular JS callback would work. Add ng-click="myFunc()".  When I try to get it to work in ng-doc it doesn't work how it should but using it from an application works fine.

I can look into the reason later but would like to get this in for the next release to allow work to continue. I can add an issue to track down the issue and update the ng-doc.

dtaylor113:
Ok, we can address ng-doc at a later time.

dlabrecq:
LGTM. I don't understand the issue with ng-doc, but I'll defer to Dave.

PR #235, opened 2016-05-16
Add support for filters and track by in pfSelect ngOptions

### LINKED ISSUE #234: pfSelect doesn't support "track by" in ngOptions
When trying to use "track by" expression with pfSelect, an error is thrown.

eg:
`<select pf-select ng-model="pet" id="pet" ng-options="o as o for o in pets track by $index"></select>`
throws
`Error: [$parse:syntax] Syntax Error: Token 'track' is an unexpected token at column 6 of the expression [pets track by $index] starting at [track by $index].`

It should be possible to use it, as it is without the pfSelect directive.

ammendonca:
Fixes #234


ammendonca:
Can someone review and merge this ? It's blocking the usage of the component with such options.

jeff-phillips-18:
LGTM

dtaylor113:
LGTM

PR #240, opened 2016-06-01
Add Fixed Accordion Utility

jeff-phillips-18:


jeff-phillips-18:
@dtaylor113 @dlabrecq Please review

dtaylor113:
Not sure why 'pfFixedAccordion' is in 'patternfly.utils', seems like it should be in just 'patternfly'.
'patternfly' currently has the following directives: pfSelect, pfFocused, pfValidation; seems like pfFixedAccordion would fit here.

jeff-phillips-18:
I've modified it to show up in the base patternfly area of ngDocs.

dlabrecq:
LGTM

PR #249, opened 2016-06-13
Fix for notification drawer open/close styling

jeff-phillips-18:
@dlabrecq @dtaylor113 Please review

The previous fix for open/close styling for Angular Bootstrap accordions had the potential to break the styling for any non-Angular Bootstrap accordions. Modified the selectors such that they only effect angular bootstrap accordions.


dtaylor113:
LGTM

dlabrecq:
LGTM

PR #253, opened 2016-06-24
Modify gitignore so generated docs will be included (and include all …

dgutride:
…generated docs after ngdocs:publish)

The only code change is in the .gitignore file, everything else is generated.


dtaylor113:
I see Notification Draw partial, LGTM :thumbsup:

jeff-phillips-18:
LGTM

PR #254, opened 2016-06-29
Add no data available state to sparkline and donut chart

cshinn:
I've added the no data available checkbox and behavior to the sparkline and donut charts.

### DIFF
--- a/src/charts/donut/donut-pct-chart.html
+++ b/src/charts/donut/donut-pct-chart.html
@@ -1,3 +1,4 @@
 <span>
-  <div pf-c3-chart id={{donutChartId}} config="config"></div>
+  <div pf-c3-chart ng-if="data.dataAvailable !== false" id={{donutChartId}} config="config"></div>
+  <div pf-empty-chart ng-if="data.dataAvailable === false" chart-height="chartHeight"></div>
 </span>
--- a/src/charts/donut/donut-pct-chart-directive.js
+++ b/src/charts/donut/donut-pct-chart-directive.js
@@ -29,8 +29,9 @@
  *
  * @param {object} data the Total and Used values for the donut chart.  Available is calculated as Total - Used.<br/>
  * <ul style='list-style-type: none'>
- * <li>.used   - number representing the amount used
- * <li>.total  - number representing the total amount
+ * <li>.used          - number representing the amount used
+ * <li>.total         - number representing the total amount
+ * <li>.dataAvailable - Flag if there is data available - default: true
  * </ul>
  *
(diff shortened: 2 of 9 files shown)


dtaylor113:
Hi, downloaded and ran this branch, sparkline and donut charts with "no data" looks good.  However, unit tests need to be added for this new functionality.  Please see angular-patternfly/test/charts/trends/trends-chart.spec.js,  the last "it("should show empty chart when the dataAvailable is set to false" test should be added to the donut and sparkline unit tests.  -thanks :-)

src/charts/donut/donut-pct-chart.html
```diff
@@ -1,3 +1,4 @@
 <span>
-  <div pf-c3-chart id={{donutChartId}} config="config"></div>
+  <div pf-c3-chart ng-if="data.dataAvailable !== false" id={{donutChartId}} config="config"></div>
+  <div pf-empty-chart ng-if="data.dataAvailable === false" chart-height="172"></div>
```
jeff-phillips-18: 172 will work for the default donut size (set in $().c3ChartDefaults().getDefaultDonutConfig()), but if the caller specified a different size, this size will not work. This should either be a parameter to the directive, or computed based of the donut chart height.

LHinson:
Nice seeing you here @cshinn :) In addition to the unit testing, we probably need to add a bit of design documentation.

LHinson:
ignore me @cshinn , just saw this one - https://github.com/patternfly/patternfly-design/pull/1#issuecomment-230297809

src/charts/donut/donut-pct-chart.html
cshinn: @jeff-phillips-18 The Utilization trend chart also has the height hard coded for that parameter because neither of those directives contain a chartHeight field. Should that be added into them?

cshinn:
I've updated the PR to include the utilization bar chart as well as unit tests for all three charts.

dtaylor113:
LGTM

dlabrecq:
LGTM. Has the chartHeight issue been resolved?

cshinn:
@dlabrecq I just updated the PR to fix the chartHeight issue for the donut chart. Utilization bar chart looks like it's going to need another design pass, because of the narrow bar width, and the fact that they are often grouped together.

dlabrecq:
Great.

Anyone else have comments before a merge?

jeff-phillips-18:
The commits need to be squashed

jeff-phillips-18:
LGTM

PR #268, opened 2016-07-25
Added highlight to all blocks within same range

dlabrecq:
Added highlight to all blocks within same range when block size is less than 16.
PTNFLY-1226

### DIFF
--- a/src/charts/heatmap/heatmap.directive.js
+++ b/src/charts/heatmap/heatmap.directive.js
@@ -18,10 +18,14 @@
  * @param {boolean=} showLegend flag to show the legend, defaults to true
  * @param {array=} legendLabels the labels for the legend - defaults: ['< 70%', '70-80%' ,'80-90%', '> 90%']
  * @param {number=} maxBlockSize the maximum size for blocks in the heatmap. Default: 50, Range: 5 - 50
+ * @param {number=} minBlockSize the minimum size for blocks in the heatmap. Default: 2
  * @param {number=} blockPadding the padding in pixels between blocks (default: 2)
  * @param {array=} thresholds the threshold values for the heapmap - defaults: [0.7, 0.8, 0.9]
  * @param {array=} heatmapColorPattern the colors that correspond to the various threshold values (lowest to hightest value ex: <70& to >90%) - defaults: ['#d4f0fa', '#F9D67A', '#EC7A08', '#CE0000']
  * @param {function=} clickAction function(block) function to call when a block is clicked on
+ * @param {number=} rangeHoverSize the maximum size for highlighting blocks in the same range. Default: 15
+ * @param {boolean=} rangeOnHover flag to highlight blocks in the same range on hover, defaults to true
+ * @param {array=} rangeTooltips the tooltips for blocks in the same range - defaults: ['< 70%', '70-80%' ,'80-90%', '> 90%']
  * @example
  <example module="patternfly.charts">
    <file name="index.html">
@@ -31,12 +35,16 @@
            <div pf-heatmap id="id" chart-title="title" data="data" chart-data-available="dataAvailable"
                 show-legend="showLegends"></div>
          </div>
-         <div class="col-md-5 example-heatmap-container">
+         <div class="col-md-3 example-heatmap-container">
            <div pf-heatmap id="id" chart-title="titleAlt" data="data" chart-data-available="dataAvailable"
                 show-legend="showLegends" legend-labels="legendLabels"  max-block-size="20" block-padding="5"
                 heatmap-color-pattern="heatmapColorPattern" thresholds="thresholds"
                 click-action="clickAction"></div>
          </div>
+         <div class="col-md-3 example-heatmap-container">
+           <div pf-heatmap id="id" chart-title="titleSmall" data="data" chart-data-available="dataAvailable"
+                show-legend="showLegends" max-block-size="15" range-tooltips="rangeTooltips"></div>
+           </div>
        </div>
        <div class="row">
          <div class="col-md-3">
@@ -119,7 +127,9 @@
        $scope.dataAvailable = true;
        $scope.title = 'Utilization - Using Defaults';
        $scope.titleAlt = 'Utilization - Overriding Defaults';
+       $scope.titleSmall = 'Utilization - Small Blocks';
        $scope.legendLabels = ['< 60%','70%', '70-80%' ,'80-90%', '> 90%'];
+       $scope.rangeTooltips = ['Memory Utilization < 70%<br\>40 Nodes', 'Memory Utilization 70-80%<br\>4 Nodes', 'Memory Utilization 80-90%<br\>4 Nodes', 'Memory Utilization > 90%<br\>4 Nodes'];
        $scope.thresholds = [0.6, 0.7, 0.8, 0.9];
        $scope.heatmapColorPattern = ['#d4f0fa', '#F9D67A', '#EC7A08', '#CE0000', '#f00'];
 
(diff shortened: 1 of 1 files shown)


dlabrecq:
screen shot 2016-07-25 at 11 45 23 am

src/charts/heatmap/heatmap.directive.js
```diff
@@ -271,8 +286,20 @@ angular.module('patternfly.charts').directive('pfHeatmap', function ($compile, $
 
         //Adding events
         blocks.on('mouseover', function () {
+          var fillColor;
           blocks.call(highlightBlock, false);
-          d3.select(this).call(highlightBlock, true);
+          if (scope.maxSize < 16) {
```
jeff-phillips-18: This should be blockSize (computed) not scope.maxSize.

jeff-phillips-18: 16 might be a good default, but the application should be able to set that value if desired.

src/charts/heatmap/heatmap.directive.js
```diff
@@ -271,8 +286,20 @@ angular.module('patternfly.charts').directive('pfHeatmap', function ($compile, $
...
+            // Get fill color for current block
+            fillColor = color(d3.select(this).map(function (d) {
+              return d[0].__data__.value;
+            }));
+            // Highlight all blocks matching fill color
+            blocks[0].forEach(function (block) {
+              highlightBlockColor(d3.select(block), fillColor);
+            });
+          } else {
```
jeff-phillips-18: I believe we want to set the tooltip based on the range as well so that there is one tooltip for the entire range not per node. This would mean we need to ignore the tooltip field on the nodes and provide a means for the application to give us the tooltip based on the current range being hovered on.

dlabrecq: Right. Will try to add this.

src/charts/heatmap/heatmap.directive.js
dlabrecq: I was looking at the maxSize attribute and we don't allow it to be less than 5 or greater than 50. We don't allow the application to override those defaults, so I'm not sure we should do that here?

$scope.maxSize = parseInt($scope.maxBlockSize);
        if ($scope.maxSize < 5) {
          $scope.maxSize = 5;
        } else if ($scope.maxSize > 50) {
          $scope.maxSize = 50;
        }

dlabrecq: FYI, I added the maxBlockHoverSize attribute to override the default value. However, I'm not using blockSize because that adds padding. When the maxBlockSize attribute is 15, blockSize equals 20. If you inspect this via the browser, the physical block is actually 16x16. In this case, blockSize is greater than the default and highlighting wouldn't be enabled.

This is more problematic when using the maxBlockHoverSize attribute to override the default value. For example, highlighting would be disabled when max-block-size="15" and max-block-hover-size="16" because blockSize is computed to be 20. I think this would be too confusing to someone trying to override these values.

jeff-phillips-18: But scope.maxSize is passed in, it does not mean that is the size of the blocks. For example, an application may  want a maximum block size of 100. That means that even if there are only 2 data items, only use 100x100px blocks, do not fill the entire area with huge blocks. We set the max block size small in these examples to show small blocks. 

If I set a maxSize of 100 but there are 10,000,000 blocks, the computed block size will be small and THAT is when we want to use this functionality.

jeff-phillips-18: BTW, in that method (redraw) you can use fillSize if you want the size of the block without padding.

dlabrecq: Thanks for clarifying. I think what I really want to test is blockSize - scope.padding (i.e., the fillSize var). It wasn't correct having to set max-block-size="13" in order for blockSize to be less than 16 and highlighting finally becomes enabled.

dlabrecq:
Added tooltips specific to range and met with @LHinson to review.

screen shot 2016-07-27 at 3 00 15 pm

dlabrecq:
@jeff-phillips-18 would you mind taking another look?

src/charts/heatmap/heatmap.directive.js
```diff
@@ -246,14 +261,34 @@ angular.module('patternfly.charts').directive('pfHeatmap', function ($compile, $
...
+            return d[0].__data__.value;
+          }));
+          // If given color matches, apply highlight
+          if (blockColor === fillColor) {
+            block.style('fill-opacity', 1);
+          }
+        };
+
+        // Highlight range on hover when default > fill size or when rangeOnHover is true
+        var highlightRange = ((scope.rangeOnHover === undefined && fillSize < 16) || scope.rangeOnHover) ? true : false;
```
jeff-phillips-18: I believe that 16 is a good default but should be customizable by the application

src/charts/heatmap/heatmap.directive.js
```diff
@@ -246,14 +261,34 @@ angular.module('patternfly.charts').directive('pfHeatmap', function ($compile, $
...
+
         var svg = window.d3.select(thisComponent);
         svg.selectAll('*').remove();
 
+        // Highlight range on hover when default > fill size or when rangeOnHover is true
+        if ((scope.rangeOnHover === undefined && fillSize < 16) || scope.rangeOnHover) {
+          highlightRange = true;
+        }
+
```
jeff-phillips-18: This was set above via the same logic. Though I don't really understand the "when default > fill size"

dlabrecq: I meant when the fill size is less than the default value of 16.

If rangOnHover has not been set, the highlighting will be enabled when "when the fill size is less than the default value of 16". Otherwise rangOnHover=true/false overrides that.

src/charts/heatmap/heatmap.directive.js
dlabrecq: Currently, the feature can be forced on using rangeOnHover=true. This overrides the default value of 16. Similarly, it can be completely disabled using rangeOnHover=false.

I actually had both max-block-size="15" and max-block-hover-size="16" in my initial example. However, it wouldn't be enabled unless max-block-size < max-block-hover-size. It didn't make a lot of sense/confusing to have both. I suppose it might be less confusing if max-block-size wasn't used.

Should I can add this back in?

dlabrecq:
@jeff-phillips-18 I added the minBlockSize and rangeHoverSize attributes per our discussion. The rangeOnHover=false will still disable the feature.

src/charts/heatmap/heatmap.directive.js
```diff
@@ -144,25 +154,36 @@ angular.module('patternfly.charts').directive('pfHeatmap', function ($compile, $
...
     controller: function ($scope) {
       var thresholdDefaults = [0.7, 0.8, 0.9];
       var heatmapColorPatternDefaults = ['#d4f0fa', '#F9D67A', '#EC7A08', '#CE0000'];
       var legendLabelDefaults = ['< 70%', '70-80%' ,'80-90%', '> 90%'];
+      var rangeTooltipDefaults = ['< 70%', '70-80%' ,'80-90%', '> 90%'];
       var heightDefault = 200;
 
       //Allow overriding of defaults
+      if ($scope.minBlockSize === undefined || isNaN($scope.minBlockSize)) {
+        $scope.minSize = 5;
```
jeff-phillips-18: The design spec says "Recommended:  Minimum block size is 2x2"

src/charts/heatmap/heatmap.directive.js
```diff
@@ -174,6 +195,23 @@ angular.module('patternfly.charts').directive('pfHeatmap', function ($compile, $
...
+      } else {
+        $scope.rangeHoverSize = parseInt($scope.rangeHoverSize);
+        if ($scope.rangeHoverSize < $scope.minSize) {
+          $scope.rangeHoverSize = $scope.minSize;
+        } else if ($scope.rangeHoverSize > $scope.maxSize) {
+          $scope.rangeHoverSize = scope.maxSize;
+        }
+      }
+
+      $scope.rangeOnHover = ($scope.rangeOnHover === undefined) ? true : false;
```
jeff-phillips-18: Passing range-on-hover="false" will result in rangeOnHover being true.

src/charts/heatmap/heatmap.directive.js
jeff-phillips-18: I don't see scope.minBlockSize being used anywhere. After the call to determineBlockSize, the maxBlockSize is checked but I don't see a minBlockSize check.

dlabrecq: I read this in Google doc; “In angular patternfly we set the max size to be 50px but the application can set the max for a particular heat map to be from 5-50”. I assumed we don't want to make it smaller than that?

Do we want to allow it to be 2x2 only when range hover is enabled?

Regarding the check after determineBlockSize, we never had a test for min block size. I'll look at adding one.

src/charts/heatmap/heatmap.directive.js
dlabrecq: Good catch.

src/charts/heatmap/heatmap.directive.js
jeff-phillips-18: I think it is OK to have a max range of 5-50 and still set the min to 2. I believe we should follow the spec regardless of range hover being enabled.

dlabrecq:
Added a check of scope.minSize (i.e., minBlockSize) after calling determineBlockSize, similar to the scope.maxSize test.

dlabrecq:
@dtaylor113 @jeff-phillips-18 would you please take a look?

jeff-phillips-18:
LGTM

priley86:
👍

PR #272, opened 2016-08-03
bootstrap-select tests

priley86:
A couple things happening with this PR for release prep.

* Referencing #master-dist for testing until release is complete:
```
"patternfly": "git://github.com/patternfly/patternfly.git#master-dist"
```
* Re-referencing `patternfly-bootrap-treeview` after [#382](https://github.com/patternfly/patternfly/pull/382) in Patternfly
* Noticed that `heatmap.spec.js` would fail when running `karma start` and debugging in Chrome due to styles being converted to rgb
* Fixes tests in `select.spec.js` due to `bootstrap-select` dependency upgrade in Patternfly [#365](https://github.com/patternfly/patternfly/pull/365)

Bootstrap-select now renders select options within the `bootstrap-select` div as of `v1.10.0`, so the sibling selector should now be `.dropdown-menu`.

i.e. new rendering looks like this:
```
<div class="btn-group bootstrap-select">
	<button type="button" class="btn dropdown-toggle btn-default"></button>
	<div class="dropdown-menu open">
		<ul class="dropdown-menu inner" role="menu">
			<li></li>
			<li></li>
			<li></li>
		</ul>
	</div>
	<select pf-select="" ng-model="modelValue" ng-options="o as o for o in options">
		<option></option>
		<option></option>
		<option></option>
	</select>
</div>
```

old rendering looks like this:
```
 <select pf-select="" ng-model="modelValue" ng-options="o as o for o in options">
	<option></option>
	<option></option>
	<option></option>
</select>
<div class="btn-group bootstrap-select">
	<button type="button" class="btn dropdown-toggle btn-default"></button>
	<div class="dropdown-menu open">
		<ul class="dropdown-menu inner" role="menu">
			<li></li>
			<li></li>
			<li></li>
		</ul>
	</div>
</div>
```

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -39,7 +39,7 @@
     "angular-sanitize": "1.3.0 - 1.5.*",
     "angular-bootstrap": "0.13.x",
     "lodash": "3.x",
-    "patternfly": "~3.7.0"
+    "patternfly": "~3.8.0"
   },
   "devDependencies": {
     "angular-mocks": "1.3.0 - 1.5.*"
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -129,7 +129,7 @@ module.exports = function (grunt) {
             'lib/bootstrap-combobox/js/bootstrap-combobox.js',
             'lib/bootstrap-datepicker/dist/js/bootstrap-datepicker.js',
             'lib/bootstrap-select/js/bootstrap-select.js',
-            'lib/bootstrap-treeview/src/js/bootstrap-treeview.js',
+            'lib/patternfly-bootstrap-treeview/src/js/bootstrap-treeview.js',
             'lib/c3/c3.js',
             'lib/d3/d3.js',
             'lib/patternfly/dist/js/patternfly.js',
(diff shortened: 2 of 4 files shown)


bower.json
```diff
@@ -39,7 +39,7 @@
     "angular-sanitize": "1.3.0 - 1.5.*",
     "angular-bootstrap": "0.13.x",
     "lodash": "3.x",
-    "patternfly": "~3.7.0"
+    "patternfly": "git://github.com/patternfly/patternfly.git#master-dist"
```
jeff-phillips-18: I question the need to do this. If Angular Patternfly has a dependency on bootstrap-select (and specifically now on 1.10) then shouldn't this dependency be here with the necessary version?

priley86: @jeff-phillips-18 This was just temporary while I was testing last night. I'm new to our release process - should we update this reference after we publish our new Patternfly release?

dlabrecq:
LGTM. Although, considering we're releasing Patternfly today, we'll need to verify this dependency is changed when releasing angular-pattenrfly. The scripts should make the change automatically.

dlabrecq:
LGTM

PR #276, opened 2016-08-05
Add pfVerticalNavigation directive.

jeff-phillips-18:
Directive for vertical navigation.

This sets up the nav bar header with the collapse button (hamburger)
and the application brand image (or text) as well as the vertical
navigation menu containing the navigation items.

This  directive supports primary, secondary, and tertiary
navigation with options to allow pinning of the secondary and
tertiary navigation menus as well as the option for persistent
secondary menus.

### DIFF
--- a/src/navigation/vertical-navigation-directive.js
+++ b/src/navigation/vertical-navigation-directive.js
@@ -1,0 +1,997 @@
+/**
+ * @ngdoc directive
+ * @name patternfly.navigation.directive:pfVerticalNavigation
+ *
+ * @description
+ *   Directive for vertical navigation. This sets up the nav bar header with the collapse button (hamburger) and the
+ *   application brand image (or text) as well as the vertical navigation bar containing the navigation items. This
+ *   directive supports primary, secondary, and tertiary navigation with options to allow pinning of the secondary and
+ *   tertiary navigation menus as well as the option for persistent secondary menus.
+ *   <br><br>
+ *   The remaining parts of the navbar header can be transcluded.
+ *   <br><br>
+ *   Tha navigation items are marked active based on the current location and the href value for the item. If not using
+ *   href's on the items to navigate, set update-active-items-on-click to "true".
+ *   <br><br>
+ *   This directive works in conjunction with the main content container if the 'container-pf-nav-pf-vertical' class
+ *   selector is added to the main content container.
+ *
+ * @param {string} brandSrc src for brand image
+ * @param {string} brandAlt  Text for product name when brand image is not available
+ * @param {boolean} hasSubMenus Flag if there are secondary and/or tertiary navigation items, default: true
+ * @param {boolean} persistentSecondary Flag to use persistent secondary menus, default: false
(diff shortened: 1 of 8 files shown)


jeff-phillips-18:
@dgutride @dtaylor113  Please review

Please try out the NG Docs for this and give me any usability thoughts. Due to the full page nature of vertical navigation, I had to take some liberties to be able to demo this in the NG Docs.

src/navigation/vertical-navigation-directive.js
```diff
@@ -0,0 +1,1006 @@
+/**
+ * @ngdoc directive
+ * @name patternfly.navigation.directive:pfVerticalNavigation
+ *
+ * @description
+ *   Directive for vertical navigation. This sets up the nav bar header with the collapse button (hamburger) and the
+ *   application brand image (or text) as well as the vertical navigation bar containing the navigation items. This
+ *   directive supports primary, secondary, and tertiary navigation with options to allow pinning of the secondary and
+ *   tertiary navigation menus as well as teh option for persistent secondary menus.
```
dgutride: teh option...

dgutride:
Pulled/locally - built and ran without issues.  One thought on the demo area - I like how it takes you in and out of that mode, but maybe a small bit of text explaining that they will be taken to a new window and can return from there could make it a little less jarring.  The top button once inside was very clear, though so it's pretty obvious how they can get back there.  One other option is to launch it in a new window (if that's possible) as long as we tell them we are doing that.  Just one spelling issue I saw in the help text.

jeff-phillips-18:
@dgutride I thought about launching a window but I'm unsure how to get that to work such that the files are served up for the new window. All the components are added via the ng-docs and I don't know how to make it available to a new window.

I'm not exactly sure what I could add for text to explain that the current content of the page will be hidden while in demo mode.

src/navigation/vertical-navigation-directive.js
jeff-phillips-18: Fixed, thanks!

jeff-phillips-18:
Updated to fix @dgutride issue (mis-spelling)

src/navigation/vertical-navigation-directive.js
```diff
@@ -0,0 +1,1006 @@
...
+                    }
+                  });
+                }
+              });
+            }
+          });
+        };
+
+        if (!$scope.updateActiveItemsOnClick) {
+          rootScope.$on( "$routeChangeSuccess", function (event, next, current) {
```
dgutride: This event should be disposed of when the directive is removed.

dtaylor113:
Going to test this out soon....

jeff-phillips-18:
Added call to remove route change listener on destroy (  @dgutride issue above )

jeff-phillips-18:
Added a blurb to forewarn about the content being covered and how to get back ( @dgutride @dtaylor113  comments ).
Moved styling into example.css

dgutride:
This looks good to me.

jeff-phillips-18:
@dtaylor113 Please re-review

dtaylor113:
LGTM

PR #277, opened 2016-08-09
update package.json

priley86:
* updates package.json with dependencies (so npm published distrubution is accurate) for [PTNFLY-1626](https://patternfly.atlassian.net/browse/PTNFLY-1626)
* add `.npmignore` to allow `dist` files in npm distribution
* add index.js to allow `Browserify` or CommonJS support. Noted [here](http://blog.npmjs.org/post/114584444410/using-angulars-new-improved-browserify-support)

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,8 +4,17 @@
   "version": "3.8.1",
   "license": "Apache-2.0",
   "description": "Angular extension of the PatternFly project.",
+  "keywords": ["angular", "patternfly"],
+  "main": "index.js",
   "homepage": "https://github.com/patternfly/angular-patternfly",
-  "dependencies": {},
+  "dependencies": {
+    "angular": "1.3.0 - 1.5.*",
+    "angular-animate": "1.3.0 - 1.5.*",
+    "angular-sanitize": "1.3.0 - 1.5.*",
+    "angular-ui-bootstrap": "0.13.x",
+    "lodash": "3.x",
+    "patternfly": "~3.8.1"
+  },
   "devDependencies": {
     "express": "3.4.4",
     "grunt": "0.4.5",
--- a/.npmignore
+++ b/.npmignore
@@ -22,7 +22,6 @@ dist/docs/index.html
 
 
 /docs
-/dist
 node_modules
 npm-debug.log
 lib
--- a/index.js
+++ b/index.js
@@ -1,0 +1,8 @@
+require('angular');
+require('angular-animate');
+require('angular-sanitize');
+require('angular-ui-bootstrap');
+require('lodash');
+require('../patternfly/dist/js/patternfly.min');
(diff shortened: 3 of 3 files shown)


priley86:
@dgutride @dlabrecq @jeff-phillips-18 please take a look

package.json
```diff
@@ -4,8 +4,17 @@
...
+  "keywords": ["angular", "patternfly"],
+  "main": "index.js",
   "homepage": "https://github.com/patternfly/angular-patternfly",
-  "dependencies": {},
+  "dependencies": {
+    "angular": "1.3.0 - 1.5.*",
+    "angular-animate": "1.3.0 - 1.5.*",
+    "angular-sanitize": "1.3.0 - 1.5.*",
+    "angular-ui-bootstrap": "0.13.x",
+    "lodash": "3.x",
```
dtaylor113: Hmm...there was some talk a while ago about consumers not wanting to be required to use lodash.  At the time we coded some functions to test to see if lodash was installed, if not, do the function in plain angularjs.

priley86: @dtaylor113 i'm not positive, but I believe angular-p.f. still depends on Lodash. @dlabrecq can you confirm?

jeff-phillips-18:
Agreed, the strategy was to not require lodash. Hopefully nothing was added to make it a requirement.

package.json
dlabrecq: I see lodash in bower.json.

priley86:
i've updated `index.js` and added a simple example of using angular-patternfly with Browserify [here](https://github.com/priley86/angular-browserify)

package.json
priley86: It appears Lodash is still in use. I see references to `._remove`, `_.findWhere` and `_.find` which are not scoped like the references to `_.merge` which check for `angular.merge` first.

I think we should address this in a separate PR.

dtaylor113: @priley86 Ok on addressing that in a separate PR.  Please create a JIRA story for this issue.

priley86: Sure! I've created [PTNFLY-1711](https://patternfly.atlassian.net/browse/PTNFLY-1711) to address this. Let me know if this works.

priley86:
any more thoughts here ?

jeff-phillips-18:
I'm OK with this then.

priley86:
@jeff-phillips-18 @dtaylor113  I'm happy to make these updates for lodash in a future PR. Just didn't want to conflate it.

priley86:
@jeff-phillips-18 @dtaylor113 ok to merge?

dtaylor113:
I'm ok with this as well.  The lodash requirement would only be for the npm only install, correct?  If so, then this won't break any existing customers ('cause angular-pf doesn't have an npm yet :-)

PR #278, opened 2016-08-10
Locked version of net-ssh due to deploy failure for OpenShift.

dlabrecq:
This is the same issue we recently encountered with pf.org.


priley86:
👍  good find and happy to see this fix merged

dgutride:
Looks good

PR #279, opened 2016-08-11
Add pfToastNotification and pfToastNotificationList Directives

jeff-phillips-18:
This also fixes the notification service to:

1) appropriately timeout on each toast notification and dismiss them singularly. Also does not dismiss the toast notification if the user is currently hovering over it (viewing).

2) provide the list of notifications via the notification service rather than requiring the use of $rootScope.

### DIFF
--- a/src/notification/notification.js
+++ b/src/notification/notification.js
@@ -83,7 +83,7 @@ angular.module('patternfly.notification').provider('Notifications', function () 
   'use strict';
 
   // time (in ms) the notifications are shown
-  this.delay = 5000;
+  this.delay = 8000;
   this.verbose = true;
   this.notifications = {};
   this.persist = {'error': true, 'httpError': true};
@@ -108,18 +108,6 @@ angular.module('patternfly.notification').provider('Notifications', function () 
     var verbose = this.verbose;
     var persist = this.persist;
 
-    var scheduleMessagePop = function () {
-      $timeout(function () {
-        var i;
-
-        for (i = 0; i < $rootScope.notifications.data.length; i++) {
-          if (!$rootScope.notifications.data[i].isPersistent) {
-            $rootScope.notifications.data.splice(i, 1);
-          }
-        }
-      }, delay);
-    };
-
     var modes = {
       info: { type: 'info', header: 'Info!', log: 'info'},
       success: { type: 'success', header: 'Success!', log: 'info'},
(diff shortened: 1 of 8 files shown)


dtaylor113:
When I turn on the close and kabob options, their icons overlap:
image

dtaylor113:
Very small nit, the Actions: textarea in the demo doesn't scroll when new msgs are added, so after the first three actions I couldn't tell if actions were working 'cause the actions textarea didn't appear to update.

dtaylor113:
Doc talks about an 8 second delay, does this demo show this?  It would be kinda cool if we had a countdown timer in the demo which when reached 0 the notification would go away.

dtaylor113:
LGTM

jeff-phillips-18:
The documentation states that having close and menuActions is not supported. I suppose we could hide the close  button if there are menu actions but this is what base patternfly gives.

jeff-phillips-18:
Ah yes... in other doc pages we put the new messages above older messages. Will fix.

jeff-phillips-18:
The demo does show the notifications disappearing after 8 seconds if not persistent. Also shows that if the user is hovering on a notification it is not dismissed after the 8 seconds until the user moves away.

jeff-phillips-18:
Addressed the overlapping kebab and close buttons and updated documentation to say that the close button will not be show ( @dtaylor113 issue)

Addressed the event text issue ( @dtaylor113 )

jeff-phillips-18:
@dgutride @dlabrecq Please review/merge if you get a few moments.

src/notification/notification.js
```diff
@@ -172,6 +180,23 @@ angular.module('patternfly.notification').provider('Notifications', function ()
...
+    };
+
+    notifications.setViewing = function (notification, viewing) {
+      notification.viewing = viewing;
+      if (!viewing && !notification.show) {
+        notifications.remove(notification);
+      }
+    };
+
+    notifications.data = $rootScope.notifications.data;
```
priley86: @jeff-phillips-18 curious why we still need to use $rootScope to pass thru the data?  Can we not just reference `Notifications.notifications.data` in `pfNotificationList` directive? I guess I'm OK with exposing them in $rootScope, just wasn't quite sure based on the example.

jeff-phillips-18: The use of $rootScope is from prior to this PR. Since there are (or at least could be) applications built using the Notifications service prior to this, I didn't want to remove it and break them or cause side effects. I didn't much care for its use either so I made it such that applications do not have to use it.

priley86: @jeff-phillips-18 thanks for the explanation.

OK with merging after rebase. 👍

PR #297, opened 2016-08-18
add build steps before publish

priley86:
* added build steps for next time


priley86:
@dlabrecq please review

dlabrecq:
We don’t need to build because the script creates a new clone, then does a checkout of master-dist. That’s all we need because the branch has been pre-built by Travis. See #298.

PR #298, opened 2016-08-18
Removed -B switch from git checkout.

dlabrecq:
We used the -B switch with git checkout. The switch will create the branch if it doesn’t exist, but for some reason, it omits the dist directory. Simply removing the -B switch fixes that issue.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -12,6 +12,7 @@ default()
 
   BOWER_JSON=bower.json
   PACKAGE_JSON=package.json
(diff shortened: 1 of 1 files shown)


priley86:
@dlabrecq i'm ok with this since we now have: 
```
  p) PUBLISH=1
          BRANCH=master-dist;;		          BRANCH=master-dist;;
```

I will close #297 since this should address the problem.

priley86:
👍

PR #303, opened 2016-08-29
Add ability to include badges on nav items in pfVerticalNavigation

jeff-phillips-18:
This also fixes an issue where the body content might not have been
setup at startup. Rather than getting it once and holding a reference,
get the element each time

Removed the hasSubmenus setting since this is no longer required. No backward compatibility issue since setting this has no effect.

This depends on https://github.com/patternfly/patternfly/pull/439 getting merged into Patternfly.

image

### DIFF
--- a/misc/examples.css
+++ b/misc/examples.css
@@ -123,3 +123,10 @@ hr {
 .example-info-text:first-of-type {
   margin-top: 10px;
 }
+
+.example-error-background {
+  background-color: #cc0000 !important;
+}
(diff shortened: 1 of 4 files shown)


jeff-phillips-18:
@dtaylor113 @dgutride @dlabrecq  Please review

dtaylor113:
LGTM

dgutride:
Look good

dlabrecq:
A little late, but LGTM.

PR #310, opened 2016-09-06
Bug/307 missing support for translations

AllenBW:
Updates patternfly components with hard coded text to support `translation` attribute.


dgutride:
Hi Allen - thanks for submitting this PR.  I'm not sure if this will fit into our current support for localization so to kick off the conversation - using the translate attribute implies use of angular-translate (correct)?  Our current support for localization is to let people pass in strings to the directives presuming they are already translated by the time they reach the directive.  

Adding some of the others on the PatternFly team to get them into this conversation too  @bleathem , @dlabrecq  - Can you confirm my understanding of the translate attribute above?

dtaylor113:
Seems like the new translate attribute is causing unit test failures:
PhantomJS 1.9.8 (Linux) Directive:  pfToolbar should have correct number of results FAILED
	Expected '5 <span translate="">Results</span>' to be '5 Results'.
	Error: Expected '5 <span translate="">Results</span>' to be '5 Results'.

AllenBW:
@dtaylor113 yep!  working that test presently 👍

dtaylor113:
Hi @AllenBW , unfortunately the use of 'translate' in form-buttons.html is not the norm.  In fact it is the only directive in all of angular-patternfly which has the translate tag.   Forms-buttons.html was probably one of the first directives added to angular-patternfly before UXD took it over.  As such, it is not part of a-pf best practices, and should not be used as an example.

dtaylor113:
Hi @AllenBW, the 'translate' attribute used in Forms-buttons.html seems to be referring to an HTML 5 translate attribute, which seems to be a boolean value: 
"The translate attribute in HTML5 indicates that the content of the element should or should not be translated. There is no effect on the rendered page"
https://www.w3.org/International/questions/qa-translate-flag.en

Was this your intended use of 'translate'?   Sorry, I'm just getting up to speed on the 'translate' attribute.  Looking at SSUI, the translate attribute seems to be going to the HTML 5 translate attribute as well, so maybe that is the intended usage.  I thought it was something like an 'angular-translate' dependency.

AllenBW:
Still working on understanding it as well @dtaylor113, you have my general intent correct.  `translate="yes"` is equivalent to `translate`, indicates a field is in need of translation.  As I am sure you have notice, no browsers currently support the attribute, (so what am I doing with this pr!?) http://zanata.org/ (presently used in ssui) also pays attention to `translate` and plugs in translations

PR #320, opened 2016-09-27
PatternFly Angular wizard implementation

dgutride:
This PR includes the initial implementation, documentation and unit tests for the angular-patternfly wizard.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -42,7 +42,7 @@
     "angular-sanitize": "1.3.0 - 1.5.*",
     "angular-bootstrap": "0.13.x",
     "lodash": "3.x",
-    "patternfly": "git@github.com:patternfly/patternfly.git#master-dist"
+    "patternfly": "http://github.com/patternfly/patternfly.git#master-dist"
   },
   "devDependencies": {
     "angular-mocks": "1.3.0 - 1.5.*"
--- a/bower.json
+++ b/bower.json
@@ -5,7 +5,10 @@
     "Red Hat"
   ],
   "description": "Angular extension of the PatternFly project.",
-  "keywords": [ "angular", "patternfly"],
+  "keywords": [
+    "angular",
+    "patternfly"
+  ],
   "license": "Apache-2.0",
   "ignore": [
     "node_modules",
@@ -39,9 +42,17 @@
     "angular-sanitize": "1.3.0 - 1.5.*",
     "angular-bootstrap": "0.13.x",
     "lodash": "3.x",
-    "patternfly": "~3.11.0"
+    "patternfly": "a014002ebcfad2c6ff2fbeda52dfcd425e016f2a"
   },
   "devDependencies": {
     "angular-mocks": "1.3.0 - 1.5.*"
+  },
+  "resolutions": {
+    "bootstrap-datepicker": "~1.6.4",
+    "bootstrap-select": "~1.10.0",
+    "bootstrap-touchspin": "~3.1.1",
+    "datatables-colreorder": "~1.3.2",
+    "font-awesome": "~4.6.3",
+    "moment": "~2.14.1"
   }
 }
--- a/bower.json
+++ b/bower.json
@@ -42,17 +42,9 @@
     "angular-sanitize": "1.3.0 - 1.5.*",
     "angular-bootstrap": "0.13.x",
     "lodash": "3.x",
-    "patternfly": "a014002ebcfad2c6ff2fbeda52dfcd425e016f2a"
+    "patternfly": "git@github.com:patternfly/patternfly.git#master-dist"
   },
   "devDependencies": {
     "angular-mocks": "1.3.0 - 1.5.*"
-  },
-  "resolutions": {
-    "bootstrap-datepicker": "~1.6.4",
-    "bootstrap-select": "~1.10.0",
-    "bootstrap-touchspin": "~3.1.1",
-    "datatables-colreorder": "~1.3.2",
-    "font-awesome": "~4.6.3",
-    "moment": "~2.14.1"
   }
 }
--- a/src/wizard/wizard-directive.js
+++ b/src/wizard/wizard-directive.js
@@ -24,15 +24,15 @@
   * @param {string=} cancelTitle The text to display on the cancel button
   * @param {string=} backTitle The text to display on the back button
   * @param {string=} nextTitle The text to display on the next button
-  * @param {function (step)=} backCallback Called to notify when the back button is clicked
-  * @param {function (step)=} nextCallback Called to notify when the next button is clicked
-  * @param {function ()=} onFinish Called to notify when when the wizard is complete.  Returns a boolean value to indicate if the finish operation is complete
-  * @param {function ()=} onCancel Called when the wizard is canceled, returns a boolean value to indicate if cancel is successful
+  * @param {function(step)=} backCallback Called to notify when the back button is clicked
+  * @param {function(step)=} nextCallback Called to notify when the next button is clicked
+  * @param {function()=} onFinish Called to notify when when the wizard is complete.  Returns a boolean value to indicate if the finish operation is complete
+  * @param {function()=} onCancel Called when the wizard is canceled, returns a boolean value to indicate if cancel is successful
   * @param {boolean} wizardReady Value that is set when the wizard is ready
   * @param {boolean=} wizardDone  Value that is set when the wizard is done
   * @param {string} loadingWizardTitle The text displayed when the wizard is loading
   * @param {string=} loadingSecondaryInformation Secondary descriptive information to display when the wizard is loading
-  * @param {number=} wizardHeight The height the wizard should be set to.  If this is not set, the wizard height is adjusted with the window resize event.
+  * @param {string=} contentHeight The height the wizard content should be set to.  This defaults to 300px if the property is not supplied.
   *
   * @example
   <example module="patternfly.wizard" deps="patternfly.form">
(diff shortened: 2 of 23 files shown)


bleathem:
Thanks @dgutride, I'll take a close look.  Probably best if @jeff-phillips-18, @dlabrecq, and @dtaylor113 take a look too.

bower.json
```diff
@@ -39,9 +42,17 @@
     "angular-sanitize": "1.3.0 - 1.5.*",
     "angular-bootstrap": "0.13.x",
     "lodash": "3.x",
-    "patternfly": "~3.11.0"
+    "patternfly": "a014002ebcfad2c6ff2fbeda52dfcd425e016f2a"
```
jeff-phillips-18: So this is WIP? I assume this will require patternfly 3.12?

bower.json
```diff
@@ -39,9 +42,17 @@
...
     "angular-mocks": "1.3.0 - 1.5.*"
+  },
+  "resolutions": {
+    "bootstrap-datepicker": "~1.6.4",
+    "bootstrap-select": "~1.10.0",
+    "bootstrap-touchspin": "~3.1.1",
+    "datatables-colreorder": "~1.3.2",
+    "font-awesome": "~4.6.3",
+    "moment": "~2.14.1"
   }
```
jeff-phillips-18: I assume this all goes away when we update to the correct version of patternfly

src/wizard/wizard-directive.js
```diff
@@ -0,0 +1,677 @@
+/**
+  * @ngdoc directive
+  * @name patternfly.wizard.directive:pfWizard
+  *
+  * @description
+  * Directive for rendering a Wizard modal.
+  *
```
jeff-phillips-18: I think more information is needed here. The overall structure of the wizard directive, step directive, and sub step directives.

src/wizard/wizard-step-directive.js
```diff
@@ -0,0 +1,401 @@
+/**
+ * @ngdoc directive
+ * @name patternfly.wizard.directive:pfWizardStep
+ *
+ * @description
+ * Directive for rendering a Wizard step.  Each step can stand alone or have substeps.
+ *
```
jeff-phillips-18: Indicate that a pfWizardStep must be a child of a pfWizard

test/wizard/wizard.spec.js
```diff
@@ -0,0 +1,209 @@
...
+    // there are two dependent timeouts in the wizard that need to be flushed
+    $timeout.flush();
+    $timeout.flush();
+  });
+
+  it('should dispatch the cancel event on the close button click', function () {
+    var closeButton = element.find('.close');
+    $rootScope.$on('wizard.done', function (event, data) {
+      expect(data).toBe('cancel');
+    });
```
jeff-phillips-18: I don't see that we check that this was called, only that if it was called, the data is correct.

test/wizard/wizard.spec.js
```diff
@@ -0,0 +1,209 @@
...
+  it('should have allowed navigation to review page', function () {
+    var nextButton = element.find('.wizard-pf-next');
+    var nameBox = element.find('#new-name');
+    nameBox.val('test').triggerHandler('input');
+    $scope.$digest();
+
+    $scope.currentStep = 'Review';
+    $scope.$digest();
+    $timeout.flush();
+    $timeout.flush();
```
jeff-phillips-18: Is this all necessary?

src/wizard/wizard-directive.js
```diff
@@ -0,0 +1,677 @@
...
+        if (newVal === true) {
+          $scope.openModal();
+        }
+      });
+
+      if ($window.innerHeight > 400) {
+        $element.height($window.innerHeight - 70);
+      } else {
+        $element.height(400);
+      }
```
jeff-phillips-18: 400 comes from where? It would be nice if this value were settable. Some wizards could be smaller, some could be larger depending upon the contents of the pages.

This value could be binded to in the html rather than using $element.height()

src/wizard/wizard-directive.js
```diff
@@ -0,0 +1,677 @@
...
+          modalInstance.dismiss(reason);
+          wizardDoneListener();
+        };
+
+        modalInstance.result.then(function () { }, function () { });
+
+        wizardDoneListener = $rootScope.$on('wizard.done', closeWizard);
+      };
+    }
+  ]);
```
dtaylor113: I think a good task for the next breakaway sprint would be to see if we could somehow separate out the ngdoc code from the directive code.  When combined it makes for a very large source code file.

test/wizard/wizard.spec.js
dgutride: Yes - there are several layers that need to be rendered  and timeouts that cascade - it doesn't show to the naked eye when rendering but they need to be flushed when in a unit test.

test/wizard/wizard.spec.js
dgutride: The eventFire on line 112 is what causes the wizard.done event to be emitted.  Line 109 is hit later.

bower.json
dgutride: We can't use #master, but I changed this to master-dist for clarity instead of using the commit.

dgutride: Correct - cleaned this up a bit with the last commit.

bower.json
dgutride: Removed it now.

src/wizard/wizard-directive.js
dgutride: More details and an html snippet added to clarify the structure.

src/wizard/wizard-step-directive.js
dgutride: Done

test/wizard/wizard.spec.js
dgutride: You are correct - I removed the .$on portion and added a better emit check lower to make sure the data is correct.

src/wizard/wizard-directive.js
dgutride: Modified this logic to use a minHeight optional wizard property now.  It needs to adjust on window resize so I left it in the link function.

src/wizard/wizard-directive.js
```diff
@@ -0,0 +1,696 @@
...
+</file>
+</example>
+*/
+
+angular.module('patternfly.wizard').directive('pfWizard', function ($window) {
+  'use strict';
+  return {
+    restrict: 'A',
+    transclude: true,
+    replace: true,
```
dtaylor113: directive replace is depreciated.  Do we really need it here?

src/wizard/wizard-directive.js
```diff
@@ -0,0 +1,696 @@
...
+          $scope.prevStepEnabledWatcher();
+        }
+        if ($scope.prevStepTooltipWatcher) {
+          $scope.prevStepTooltipWatcher();
+        }
+
+        // Add watchers for the selected step
+        $scope.nextStepEnabledWatcher = $scope.$watch('selectedStep.nextEnabled', function (value) {
+          $scope.nextEnabled = value;
+          //console.log('next enabled: ' + value);
```
dtaylor113: Just FYI, best practice to use $log, in this case $log.debug.

bower.json
dtaylor113: coolness, thanks

src/wizard/wizard-directive.js
```diff
@@ -3,22 +3,36 @@
...
+  * @param {function (step)=} nextCallback Called to notify when the next button is clicked
+  * @param {function ()=} onFinish Called to notify when when the wizard is complete.  Returns a boolean value to indicate if the finish operation is complete
+  * @param {function ()=} onCancel Called when the wizard is canceled, returns a boolean value to indicate if cancel is successful
   * @param {boolean} wizardReady Value that is set when the wizard is ready
-  * @param {boolean} wizardDone  Value that is set when the wizard is done
+  * @param {boolean=} wizardDone  Value that is set when the wizard is done
   * @param {string} loadingWizardTitle The text displayed when the wizard is loading
-  * @param {string} loadingSecondaryInformation Secondary descriptive information to display when the wizard is loading
+  * @param {string=} loadingSecondaryInformation Secondary descriptive information to display when the wizard is loading
+  * @param {number=} minHeight The minimum height the wizard should adjust to if the window height is decreased.  The wizard height is adjusted with the window resize event and this sets the minimum.
```
jeff-phillips-18: I was thinking more of a height than a minimum height. If all the pages have only a couple of fields, there is no reason to have the wizard be the full height of the window, making the navigation buttons far from the fields. So, maybe a minimum height and a maximum height? or a set height?

dgutride: @jeff-phillips-18  - I did it this way to keep it inline with the design and js from PatternFly which keeps the height at a minHeight or within 70 pixels of the page.  We'd probably have to go back to the designer to adjust it at this point or we'd be inconsistent between the two.

jeff-phillips-18: 😞 

Does the design specifically call this out or is this just how the Patternfly implementation was done?

dgutride: Good question - I can ask that during the review with the stakeholders - there was a concern about the height being variable that was addressed by the last PF checkin.  It was the reason I paid particular attention to the height in the code.

jeff-phillips-18: Agreed about the variable height. The height should not change when navigating through pages, but the height used should be configurable imo. 👍  to getting stakeholders input.

@LHinson @serenamarie125

serenamarie125: @dgutride @jeff-phillips-18 I agree that the height should be configurable.

dgutride:
@jeff-phillips-18  - I think I've addressed all of the comments in this review.  Please let me know if there is anything else to do.

PR #322, opened 2016-10-07
Add ability to set a class on list view rows, use customScope

jeff-phillips-18:
Set a class specified by the row item on the row’s container
(list-group-item).

Adds a customScope  parameter to allow the transcluded html to use
variables or functions defined by the applications scope.

@dtaylor113 @dgutride


dgutride:
This looks good to me.

dtaylor113:
LGTM

Issue #306, opened 2016-08-31
Cannot disable the kebab dropdown in pf-list-view


AparnaKarve:
I have a use case where I need to either hide or disable the kebab dropdown in the PF list view.

I noticed that there is a `ng-class="{'disabled': checkDisabled(item)}"` in the generated html to disable the kebab dropdown as shown below, which does not work.

``` html
<button class="btn btn-link dropdown-toggle" type="button" id="dropdownKebabRight_0" ng-class="{'disabled': checkDisabled(item)}" ng-click="setupActions(item, $event)" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false"><span class="fa fa-ellipsis-v"></span></button>
```

Also, is there an option to hide the kebab dropdown based on a condition?

/cc @jeff-phillips-18

jeff-phillips-18:
@AparnaKarve Please checkout #324 to see if this will meet your needs.

PR #324, opened 2016-10-12
Add ability to set a class or hide menu actions in List View

jeff-phillips-18:
This addresses issue #306 

@dgutride @dtaylor113 @AparnaKarve

### DIFF
--- a/src/views/listview/list-view-directive.js
+++ b/src/views/listview/list-view-directive.js
@@ -43,6 +43,8 @@
  *     <li>.isDisabled - (Boolean) set to true to disable the action
  *     <li>.isSeparator - (Boolean) set to true if this is a placeholder for a separator rather than an action
  *   </ul>
+ * @param {function (item))} hideMenuForItemFn function(item) Used to hide all menu actions for a particular item
+ * @param {function (item))} menuClassForItemFn function(item) Used to specify a class for an item's dropdown kebab
  * @param {function (action, item))} updateMenuActionForItemFn function(action, item) Used to update a menu action based on the current item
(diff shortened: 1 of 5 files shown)


dtaylor113:
Looks good.  Any unit test coverage for these new options?

jeff-phillips-18:
@dtaylor113 Added unit tests for these options

src/views/listview/list-view-directive.js
```diff
@@ -238,6 +242,18 @@
           },
         ];
 
+        $scope.getMenuClass = function (item) {
+          var menuClass = "";
+          if (item.name === "Jim Beam") {
```
dgutride: Maybe a slightly less risque name here?  Not sure if we have public standards for documentation or if I'm just being overly cautious...

jeff-phillips-18:
Heh, that name has been used here since its inception. Do you think I should change it?

src/views/listview/list-view-directive.js
jeff-phillips-18: I've changed the name to protect the innocent 👮

dgutride:
I'm good with this.

AparnaKarve:
Thanks for this PR @jeff-phillips-18 
I haven't tested this yet, but I'm A-OK with @dtaylor113 and @dgutride 's review 👍

AparnaKarve:
Quick question: To be able to test these changes with SSUI, do I have to wait for the next version of angular-patternfly ? Currently SSUI uses `"angular-patternfly": "^3.11.0"`

jeff-phillips-18:
Yes, angular patternfly should release 3.12.0 either today or very shortly and this PR looks to be included.

PR #327, opened 2016-10-12
Fix for list view actions button tooltips

jeff-phillips-18:
Also adds ngDoc example for a start button which changes text upon
click and disables upon click.

@dgutride  @dtaylor113 @AparnaKarve

### DIFF
--- a/src/views/listview/list-view.html
+++ b/src/views/listview/list-view.html
@@ -9,7 +9,7 @@
     <div class="list-view-pf-actions"
          ng-if="(actionButtons && actionButtons.length > 0) || (menuActions && menuActions.length > 0)">
       <button class="btn btn-default {{actionButton.class}}" ng-repeat="actionButton in actionButtons"
-              title="actionButton.title"
+              title="{{actionButton.title}}"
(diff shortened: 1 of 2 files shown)


src/views/listview/list-view.html
```diff
@@ -9,7 +9,7 @@
     <div class="list-view-pf-actions"
          ng-if="(actionButtons && actionButtons.length > 0) || (menuActions && menuActions.length > 0)">
       <button class="btn btn-default {{actionButton.class}}" ng-repeat="actionButton in actionButtons"
-              title="actionButton.title"
+              title="{{}}actionButton.title}}"
```
dgutride: seems to be an extra }} in here

src/views/listview/list-view.html
```diff
@@ -9,7 +9,7 @@
...
               ng-class="{'disabled' : checkDisabled(item) || !enableButtonForItem(actionButton, item)}"
```
dtaylor113: Is this correct handlebars!?

jeff-phillips-18: everyone catches my merge issue but me 😢 

Thanks guys! Fixed.

PR #330, opened 2016-10-24
Add list view expansion to Angular-Patternfly

dgutride:
This change modifies the listview to add expanding row capabilities.  The docs have also been updated to include information and an example related to enabling this.  Rows can now be expanded using an arrow on the far-left that will allow additional information to be displayed.  This behavior can be enabled for the entire list and disabled for individual rows (if needed).
@serenamarie125 @jeff-phillips-18 

Test plan: Added new unit test for the functionality and tested in ng-docs

### DIFF
--- a/misc/examples.css
+++ b/misc/examples.css
@@ -64,6 +64,10 @@ hr {
   padding-left: 5px;
 }
 
+.list-view-container .fa {
+  padding-left: 0px;
+}
+
 .card-view-container {
   border: 1px solid #000000;
   margin: 10px;
--- a/test/views/listview/list-view.spec.js
+++ b/test/views/listview/list-view.spec.js
@@ -528,4 +528,18 @@ describe('Directive:  pfDataList', function () {
     var openItem = element.find('.list-group-item-container');
     expect(openItem.length).toBe(1);
   });
+
+  it('should allow expanding rows to disable individual expansion', function () {
+    $scope.systemModel[0].disableRowExpansion = true;
+    $scope.listConfig.useExpandingRows = true;
+    var htmlTmp = '<div pf-list-view items="systemModel" ' +
+      '  config="listConfig">' +
+      '</div>';
+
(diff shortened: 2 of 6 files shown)


jeff-phillips-18:
Please add more comments to this PR so that release notes can be created with details.

jeff-phillips-18:
Add the list-view-pf-view selector added a 30px top margin inside the list view scrolling container. Why would we want that?

jeff-phillips-18:
The columns are no longer aligned and there is very little space (28px) left for the item name.

jeff-phillips-18:
Adding the expand indicator makes the rows grow in height for no good reason. Looks like the PF class list-view-pf-expand is adding a touch too much margin.

test/views/listview/list-view.spec.js
```diff
@@ -513,4 +513,19 @@ describe('Directive:  pfDataList', function () {
...
+
+    items = element.find('.list-view-pf-expand .fa-angle-right');
+    expect(items.length).toBe(5);
+
+    eventFire(items[0], 'click');
+
+    var openItem = element.find('.list-group-item-container');
+    expect(openItem.length).toBe(1);
+  });
 });
```
jeff-phillips-18: Can we add a test to make sure items that don't expand don't show the expand indicator?

misc/examples.css
```diff
@@ -64,6 +64,10 @@ hr {
   padding-left: 5px;
 }
 
+.list-view-container .fa {
+  padding-left: 0px;
+}
+
```
dgutride: Because the example css overwrite the default span behavior for list-view-container spans just above this.  This fixes that for the specific case I had to add. @jeff-phillips-18

test/views/listview/list-view.spec.js
dgutride: @jeff-phillips-18  - test added, good idea :)

PR #331, opened 2016-10-25
Wrap wizard button in an IIFE to prevent access globally.

dgutride:
ManageIQ identified an issue with this function because the same function had been defined globally in their codebase.  This wraps the function an IIFE which causes the directives to be immediately created and the function is disposed of (and not available on the window object any longer).  

Testing done:
Prior to the fix - wizardButtonDirective() can be called in the console when the wizard code is loaded, afterwards, the directives are created upon load and then the function isn't available globally.

### DIFF
--- a/src/wizard/wizard-buttons.js
+++ b/src/wizard/wizard-buttons.js
@@ -1,6 +1,6 @@
 (function () {
   'use strict';
-  function wizardButtonDirective (action) {
+  function pfWizardButtonDirective (action) {
     angular.module('patternfly.wizard')
       .directive(action, function () {
         return {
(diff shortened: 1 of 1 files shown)


src/wizard/wizard-buttons.js
```diff
@@ -1,30 +1,31 @@
...
-wizardButtonDirective('pfWizPrevious');
-wizardButtonDirective('pfWizFinish');
-wizardButtonDirective('pfWizCancel');
-wizardButtonDirective('pfWizReset');
+  wizardButtonDirective('pfWizNext');
+  wizardButtonDirective('pfWizPrevious');
+  wizardButtonDirective('pfWizFinish');
+  wizardButtonDirective('pfWizCancel');
+  wizardButtonDirective('pfWizReset');
+})();
```
jeff-phillips-18: I would still like to see these function prefixed with pf. For example, pfWizardButtonDirective('pfWizNext')

PR #335, opened 2016-10-30
Update test script in package.json to run tests

huerlisi:
The `test` script in `package.json` was just a dummy. But we provide
proper tests by running `grunt test`.

This patch updates `package.json` to have a proper `test` script.


jeff-phillips-18:
👍

PR #337, opened 2016-11-03
Remove JQuery dependency from default modules

jeff-phillips-18:
## Description

This PR removes the JQuery dependency from the default patternfly module. Two of the existing directives have been moved from the default module to patternfly.jquery: pfSelect and pfDatepicker. To use either of these directives add the patternfly.jquery module to your application module.

An additional directive, pfBootstrapSelect, has been added to achieve the same basic features of pfSelect but without the JQuery dependency.

The angular patternfly code has been adjusted to use the non-JQuery patternfly settings. To keep JQuery independence, include only patternfly-settings.js rather than patternfly.js.

@dtaylor113 @dgutride @dlabrecq @bleathem

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,16 +83,22 @@ Note:
         <script src="bower_components/patternfly/components/c3/c3.min.js"></script>
         <script src="bower_components/patternfly/components/d3/d3.min.js"></script>
 
-5. (optional) The 'patternfly.charts' module is not a dependency in the default angular 'patternfly' module.
-   In order to use patternfly charts you must add 'patternfly.charts' as a dependency in your application:
+        <!-- Patternfly settings-->
+        <script src="bower_components/patternfly/dist/js/patternfly-settings.js"></script>
+
+5. (optional) The 'patternfly.charts' module and 'patternfly.jquery' modules are not dependencies in the default angular 'patternfly' module.
+   In order to use these you must add 'patternfly.charts' and/or 'patternfly.jquery' as dependencies in your application:
 
         my-app.module.js:
 
         angular.module('myApp', [
           'patternfly',
-          'patternfly.charts'
-        ]);
+          'patternfly.charts',
+          'patternfly.jquery'
+]);
 
+    Note: the pfSelect and datepicker directives are only available in the patternfly.jquery module. Using these directives requires the inclusion of the jQuery library.
+    
 ## API documentation
 
 The API documentation can be built with:
--- a/src/bootstrap-select/bootstrap-select.directive.js
+++ b/src/bootstrap-select/bootstrap-select.directive.js
@@ -1,0 +1,89 @@
+/**
+ * @ngdoc directive
+ * @name patternfly.bsselect:pfBootstrapSelect
+ * @element select
+ *
+ * @param {array=} selections - List of possible selections. Value displayed will be the objects' label field or the object itself
+ * @param {object?} currentSelection - The current selected object
+ * @param {string?} updateOnSelect - Flag to update the current selected object on user seleection, optional - default is true
+ * @param {function?} onChange - Optional callback function when user selection is made
+ * @param {string?} placeholder - placeholder text to show when there is no current selection
+ * @param {string?}  dropdownMenuClass - Class to add to the dropdown menu (typically dropdown-menu-right and such)
+ *
+ * @description
+ * An AngularJS wrapper for the {@link  https://angular-ui.github.io/bootstrap/#/dropdown} Angular Bootstrap dropdown directive
+ *
+ * @example
+ <example module="patternfly.bsselect" deps="">
+
+ <file name="index.html">
+ <div ng-controller="SelectDemoCtrl">
+
+ <form class="form-horizontal">
+   <div class="form-group">
+     <label class="col-sm-2 control-label" for="pet">Preferred pet:</label>
+     <div class="col-sm-4">
+       <div pf-bootstrap-select selections="pets" current-selection="pet" placeholder="No pet selected">
+         </div>
+       </div>
+   </div>
+ </form>
+
+ <p>Your preferred pet is {{pet || 'None'}}.</p>
+
+ </div>
+ </file>
+
+ <file name="script.js">
+ angular.module( 'patternfly.bsselect' ).controller( 'SelectDemoCtrl', function( $scope ) {
+       $scope.pets = ['Dog', 'Cat', 'Chicken'];
+       $scope.pet = null;
+     });
+ </file>
+
+ </example>
+ */
+angular.module('patternfly.bsselect').directive('pfBootstrapSelect', ['$timeout', function ($timeout) {
+  'use strict';
+
+  return {
+    restrict: 'A',
+    scope: {
+      selections: '=',
+      currentSelection: '=?',
+      updateOnSelect: '@',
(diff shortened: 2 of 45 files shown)


src/bootstrap-select/bootstrap-select.directive.js
```diff
@@ -0,0 +1,89 @@
...
+     });
+ </file>
+
+ </example>
+ */
+angular.module('patternfly.bsselect').directive('pfBootstrapSelect', ['$timeout', function ($timeout) {
+  'use strict';
+
+  return {
+    restrict: 'A',
```
dtaylor113: Why just an 'A'?  I expected <pf-bootstrap-select..., not having to have to use a <div pf-bootstrap-select=...
Not a big issue though

src/navigation/vertical-navigation-directive.js
```diff
@@ -43,8 +43,8 @@
  * @example
```
dtaylor113: What are all these changes to vert. nav?  Doesn't seem to have anything to do with removing jQuery or pf-select?

src/notification/inline-notification.directive.js
```diff
@@ -39,13 +39,13 @@
          </div>
          <div class="form-group">
            <label class="col-sm-2 control-label" for="type">Type:</label>
-           <div class="col-sm-10">
-            <select pf-select ng-model="type" id="type" ng-options="o as o for o in types"></select>
+           <div class="col-sm-4">
+            <div pf-bootstrap-select current-selection="type" id="type" selections="types"></divselect>
```
dtaylor113: Is that correct closing tag? </divselect>?

dtaylor113:
This is great work Jeff!  I realize this was the result of a breakaway sprint story whose end goal was to remove jQuery dependency, so this PR contains all those changes to accomplish that goal.  However, this PR changes several components.  Ideally, this PR would have been better as several smaller PRs.  I'm going to clone this dev branch and do some manual testing, but so far everything looks good.
Thanks,
- Dave

src/bootstrap-select/bootstrap-select.directive.js
jeff-phillips-18: Typically the Angular Patternfly directives are Attribute only.

src/notification/inline-notification.directive.js
jeff-phillips-18: Nope. I will fix that. Thanks.

jeff-phillips-18:
Thanks @dtaylor113. Agreed which is why I did not squash commits so that the progression was evident and each commit is reviewable on its own.

src/navigation/vertical-navigation-directive.js
jeff-phillips-18: The example was using JQuery to initialize the vertical navigation. Things like:
   $(document).ready(function()
I'm trying to use as little jQuery as possible throughout even in the examples.

dtaylor113:
Think we need to udpate the README file with some of the stuff you talk about in this PR's Description.

dtaylor113:
Hi @jeff-phillips-18, ran through all ngdoc examples.  I'm seeing two issues:

In ngdocs:view

- pfFilter input field doesn't seem to be accepting input, filter not being applied.  Also seen in pfToolbar
- patternfly.jquery.pfDatepicker, throws error: 
```
angular.js:13920 Error: [$injector:modulerr] Failed to instantiate module patternfly.datepicker due to:
Error: [$injector:nomod] Module 'patternfly.datepicker' is not available!
```
Everything else looks good :+1:

dtaylor113:
Probably not due to this PR, but pfValidation:
image
After clicking on the button with no label:
image

jeff-phillips-18:
Fixed the 3 issues noted by @dtaylor113

README.md
```diff
@@ -83,16 +83,22 @@ Note:
...
 
         angular.module('myApp', [
           'patternfly',
-          'patternfly.charts'
-        ]);
+          'patternfly.charts',
+          'patternfly.jquery'
+]);
 
+    Note: the pfSelect and datepicker directives are only available in the patternfly.jquery module. Using these directives requires the inclusion of the jQuery library.
```
dtaylor113: Should 'datepicker' be 'pfDatepicker' or are you also including the straight bootstrap-datepicker as requiring jQuery?

jeff-phillips-18:
I haven't changed the name and would prefer not to to keep as much backwards compatibility as possible.

README.md
dtaylor113: I know base refs README talks about Wiredep.  I just played around with angular-patternfly-demo-app which uses something like wiredep (grunt-usemin?) and it was always including patternfly.js.  Since we talk about just using patternfly-settings.js here, should we mention how to config wiredep to grab the correct file, or is that up to the app dev to know/figure out?

dtaylor113:
LGTM :+1:

README.md
jeff-phillips-18: I would prefer to leave that exercise to the reader. You can setup wiredep to include/exclude how you like.

dtaylor113:
Any movement on this?  Can we get another reviewer? @dgutride , @dlabrecq   -thanks :smile:

README.md
```diff
@@ -83,16 +83,22 @@ Note:
...
+   In order to use these you must add 'patternfly.charts' and/or 'patternfly.jquery' as dependencies in your application:
 
         my-app.module.js:
 
         angular.module('myApp', [
           'patternfly',
-          'patternfly.charts'
-        ]);
+          'patternfly.charts',
+          'patternfly.jquery'
```
bleathem: We seem to have lost the granularity of our modules.  Should we consider keeping the granularity, but adding a `-jquery` suffix to the modules to indicate they bring in the jQuery dependency (with possibly a common `jquery` parent dependency)?  eg. `patternfly-datepicker-jquery` or `patternfly-form-jquery`.

bleathem:
I've reviewed each of the commits - thanks for not squashing them @jeff-phillips-18, it's much easier to review each change in context.  The changes overall seem reasonable to me, I'll leave it up to those more familiar with the angular-patternfly implementation to assess there technical validity (thanks @dtaylor113).

I had one question I asked inline about the module consolidation into the `module.jquery` module.

The only other question I have is to confirm that this PR does not introduce a breaking change for downstream applications. Rather they will have to "opt in" to get the jQuery independence provided by this PR.

Nice work overall @jeff-phillips-18, a very well thought out and comprehensive approach!

README.md
jeff-phillips-18: There are 2 jquery dependent directives. We could add modules for each patternfly.datepicker and patternfly.select then have patternfly.jquery pull both in. I wouldn't be a fan of patternfly-form-jquery with only one directive then patternfly.jquery pull in patternfly-form-jquery. that's 3 modules for one directive.

jeff-phillips-18:
Well, this is the opposite @bleathem. Any downstream applications currently using pfSelect or pfDatepicker will need to "opt in" in order to continue using them (by adding a dependency on the patternfly.jquery module).

To "opt out" of jquery dependence would require a patternfly module that was jquery independent. I couldn't think of a good naming convention for that. Any thoughts or is release noting this good enough?

jeff-phillips-18:
@bleathem ^^

PR #347, opened 2016-11-18
Require the child patternfly module instead of a sibling one

### LINKED ISSUE #340: Webpack puts jQuery twice in the bundle
Hi all. I'm trying to use Webpack to bundle an application with Angular-Patternfly. It seems that jQuery is loaded twice because it looks like the angular-patternfly module requires a form of "peer dependency" for patternfly. Here's the node_modules/angular-patternfly/index.js:
```
require('angular');
require('angular-animate');
require('angular-sanitize');
require('angular-ui-bootstrap');
require('lodash');
require('../patternfly/dist/js/patternfly.min');
require('./dist/angular-patternfly');
module.exports = 'angular-patternfly';
```
And [here](https://github.com/patternfly/angular-patternfly/files/578661/stats.json.txt)'s the webpack graph's JSON file (by the way, in order to make Webpack find the correct patternfly module I had to `npm i patternfly`, which seems kinda weird since angular-patternfly already brings it in as a dependency). You can see the graph uploading that file in the [Webpack's analyze tool](http://webpack.github.io/analyse).

Then, in the node_modules/angular-patternfly/index.js file, I removed the leading `../` from the patternfly entry, so that it would bring in its own dependency, just like the other modules, and indeed it loads jQuery correctly only once (again, [here](https://github.com/patternfly/angular-patternfly/files/578678/stats.json.txt)'s the graph's JSON file for the Webpack analyse tool).

So, is there any reason why Angular-Patternfly doesn't require its own dependency, but rather a peer one?

lorthirk:
Fixes #340 

**ATTENTION: As per @dgutride suggestion, #337 should be merged before merging this PR**

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,6 +3,6 @@ require('angular-animate');
 require('angular-sanitize');
 require('angular-ui-bootstrap');
 require('lodash');
-require('../patternfly/dist/js/patternfly.min');
+require('patternfly/dist/js/patternfly.min');
 require('./dist/angular-patternfly');
 module.exports = 'angular-patternfly';


jeff-phillips-18:
Not sure this should wait for #337. That PR will remove the use of patternfly.js and instead require patternfly-settings. We will have to make sure that index.js is also updated in that PR.

So, 👍  from me.

dgutride:
Ok, let's get it in - thanks @lorthirk

PR #353, opened 2016-12-01
Add `uib` prefixes in accordance to migration guide

jjlangholtz:
Migrate deprecated `angular-ui-bootstrap` usage to new `uib`-prefixed
versions. Most of the changes apply to dropdowns, modals and tooltips.

Also remove classes and aria attributes that are added during link phase
for dropdown directives.

Tested against latest ManageIQ Service UI and was not able to locate any
issues.

https://github.com/angular-ui/bootstrap/wiki/Migration-guide-for-prefixes


dtaylor113:
Hi @jjlangholtz, funny we were just discussing this week how/when to upgrade to latest angular-ui-bootstrap.  Thanks for starting this effort! :+1: 
However, I don't see where you actually updated the ui-bootstrap version in bower.json ?
Thanks,
- Dave

jjlangholtz:
@dtaylor113 I believe the version was already bumped in a recent PR https://github.com/patternfly/angular-patternfly/blob/master/bower.json#L43, I was thinking that it could actually be bumped to the latest version `2.2.0` but I believe that would be a breaking change for clients of `angular-patternfly` and should probably involve a version bump here as well.

dtaylor113:
Hi @jjlangholtz, you are correct, we did just update to angular-bootstrap "0.14.x".  The migration guide says the new prefixes were introduced in 0.14, so it seems like release 0.14 works with and without the 'uib-' prefixes?
Thanks,
- Dave

jjlangholtz:
@dtaylor113 so it looks for the `0.14.x` series both versions are supported, but usage of the non-prefixed versions shows deprecation warnings in the console. For versions `1.x` and above the non-prefixed versions are not supported.

dtaylor113:
Hi @jjlangholtz, thanks for the uib prefix info.

Cloned this branch, unit tests all passed, but saw the following issues when reviewing the ngdoc examples:

- pfCard - Timeframe Filters: The "Last n Days" dropdowns do not open.  Here is a working [example](http://angular-patternfly.rhcloud.com/#/api/patternfly.card.directive:pfCard%20-%20Timeframe%20Filters#example_demo)
- pfCard - Trends: The "Last 15 Days" dropdown doesn't open.

Thanks,
- Dave

PR #354, opened 2016-12-02
Automate the Patternfly releases process via Travis

dlabrecq:
Automate the Patternfly releases process via Travis. This uses a custom tag, which automatically triggers Travis to build, publish, and release. If the Patternfly release is successful, RCUE and Angular Patternfly are built next. If Angular is successful, Patternfly Org is released s well.

Instead of spending between half and a full day releasing Patternfly, Angular PAtternfly, RCUE, and Patternfly Org; the automation manages to get this effort down to about 15-30 minutes.

Should the build fail at any point, it can be fixed and restarted. For example, If Angular Patternfly fails, it can be restarted and Patternfly Org will be released as well. We don't necessarily need to bump the npm version number again or rebuild Patternfy. The npm publish is one of the last steps in the build.

Of course, we still have the ability to run the release manually with the existing scripts I created earlier. The new automation uses these scripts itself.

Spoke to Brian about creating a separate repo for the build and GitHub publish scripts. That way, there is one version of the scripts for all repos. Although, combining scripts would be a next step as each repo has subtle differences in the build.


jeff-phillips-18:
changes look good to me.
Travis build failure.
Please squash commits or add a message to the second commit.

dlabrecq:
Seems the access token may be expired? Updated the AUTH_TOKEN variable via the Travis CI.

Also added a check to ensure the publish script is not executed for pull requests -- no need to step into that part of the code.

Appears to be all set now.

jeff-phillips-18:
👍  LGTM

PR #355, opened 2016-12-04
Refactor unit tests to support breaking changes in UI Router 1.0

lorthirk:
This is to make Angular UI router unit tests pass when using UI Router 1.0 (tested with 1.0.0-beta4). The directive itself was already good, only the unit tests needed a minor refactor.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -46,6 +46,6 @@
   },
   "devDependencies": {
     "angular-mocks": "1.3.0 - 1.5.*",
-    "angular-ui-router": "^0.3.2"
+    "angular-ui-router": "1.0.0-beta.3"
   }
 }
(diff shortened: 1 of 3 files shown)


bower.json
```diff
@@ -46,6 +46,6 @@
   },
   "devDependencies": {
     "angular-mocks": "1.3.0 - 1.5.*",
-    "angular-ui-router": "^0.3.2"
+    "angular-ui-router": "1.0.0-beta.3"
```
jeff-phillips-18: I don't think we want to update the devDependencies to the beta version. Is this necessary?

dtaylor113: idk.  It is a dev dependency, which to me seems like it could be cutting edge since it shouldn't effect production code.  Just my 2 cents :-)

lorthirk: Moreover, I think that 1.0.0-beta should be api-compatible with what will be 1.0.0 release.

PR #361, opened 2016-12-07
Remove JQuery Dependency from Angular Patternfly 4

jeff-phillips-18:
## Description

This PR removes the JQuery dependency from the angular patternfly. Three of the existing directives have been removed:

pfSelect - this can be replace with Angular UI Boostrap's uibDropdown
pfDatepicker - this can be replaced with Angular UI Bootstrap's uibDatepicker
pfDateTimepicker - currently there is no non-jquery support for this component

The angular patternfly code has been adjusted to use the non-JQuery patternfly settings. To keep JQuery independence, include only patternfly-settings.js rather than patternfly.js.

@dtaylor113 @dgutride @dlabrecq @bleathem


dtaylor113:
Hi, I noticed in HeatMap when you hover over a box, you do not see any tooltips.  I thought we had them at some point.  I don't see it working in the A-PF3 ngdocs, so I don't think this PR caused this.

dtaylor113:
Cloned this branch, all unit tests pass, ran through most ngdoc examples -everything looks good!

dtaylor113:
Small nit, but I just noticed in the vertical nav. ngdoc examples, the Help and Username top right dropdowns in the masthead do not open :-(

jeff-phillips-18:
I noticed that as well but since its not really part of vertical nav I didn't take the time to work on that.

PR #369, opened 2016-12-09
Convert card module from directives to components

dgutride:
This completes the entire card module conversion - all unit tests work, ngdocs work correctly and all examples are building and working.

### DIFF
--- a/src/card/basic/card-filter.html
+++ b/src/card/basic/card-filter.html
@@ -1,11 +1,13 @@
-<button type="button" uib-dropdown-toggle class="btn btn-default">
-  {{$ctrl.currentFilter.label}}
-  <span class="caret"></span>
-</button>
-<ul uib-dropdown-menu class="dropdown-menu-right" role="menu">
-  <li ng-repeat="item in $ctrl.filter.filters" ng-class="{'selected': item === $ctrl.currentFilter}">
-    <a role="menuitem" tabindex="-1" ng-click="$ctrl.filterCallBackFn(item)">
-      {{item.label}}
-    </a>
-  </li>
-</ul>
+<div uib-dropdown class="card-pf-time-frame-filter">
+  <button type="button" uib-dropdown-toggle class="btn btn-default">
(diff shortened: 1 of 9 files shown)


dtaylor113:
Why is pfAggregateStatusCard marked as Deprecated?
image

dgutride:
@dtaylor113 - the boolean/alt-layout is marked as deprecated, not the status card - this is not a change from the current behavior

dtaylor113:
LGTM.  Minor issue, not introduced in this PR, but the card examples titles are sort of looking like params?  I don't think the @name likes dashes:

`@name patternfly.card.directive:pfCard - Timeframe Filters`

image

dlabrecq:
LGTM. I'd merge it, but not sure there are any changes need in regard to @dtaylor113 's comment?

dtaylor113:
My issues shouldn't hold up merging this PR :-)

src/card/basic/card-filter.html
```diff
@@ -1,11 +1,13 @@
...
-  <span class="caret"></span>
-</button>
-<ul uib-dropdown-menu class="dropdown-menu-right" role="menu">
-  <li ng-repeat="item in filter.filters" ng-class="{'selected': item === currentFilter}">
-    <a role="menuitem" tabindex="-1" ng-click="filterCallBackFn(item)">
-      {{item.label}}
-    </a>
-  </li>
-</ul>
+<div uib-dropdown class="card-pf-time-frame-filter">
```
dgutride: Yeah - it actually came in from master (it was a contributed change after we branched for 4.0) - I ported it in so I wouldn't lose the change.

@dlabrecq

dgutride:
@dlabrecq @dtaylor113  - I looked into those dashes in the name and we can't change that.  ngdocs uses the @name parameter to generate the left-hand navigation and if we remove those dashes, we'll end up with multiple entries in the navigation with the same name.

dlabrecq:
Ok @dtaylor113, I'll merge.

Issue #368, opened 2016-12-09
Items in an expansion list view that have disabled expansion are ill aligned.


jeff-phillips-18:
If an item in a list view has the disableRowExpansion flag set, the columns for that row get mis-aligned:

![image](https://cloud.githubusercontent.com/assets/11633780/21056420/897d9690-be03-11e6-9af2-232ed4173600.png)

jeff-phillips-18:
Mistake is in application code. Closing.

dtaylor113:
This appears to have been fixed via PR #370

PR #370, opened 2016-12-09
Fix for alignment of list view rows when some items do not expand

jeff-phillips-18:
This fixes issue #368


jeff-phillips-18:
@dgutride @dtaylor113

PR #371, opened 2016-12-09
Convert about modal from directive to component style

dgutride:
All tests pass, ng-docs checked for proper syntax and ng-doc test also works the same as before.


dgutride:
screenshot-modal
doc-1
doc-2

dgutride:
@dtaylor113  - screenshots added

dtaylor113:
@dgutride , thanks for the screenshots.  LGTM.  Although, a little confused on the use of pfAboutModalTransclude.  Is this a behind-the-scenes usage?  I don't see where a user specifies and uses 'pfAboutModalTransclude'?

dgutride:
@dtaylor113  - this happens behind the scenes, a user shouldn't use that - this was ported over, it's not new

src/modals/about-modal.component.js
```diff
@@ -0,0 +1,171 @@
...
+    // When the content is retrieved, it is compiled and linked against the provided scope by the $uibModal service.
+    // Unfortunately, there is no way to provide transclusion there.
+    //
+    // The solution below embeds a placeholder directive (i.e., pfAboutModalTransclude) to append the transcluded DOM.
+    // The transcluded DOM is from a different location than the modal, so it needs to be handed over to the
+    // placeholder directive. Thus, we're passing the actual DOM, not the parsed HTML.
+    ctrl.openModal = function () {
+      //$window.console.log('hi mom');
+      $uibModal.open({
+        component: 'pfModalContent',
```
dlabrecq: Just curious why you chose to create a component here instead of a controller?

dlabrecq:
@dtaylor113 pfAboutModalTransclude is something I introduced when creating the about modal. It's not intended to be used outside the about modal.

From my comments in the code:
"The ui-bootstrap modal only supports either template or templateUrl as a way to specify the content. When the content is retrieved, it is compiled and linked against the provided scope by the $uibModal service. Unfortunately, there is no way to provide transclusion there.

The solution below embeds a placeholder directive (i.e., pfAboutModalTransclude) to append the transcluded DOM. The transcluded DOM is from a different location than the modal, so it needs to be handed over to the placeholder directive. Thus, we're passing the actual DOM, not the parsed HTML."

src/modals/about-modal.component.js
dgutride: @dlabrecq  - this is now supported in uibModal which allows it to render the component directly, it was a recent enhancement in ui-bootstrap

PR #375, opened 2016-12-19
Convert patternfly.form module directives to components.

jeff-phillips-18:


jeff-phillips-18:
@dgutride @dtaylor113 @dlabrecq

dtaylor113:
LGTM.  Whjy are things moved into 'src/form/examples/', directives in 'examples' seem to be core directives (form-buttons, form-group, remain char count.), not examples.

jeff-phillips-18:
Only the ng-doc examples where moved there. I think this makes it easier to read the real component code and not have to look past all the ng-doc stuff.

waldenraines:
I know that you didn't write this originally but what are your thoughts on changing the form components to use normal BS3 form markup instead of `.form-horizontal` markup?

jeff-phillips-18:
Sorry, what is abnormal about form-horizontal?

waldenraines:
> Sorry, what is abnormal about form-horizontal?

It's not that there is anything abnormal about it but I think the default should be the normal form and if we want a horizontal form we should have a different directive or different option for that.

jeff-phillips-18:
Probably a better way to go. Currently we default classes for the label and input components which really force the form-horizontal use. If we don't do that, the application could use either but will have to set the pfLabelClass and pfInputClass to use horizontal.

waldenraines:
> Probably a better way to go. Currently we default classes for the label and input components which really force the form-horizontal use. If we don't do that, the application could use either but will have to set the pfLabelClass and pfInputClass to use horizontal.

Exactly.  This was originally copy/pasted from Katello where we were using horizontal and never updated.

src/form/examples/form-group.js
```diff
@@ -0,0 +1,56 @@
+/**
+ * @ngdoc directive
+ * @name patternfly.form.directive:pfFormGroup
+ *
```
dgutride: restrict E should be in the header

src/form/form-buttons/form-buttons.component.js
```diff
@@ -0,0 +1,34 @@
...
+    form: '^form'
+  },
+  templateUrl: 'form/form-buttons/form-buttons.html',
+  controller: function () {
+    'use strict';
+
+    var ctrl = this;
+
+    if (ctrl.pfWorking === undefined) {
+      ctrl.pfWorking = false;
```
dgutride: Generally any initialization code should be moved to $onInit

src/form/form-group/form-group.component.js
```diff
@@ -0,0 +1,64 @@
...
+    pfInputClass: '@'
+
+  },
+  require: {
+    form: '^form'
+  },
+  transclude: true,
+  templateUrl: 'form/form-group/form-group.html',
+
+  controller: function ($scope, $element) {
```
dgutride: $scope should be removed from the function

jeff-phillips-18:
@dgutride Remaining characters count is not a component. Not sure how we would convert it.

jeff-phillips-18:
@dgutride  Updated per your comments.

src/form/examples/form-group.js
```diff
@@ -0,0 +1,57 @@
...
+           <input id="name" name="name" ng-model="item.name" type="text" required/>
+         </pf-form-group>
+         <pf-form-group pf-label="Description">
+           <textarea id="description" name="description" ng-model="item.description">
+             {{ item.description }}
+           </textarea>
+         </pf-form-group>
+       </form>
+       <p>Horizontal Form</p>
+       <form class="form-horizontal">
```
waldenraines: Should this be updated to reflect the discussed changes to `pf-form-group` (i.e. to remove the horizontal form specific classes)?

src/form/form-group/form-group.html
```diff
@@ -1,9 +1,9 @@
-<div class="form-group" ng-class="{ 'has-error' : hasErrors() }">
-  <label for="{{ pfField }}" class="control-label {{ pfLabelClass }}">{{ pfLabel }}</label>
-  <div class="{{ pfInputClass }}">
+<div class="form-group" ng-class="{ 'has-error' : $ctrl.hasErrors() }">
+  <label for="{{ $ctrl.pfField }}" class="control-label {{ $ctrl.pfLabelClass }}">{{ $ctrl.pfLabel }}</label>
```
waldenraines: Should this be updated to reflect the discussed changes to `pf-form-group` (i.e. to remove the horizontal form specific classes)?

waldenraines:
Ah, missed the boat.  I thought we were going to change the form groups to not be `.form-horizontal` centric?

jeff-phillips-18:
The are no longer .form-horizontal centric.

src/form/form-group/form-group.html
jeff-phillips-18: This is the example code. It shows both normal and horizontal forms:

image

waldenraines:
> The are no longer .form-horizontal centric.

👍

PR #376, opened 2016-12-20
Convert sort, filter, and toolbar modules directives to components

jeff-phillips-18:
This PR also downgrades angular bootstrap version to 2.2.x due to issue https://github.com/angular-ui/bootstrap/issues/6364


src/toolbars/toolbar-component.js
```diff
@@ -0,0 +1,77 @@
...
+
+    function addFilter (field, value) {
+      var newFilter = {
+        id: field.id,
+        title: field.title,
+        value: value
+      };
+      if (!filterExists(newFilter)) {
+        ctrl.config.filterConfig.appliedFilters.push(newFilter);
+
```
dtaylor113: Just noticed here (this is an existing bug) that you don't have:

```
if (newFilter.type === 'select') {
    enforceSingleSelect(newFilter);
}
```
Like you do for pfFilter.   So in the pfToolbar example, the birth month dropdown is accumulative, whereas in the pfFilter it replaces.  Not a biggie :-)

src/filters/filter-component.js
```diff
@@ -0,0 +1,50 @@
...
+    var ctrl = this;
+
+    ctrl.$onInit = function () {
+
+      angular.extend(ctrl,
+        {
+          addFilter: addFilter
+        }
+      );
+    };
```
dtaylor113: I like the use of angular.extend to add functions to ctrl, I'm going to use that in my charts

test/filters/filter.spec.js
```diff
@@ -19,8 +18,6 @@ describe('Directive:  pfFilter', function () {
     var el = $compile(element)(scope);
 
     scope.$digest();
-
-    isoScope = el.isolateScope();
```
dtaylor113: Note: I've had success replace isoScope with the following:

```diff
-    isoScope = el.isolateScope();
+    isoScope = el.controller('pfDonutPctChart');
```
Where 'pfDonutPctChart' is the name of the component controller your after.

jeff-phillips-18: Good to know. Thanks!

src/filters/filter-fields-component.js
```diff
@@ -0,0 +1,81 @@
+/**
+ * @ngdoc directive
+ * @name patternfly.filters.component:pfFilterFields
+ *
```
dgutride: I believe should this have a restrict E here to have the docs show the component correctly in the example code (instead of <ANY ...

src/filters/filter-fields-component.js
```diff
@@ -0,0 +1,81 @@
...
+    ctrl.$onInit = function () {
+      angular.extend(ctrl, {
+        selectField: selectField,
+        selectValue: selectValue,
+        onValueKeyPress: onValueKeyPress
+      });
+    };
+
+    ctrl.$postLink = function () {
+      $scope.$watch('config', function () {
```
dgutride: We'll need to refactor this as this component can't be used within Angular2 - will that be done in a separate PR so we can move this one forward?

src/filters/filter-results-component.js
```diff
@@ -0,0 +1,79 @@
+/**
+ * @ngdoc directive
+ * @name patternfly.filters.component:pfFilterResults
+ *
```
dgutride: @restrict here, too

src/filters/filter-fields-component.js
dtaylor113: Is $postlink not supported in Angular2?

jeff-phillips-18: I have been working on a separate PR to remove the $scope references, its a bit more work and I'd rather not try to address in this PR.

@dtaylor113 $postLink is OK, it's the use of $scope that will be troublesome.

PR #377, opened 2016-12-20
Convert wizard directive

dgutride:
This completes the wizard conversion from directive syntax to component syntax.  No unit tests had to be changed to support this (all pass, linters pass and test pages work as expected).


src/wizard/wizard-step.component.js
```diff
@@ -0,0 +1,325 @@
...
+    nextEnabled: '<?',
+    prevEnabled: '<?',
+    nextTooltip: '<?',
+    prevTooltip: '<?',
+    disabled: '@?wzDisabled',
+    okToNavAway: '<?',
+    allowClickNav: '<?',
+    description: '@',
+    wizardData: '=',
+    onShow: '=?',
```
dtaylor113: So it seems like for our callback functions we are using "=" or "<", should we be looking to using real output bindings "&".  On the other hand, I think if we do use "&" we have to change our API to use $event obj.

src/wizard/wizard-step.component.js
```diff
@@ -0,0 +1,325 @@
...
+    showReview: '@?',
+    showReviewDetails: '@?',
+    reviewTemplate: '@?'
+  },
+  templateUrl: 'wizard/wizard-step.html',
+  controller: function ($timeout) {
+    'use strict';
+
+    var firstRun = true,
+      ctrl = this;
```
dtaylor113: Shouldn't this be in an $onInit?

src/wizard/wizard-step.component.js
```diff
@@ -0,0 +1,325 @@
...
+      }
+    };
+
+    ctrl.stepClick = function (step) {
+      if (step.allowClickNav) {
+        ctrl.goTo(step);
+      }
+    };
+
+    this.addStep = function (step) {
```
dgutride: 1 and 3 are identical, just a remnant of refactoring to not use scope anymore - I tried to clean those up as I went through but there were quite a few of them.  2 is internal and not exposed via the controller.

src/wizard/wizard-step.component.js
```diff
@@ -0,0 +1,325 @@
...
+      angular.forEach(ctrl.getEnabledSteps(), function (step) {
+        if (step.stepTitle === titleToFind) {
+          foundStep = step;
+        }
+      });
+      return foundStep;
+    };
+
+    ctrl.steps = [];
+    ctrl.context = {};
```
dgutride: I'm ok with variable initialization to empty values outside of $oninit if you are - I think if we are doing any complex checks - i.e. see an if() outside of the init, I like to move it, though.

jeff-phillips-18: I'd prefer to see all initialization together rather than scattered between function definitions, etc.

src/wizard/wizard-step.component.js
```diff
@@ -0,0 +1,325 @@
...
+    } else {
+      ctrl.stepPriority = parseInt(ctrl.stepPriority);
+    }
+    if (angular.isUndefined(ctrl.okToNavAway)) {
+      ctrl.okToNavAway = true;
+    }
+    if (angular.isUndefined(ctrl.allowClickNav)) {
+      ctrl.allowClickNav = true;
+    }
+
```
jeff-phillips-18: Seems like all this _should_ be in $onInit

src/wizard/wizard-substep.component.js
```diff
@@ -0,0 +1,100 @@
...
+    } else {
+      ctrl.stepPriority = parseInt(ctrl.stepPriority);
+    }
+    if (angular.isUndefined(ctrl.okToNavAway)) {
+      ctrl.okToNavAway = true;
+    }
+    if (angular.isUndefined(ctrl.allowClickNav)) {
+      ctrl.allowClickNav = true;
+    }
+
```
jeff-phillips-18: This should be done in $onInit

src/wizard/wizard-substep.component.js
```diff
@@ -0,0 +1,100 @@
...
+          enabled = enabled && step.prevEnabled;
+        });
+      }
+      return enabled;
+    };
+    ctrl.$onInit = function () {
+      ctrl.title = ctrl.stepTitle;
+      ctrl.step.addStep(ctrl);
+    };
+
```
jeff-phillips-18: Prefer to see $onInit as the first thing in the controller function, easier to read/understand.

dtaylor113: Just weird that $onInit is called AFTER $onchange, making charts difficult since they rely on c3 to generate the chart in-the-correct-order :-)

src/wizard/wizard-step.component.js
dtaylor113: @dgutride ^^ any thoughts on this?  -thanks

dgutride: We should probably take a second pass after conversion to look at output bindings - that would really change the interface quite a bit and isn't completely necessary right now.  The callback functions won't stop working anytime soon and will allow for easier migration.

src/wizard/wizard-step.component.js
dgutride: @jeff-phillips-18, @dtaylor113  - I'll refactor all initialization first thing tomorrow.

src/wizard/wizard-step.component.js
dgutride: @dtaylor113  - I'll clean this up tomorrow, too - please hold . . .

src/wizard/wizard-step.component.js
dtaylor113: Ok, sounds good to me.  Thanks!

dtaylor113:
LGTM :-)

PR #380, opened 2016-12-21
Convert notification module

dgutride:
Converts notification directives to component style, also includes making sure test and lint works correctly.  

Notification service also converted to not use $rootScope any longer.


src/notification/inline-notification.component.js
```diff
@@ -58,41 +60,44 @@
...
      </div>
    </file>
 
    <file name="script.js">
-     angular.module( 'patternfly.notification' ).controller( 'NotificationDemoCtrl', function( $scope, Notifications ) {
+     angular.module( 'patternfly.notification' ).controller( 'NotificationDemoCtrl', function( $scope, Notifications, $rootScope ) {
        $scope.types = ['success','info','danger', 'warning'];
-       $scope.type = $scope.types[0];
-       $scope.isPersistent = false;
+       $scope.notifications = $rootScope.notifications;
```
jeff-phillips-18: Is this still being used? I don't see where.

jeff-phillips-18:
After making the notification persistent and then hitting the X to remove it, the demo becomes broken.

src/notification/toast-notification.component.js
```diff
@@ -176,69 +176,78 @@
...
+    Object.defineProperty(ctrl, 'showClose', {
+      get: function () {
+        return _showClose;
+      },
+      set: function (value) {
+        _showClose = value;
+        ctrl.updateShowClose();
+      }
+    });
+
```
dgutride: We'd have to change showClose from @ to < to do that, I was trying to leave the interface the same so it can use {{ }}.  I don't mind changing and documenting if you prefer @dtaylor113

dtaylor113: No, I think it preferable to keep the API the same. -thanks!

PR #392, opened 2017-01-25
Added Less css pre-processing

dtaylor113:
Replaced the rather large 'styles/angular-patternfly.css' with 'styles/angular-patternfly.less' (which contains all the @imports) and created several .less files in various component sub-directories.

This is In preparation of contributing the 'Canvas Drag & Drop' component from SUI to angular-patternfly.  This new component does not exist in patternfly base ref. so the accompanying styles need to live in angular-patternfly until such time we decide to contribute it back to base patternfly.   Base patternfly already uses .less, so this makes a future contribution easier.

Also updated the README with setup instructions for pfTableView


src/filters/filters.less
```diff
@@ -0,0 +1,30 @@
...
+    font-size: 12px;
+    font-style: italic;
+    font-weight: 400;
+  }
+}
+.input-group {
+  .input-group-btn {
+    .dropdown-menu>.selected>a {
+      background-color: @color-pf-blue !important;
+      border-color: #0076b7 !important;
```
jeff-phillips-18: This doesn't seem to be from the patternfly color palette, should it be?

src/sort/sort.less
```diff
@@ -0,0 +1,12 @@
+.sort-pf {
+  .btn-link {
+    color: #252525;
```
jeff-phillips-18: This should be a palette color if possible

styles/variables.less
```diff
@@ -0,0 +1,4 @@
+@width100:  100%;
+@width0:    0;
+@height100: 100%;
+@height0:   0;
```
jeff-phillips-18: I'm not sure these are worth having. Honestly, while reviewing I thought @width100 would be 100px so at minimum it is a bit confusing.

Gruntfile.js
```diff
@@ -299,13 +308,13 @@ module.exports = function (grunt) {
...
-      grunt.task.run(['clean', 'lint', 'test', 'ngtemplates', 'concat', 'ngAnnotate', 'uglify:build', 'cssmin', 'copymain', 'ngdocs', 'clean:templates']);
+      grunt.task.run(['clean', 'lint', 'test', 'ngtemplates', 'concat', 'ngAnnotate', 'uglify:build', 'less', 'cssmin', 'copymain', 'ngdocs', 'clean:templates']);
     });
 
     // Runs all the tasks of build with the exception of tests
     grunt.registerTask('deploy', 'Prepares the project for deployment. Does not run unit tests', function () {
       var concatSrc = 'src/**/*.js';
-      grunt.task.run(['clean', 'lint', 'ngtemplates', 'concat', 'ngAnnotate', 'uglify:build', 'cssmin', 'copymain', 'ngdocs', 'clean:templates']);
+      grunt.task.run(['clean', 'lint', 'ngtemplates', 'concat', 'ngAnnotate', 'uglify:build', 'less', 'cssmin', 'copymain', 'ngdocs', 'clean:templates']);
     });
```
dtaylor113: They are, there is a less task under the watch task.

jeff-phillips-18: Ah sorry, I just saw it still watching styles/**/*.css and missed the less section. Might want to remove the styles/**/*.css  from the all section

dtaylor113:
Hi Jeff, I'm sorry, but I don't have time to address these already-existing style issues.  This PR was just a step in the right direction, and to pave the way for the pfCanvas contribution I'm making -which I'm trying to get done by Monday :-)
I created a story [PTNFLY-2138](https://patternfly.atlassian.net/browse/PTNFLY-2138) to further get a-pf in parity with base pf.  I will add a link to your review comments here.

Is that ok?

Thanks,
- Dave

jeff-phillips-18:
@dtaylor113 That is fine. These issues, while they 'existed', were not able to be done correctly prior to using less because the variables are not available in css.

Could we just get rid of the variables.less file and use those values straight up? 
A) I can't imagine 0 or 100 ever not being 0 or 100 so its not like it reduces changes in the future
B) width100 and height100 are unclear.

dtaylor113:
Ok, no prob, I'll remove the varibles.less file :-)

PR #393, opened 2017-01-27
ListView drag and drop

dlabrecq:
## Description
A CloudForms contribution adding drag and drop features to the ListView component for Angular 1. There is a separate PR for Angular 1.5 (i.e., the branch-4.0-dev branch).

Note that Patternfly depends on a feature branch for the CSS. However, this dependency will be replaced automatically by build scripts, upon the next release.

See: https://patternfly.atlassian.net/browse/CFUX-311

## Steps to test or reproduce and link to rawgit

https://dlabrecq.github.io/angular-patternfly.github.io/index.html#/api/patternfly.views.directive:pfListView

listview dnd

### DIFF
--- a/src/views/listview/list-view-directive.js
+++ b/src/views/listview/list-view-directive.js
@@ -15,6 +15,10 @@
  * <li>.showSelectBox          - (boolean) Show item selection boxes for each item, default is true
  * <li>.selectItems            - (boolean) Allow row selection, default is false
  * <li>.dlbClick               - (boolean) Handle double clicking (item remains selected on a double click). Default is false.
+ * <li>.dragEnabled            - (boolean) Enable drag and drop. Default is false.
+ * <li>.dragEnd                - ( function() ) Function to call when the drag operation ended, default is none
+ * <li>.dragMoved              - ( function() ) Function to call when the drag operation moved an element, default is none
+ * <li>.dragStart              - ( function(item) ) Function to call when the drag operation started, default is none
  * <li>.multiSelect            - (boolean) Allow multiple row selections, selectItems must also be set, not applicable when dblClick is true. Default is false
  * <li>.useExpandingRows       - (boolean) Allow row expansion for each list item.
  * <li>.selectionMatchProp     - (string) Property of the items to use for determining matching, default is 'uuid'
@@ -152,6 +156,15 @@
         </form>
       </div>
       <div class="col-md-12">
+        <form role="form">
+          <div class="form-group">
+            <label class="checkbox-inline">
+              <input type="checkbox" ng-model="config.dragEnabled">Drag and Drop</input>
+            </label>
+          </div>
+        </form>
+      </div>
+      <div class="col-md-12">
         <label style="font-weight:normal;vertical-align:center;">Events: </label>
       </div>
       <div class="col-md-12">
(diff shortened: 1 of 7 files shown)


src/views/listview/list-view.html
```diff
@@ -1,52 +1,64 @@
...
     <div class="list-group-item-header">
-      <div class="list-view-pf-expand" ng-if="config.useExpandingRows">
-        <span class="fa fa-angle-right" ng-show="!item.disableRowExpansion" ng-click="toggleItemExpansion(item)" ng-class="{'fa-angle-down': item.isExpanded}"></span>
-        <span class="pf-expand-placeholder" ng-show="item.disableRowExpansion"></span>
-      </div>
-      <div class="list-view-pf-checkbox" ng-if="config.showSelectBox" >
-        <input type="checkbox" value="item.selected" ng-model="item.selected" ng-disabled="checkDisabled(item)" ng-change="checkBoxChange(item)"/>
+      <div class="list-view-pf-dnd-drag-items">
+        <div pf-transclude="parent" class="list-view-pf-main-info"></div>
       </div>
```
jeff-phillips-18: This div should only be added if drag and drop is enabled

src/views/listview/list-view.html
```diff
@@ -1,52 +1,64 @@
...
+      <div class="list-view-pf-dnd-original-items">
```
jeff-phillips-18: only add the class if drag/drop is enabled

src/views/listview/list-view.html
dlabrecq: Yes, I will add this only when drag and drop is enabled.

src/views/listview/list-view.html
```diff
@@ -1,52 +1,64 @@
-<div class="list-group list-view-pf">
+<div class="list-group list-view-pf" dnd-list="items" ng-class="{'list-view-pf-dnd': config.dragEnabled === true}">
+  <div class='dndPlaceholder'></div>
```
dlabrecq: The placeholder div has to be there; otherwise, I get an ugly empty space at the top of the list when enabled.

screen shot 2017-01-27 at 10 04 29 am

src/views/listview/list-view-directive.js
```diff
@@ -618,6 +660,30 @@ angular.module('patternfly.views').directive('pfListView', function ($timeout, $
       scope.checkDisabled = function (item) {
         return scope.config.checkDisabled && scope.config.checkDisabled(item);
       };
+
+      scope.dragEnd = function () {
+        if (scope.config.dragEnd) {
```
dlabrecq: Will do.

dlabrecq:
Updated per Jeffs code review comments.

src/views/listview/list-view.html
jeff-phillips-18: yeah, but if drag/drop is not enabled it should be OK without it right?
(ng-if="config.dragEnabled")

src/views/listview/list-view-directive.js
```diff
@@ -184,6 +197,27 @@
           return $scope.showDisabled && (item.name === "John Smith");
         };
 
+        var dragEnd = function() {
+          $scope.eventText = 'drag end\r\n' + $scope.eventText;
+        };
+        var dragMoved = function() {
```
dtaylor113: Shouldn't this dragMoved functionality be the default behavior in the directive instead of leaving it up to the end users to implement it correctly?

dtaylor113:
LGTM.  I like the way the drag-n-drop works with all the existing selecting mechanisms and row expand content :-)

src/views/listview/list-view.html
jeff-phillips-18: I tried this and I don't quite understand why it behaves this way but you are correct. I don't see why the ng-if makes a difference but I guess we leave it as is.

dlabrecq: I don't quite understand why it behaves like this, but suspect the library is looking for the selector when the page is loaded. If I don't provide this tag, the dnd library will use an HTML "li" element and then you see a list dot in the page when dragging.

dlabrecq: If this is a problem, I can style the HTML "li" element to not show up. Just thought this solution was more elegant than trying to hack styles.

jeff-phillips-18: Yeah, if is fine as is. No harm no foul.

PR #411, opened 2017-02-13
Filter for angular is case sensative and sholud not be

dabeng:
pfFIlter directive of angualar-patternfly should be case-insensitive. @jeff-phillips-18 , could you review this PR?


bleathem:
This change looks reasonable to me.  Thanks for the fix @dabeng

PR #413, opened 2017-02-14
Set line chart data instead of updating it

yaacov:
**Descriptions**
Currently we can only add new points to the line chart.
In ManageIQ we use the same chart object to show different data, so we need to reset the data.

This PR adds the option to delete old points, and replace them with new ons. This does not break the option to just add points, because we always keep the data in a local variable we can use to repopulate the data.

**Screenshot**
screenshot-0 0 0 0 8000-2017-02-14-16-25-38


yaacov:
@jeff-phillips-18 @simon3z @zeari please review

PR #415, opened 2017-02-14
Fix list view drop direction updates

### LINKED ISSUE #399: Last menu in listView drops down instead of up occasionally
Sometimes the last menu for list views will drop down instead of up which causes some actions to disappear. The behavior is shown below from the docs example, but we are also seeing this in the ManageIQ SUI.

list-view-last-dropdown

jjlangholtz:
In #setDropMenuLocation, the bounding rectangle for the menu was being
calculated before it was rendered causing the height to be 0.

For some reason the $timeout was not waiting for the digest cycle to
complete, however, adding a 100 ms delay consistently drops the menu in
the proper direction.

Fixes #399

### DIFF
--- a/src/views/listview/list-view-directive.js
+++ b/src/views/listview/list-view-directive.js
@@ -559,7 +559,7 @@ angular.module('patternfly.views').directive('pfListView', function ($timeout, $
               }
               nextElement = nextElement.parentElement;
             }
-          });
+          }, 100);
         };
       },
 
--- a/src/views/listview/list-view-directive.js
+++ b/src/views/listview/list-view-directive.js
@@ -419,7 +419,7 @@
   </file>
 </example>
  */
-angular.module('patternfly.views').directive('pfListView', function ($timeout, $window, pfUtils) {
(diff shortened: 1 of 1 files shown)


bleathem:
Rather than using a timeout to "hopefully" get the timing correct, is there another event we can be listening for to ensure the rendering is complete before calculating the bounding rectangle?

bleathem:
How about triggering a digest cycle in a requestanimationframe listener?  Might that get us the post-render timing we are looking for?

jjlangholtz:
Thanks for the suggestion @bleathem, just tested it out and it seems to be working great.

bleathem:
Great, glad to hear it worked.  I will point out that the requestanimationframe listener will be invoked outside of the digest loop.  You might want to consider nesting a $scope.apply, or $timeout to get back into it.  If that's not a concern then I'm ok with this approach as it is.

bleathem:
We should update the title of this PR before merging, to make sure the correct description gets into the release notes.

@jjlangholtz can you open a parallel PR ro the branch-4.0-dev branch to update the 4.0 component?

PR #417, opened 2017-02-15
change use of patternfly setting functions from '$()' to 'patternfly'

jeff-phillips-18:
to help with non-jquery applications

@dtaylor113 @dgutride @bleathem

### DIFF
--- a/src/charts/c3/c3-chart-defaults.js
+++ b/src/charts/c3/c3-chart-defaults.js
@@ -1,7 +1,7 @@
 (function () {
   'use strict';
 
-  var patternflyDefaults = $().c3ChartDefaults();
+  var patternflyDefaults = patternfly.c3ChartDefaults();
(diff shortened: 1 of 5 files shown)


bleathem:
Should we consider passing the `patternfly` global as an argument to the IFFE?  As in:
```
(function(patternfly) {
  ...
  ...
  ...
})(patternfly)
```

jeff-phillips-18:
@bleathem  Not all are in IIFE's, I could do it for both that are if you like, or  I could put the other in IIFEs as well. Preference?

bleathem:
I suggest doing it for those that are in IFFEs.  If other IFFEs don't have it, we can address that in future PRs.

jeff-phillips-18:
I will do it for all so as not to forget using 'patternfly' for these in the future.

Issue #401, opened 2017-02-09
Failed to instantiate module patternfly


alexkieling:
Error message:

```
Uncaught Error: [$injector:modulerr] Failed to instantiate module camel due to:
Error: [$injector:modulerr] Failed to instantiate module patternfly due to:
Error: [$injector:modulerr] Failed to instantiate module patternfly.toolbars due to:
Error: [$injector:modulerr] Failed to instantiate module patternfly.views due to:
Error: [$injector:modulerr] Failed to instantiate module dndLists due to:
Error: [$injector:nomod] Module 'dndLists' is not available! You either misspelled the module name or forgot to load it. If registering a module ensure that you specify the dependencies as the second argument.
```

The 'angular-drag-and-drop-lists' dependency is missing from angular-patternfly's bower.json.

dtaylor113:
I believe this was fixed via #422

PR #422, opened 2017-02-17
Added angular-drag-and-drop-lists as a dependency

dlabrecq:
Updated package.json and bower.json so applications get angular-drag-and-drop-lists automatically.


jeff-phillips-18:
Fixes #401

bleathem:
Thanks for raising the issue @jeff-phillips-18, and thanks for the fix @dlabrecq.

PR #423, opened 2017-02-17
Added angular-drag-and-drop-lists as a dependency

dlabrecq:
Updated bower.json so applications get angular-drag-and-drop-lists automatically.


bleathem:
Thanks @dlabrecq

PR #427, opened 2017-03-06
Table View: Hide/Show Columns - AngularJS

dabeng:
This is Hide/Show Columns feature of angular-datatables which behaves like the recommended PatternFly pattern.

This PR will add colvis dropdown menu to pfToolbar and pfTableView - with Toolbar.

@bleathem @jeff-phillips-18 , would you like to review it?

### DIFF
--- a/src/sort/sort.less
+++ b/src/sort/sort.less
@@ -1,4 +1,5 @@
 .sort-pf {
+  display: inline-block;
   .btn-link {
     color: #252525;
     font-size: 16px;
--- a/src/table/tableview/table-view.component.js
+++ b/src/table/tableview/table-view.component.js
@@ -3,6 +3,7 @@ angular.module('patternfly.table').component('pfTableView', {
     config: '<?',
     dtOptions: '<?',
     colummns: '<',
+    columnVisibility: '<',
     items: '<',
     actionButtons: '<?',
     menuActions: '<?',
@@ -11,7 +12,7 @@ angular.module('patternfly.table').component('pfTableView', {
   templateUrl: 'table/tableview/table-view.html',
   controller: function (DTOptionsBuilder, DTColumnDefBuilder, $element, pfUtils, $log, $filter, $timeout) {
     'use strict';
-    var ctrl = this, prevDtOptions, prevItems;
+    var ctrl = this, prevDtOptions, prevItems, prevColumnVisibility;
 
     // Once datatables is out of active development I'll remove log statements
     ctrl.debug = false;
@@ -67,6 +68,7 @@ angular.module('patternfly.table').component('pfTableView', {
       // Need to deep watch changes in dtOptions and items
       prevDtOptions = angular.copy(ctrl.dtOptions);
       prevItems = angular.copy(ctrl.items);
+      prevColumnVisibility = angular.copy(ctrl.columnVisibility);
 
       // Setting bound variables to new variables loses it's one way binding
       //   ctrl.dtOptions = pfUtils.merge(ctrl.defaultDtOptions, ctrl.dtOptions);
@@ -135,6 +137,10 @@ angular.module('patternfly.table').component('pfTableView', {
         //$timeout(function () {
         selectRowsByChecked();
         //});
+      }
+      if (!angular.equals(ctrl.columnVisibility, prevColumnVisibility)) {
+        prevColumnVisibility = angular.copy(ctrl.columnVisibility);
+        ctrl.dtColumnDefs[ctrl.columnVisibility.index].visible = ctrl.columnVisibility.visible;
       }
     };
 
--- a/src/toolbars/examples/toolbar.js
+++ b/src/toolbars/examples/toolbar.js
@@ -50,7 +50,7 @@
   <file name="index.html">
     <div ng-controller="ViewCtrl" class="row example-container">
       <div class="col-md-12">
-        <pf-toolbar id="exampleToolbar" config="toolbarConfig">
+        <pf-toolbar id="exampleToolbar" config="toolbarConfig" on-toggle-column-visibility="toggleColumnVisibility(column)">
          <actions>
            <span class="dropdown primary-action" uib-dropdown>
              <button class="btn btn-default dropdown-toggle" uib-dropdown-toggle type="button">
@@ -119,6 +119,7 @@
       <div class="col-md-12" ng-show="viewType == 'tableView'">
         <pf-table-view config="tableConfig"
                        colummns="colummns"
+                       column-visibility="columnVisibility"
                        items="items"
                        empty-state-config="emptyStateConfig">
         </pf-table-view>
@@ -319,6 +320,7 @@
       var viewSelected = function(viewId) {
         $scope.viewType = viewId;
         $scope.sortConfig.show = ($scope.viewType === "tableView" ? false : true);
+        $scope.columnVisibilityConfig.show = ($scope.viewType === "tableView" ? true : false);
       };
 
       $scope.viewsConfig = {
@@ -392,6 +394,36 @@
         onSortChange: sortChange
       };
 
+      $scope.columnVisibilityConfig = {
+        columns: [
+          {
+            name:  'Name',
+            index: 1,
+            visible: true
+          },
+          {
+            name:  'Age',
+            index: 2,
+            visible: true
+          },
+          {
+            name:  'Address',
+            index: 3,
+            visible: true
+          },
+          {
+            name:  'Birth Month',
+            index: 4,
+            visible: true
+          }
+        ],
+        show: false
+      };
+
+      $scope.toggleColumnVisibility = function (column) {
+        $scope.columnVisibility = { index: column.index, visible: column.visible };
+      };
+
       $scope.actionsText = "";
       var performAction = function (action) {
         $scope.actionsText = action.name + "\n" + $scope.actionsText;
@@ -453,6 +485,7 @@
         viewsConfig: $scope.viewsConfig,
         filterConfig: $scope.filterConfig,
         sortConfig: $scope.sortConfig,
+        columnVisibilityConfig: $scope.columnVisibilityConfig,
         actionsConfig: $scope.actionsConfig
       };
 
--- a/src/toolbars/toolbar-component.js
+++ b/src/toolbars/toolbar-component.js
@@ -1,6 +1,7 @@
 angular.module('patternfly.toolbars').component('pfToolbar', {
   bindings: {
-    config: '='
+    config: '=',
+    onToggleColumnVisibility: '&'
   },
   transclude: {
     'actions': '?'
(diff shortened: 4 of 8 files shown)


src/sort/sort.less
```diff
@@ -1,4 +1,5 @@
 .sort-pf {
+  display: inline-block;
```
jeff-phillips-18: I fear this change. sort-pf is used in many places by many applications in and out of toolbars. Why is this necessary?

src/table/tableview/table-view.component.js
```diff
@@ -3,6 +3,7 @@ angular.module('patternfly.table').component('pfTableView', {
     config: '<?',
     dtOptions: '<?',
     colummns: '<',
+    colvis: '<',
```
jeff-phillips-18: colvis is not intuitive as to what it is used for. Please spell it out as column-visibility

src/table/tableview/table-view.component.js
```diff
@@ -136,6 +138,13 @@ angular.module('patternfly.table').component('pfTableView', {
         selectRowsByChecked();
         //});
       }
+      if (!angular.equals(ctrl.colvis, prevColvis)) {
+        if (ctrl.debug) {
+          $log.debug("  colvis !== prevColvis");
+        }
```
jeff-phillips-18: The debugging should be removed.

src/toolbars/toolbar-component.js
```diff
@@ -1,6 +1,7 @@
 angular.module('patternfly.toolbars').component('pfToolbar', {
   bindings: {
-    config: '='
+    config: '=',
+    onToggleColvis: '&'
```
jeff-phillips-18: colvis is not intuitive as to what it is used for. Please spell it out as onToggleColumnVisibility

src/table/tableview/table-view.component.js
```diff
@@ -67,6 +68,7 @@ angular.module('patternfly.table').component('pfTableView', {
       // Need to deep watch changes in dtOptions and items
```
jeff-phillips-18: Comment update

src/table/tableview/table-view.component.js
jeff-phillips-18: I don't think I agree with using colvis to update the visibility of columns. For one thing, how would I set the table up initially? How would I set 2 or more columns to be not visible?

src/table/tableview/table-view.component.js
```diff
@@ -136,6 +138,13 @@ angular.module('patternfly.table').component('pfTableView', {
         selectRowsByChecked();
         //});
       }
+      if (!angular.equals(ctrl.colvis, prevColvis)) {
+        if (ctrl.debug) {
+          $log.debug("  colvis !== prevColvis");
+        }
+        prevColvis = angular.copy(ctrl.colvis);
+        ctrl.dtColumnDefs[ctrl.colvis.index].visible = ctrl.colvis.visible;
```
jeff-phillips-18: Need to check if the column defs actually contain this field.

src/table/tableview/table-view.component.js
jeff-phillips-18: onToggleColvis should be in the config object to be consistent with other callback functions.

src/toolbars/toolbars.less
```diff
@@ -29,3 +29,22 @@
 .toolbar-pf-actions.no-filter-results {
   margin-bottom: 10px;
 }
+.dropdown-colvis-pf {
+  margin-left: 15px;
+  .dropdown-menu {
+    > li {
+      padding: 5px 10px;
+      &:hover {
+        background-color: #def3ff;
```
jeff-phillips-18: This should be a color palette variable.

src/toolbars/toolbars.less
```diff
@@ -29,3 +29,22 @@
...
+      }
+      > input {
+        margin-top: 0;
+        margin-right: 10px;
```
jeff-phillips-18: alphabetical order please.

src/toolbars/toolbars.less
```diff
@@ -29,3 +29,22 @@
...
+        vertical-align: middle;
+      }
+      > label {
+        margin-bottom: 0;
+      }
+    }
+  }
+}
```
jeff-phillips-18: Please add a new line.

jeff-phillips-18:
Overall, I don't really agree with using the first item in the list of items to determine the columns in the table view component. Seems to me that dtColumnDefs should be a binding to allow the application to set it up how they want. What happens when the items are empty when the table is initialized and then items are filled in later? While this is not really part of this PR, trying to set the initial column visibilities may cause you to rethink this.

jeff-phillips-18:
Need ng-doc updates to include new bindings for both toolbar and table component.

jeff-phillips-18:
Need updates to table component unit test.

jeff-phillips-18:
Also, while we are still in alpha, it might be a good time to put the table view component with list view and card view components in src/views.

src/sort/sort.less
dabeng: Without this setting, Sort control group will look like the following screenshot.
screen shot 2017-03-07 at 11 00 47 am

src/table/tableview/table-view.component.js
dabeng: (1) colvis is a terminology of DataTables plugin. From my point of view, we should respect the original ideas of authorship and usage habit of veteran datatables users. FYI, https://datatables.net/reference/button/colvis.

dabeng: 2.
(1)  The task "hide some columns initially" has nothing to do with colvis feature included in toolbar. No matter toolbar provides colvis or not, if users want to hide columns on initial timing, pfTableview should supply the feasible solution to handle it.

(2) However, current code use the hybrid mode to render datatables -- pre-populated html dom table drived by angular + core api of jquery datatables. Such an approach directly causes the lightweight solution as shown below not working.
```js
      $scope.dtOptions = {
        paginationType: 'full',
        displayLength: 10,
        dom: "tp",
        columnDefs: [                      //
          { targets: [3], visible: false },//
        ]                                  //
      };
```
(3) FYI, the following is just a temporary solution.
```js
      $scope.dtOptions = {
        paginationType: 'full',
        displayLength: 10,
        dom: "tp",
        initComplete: function( settings, json ) {//
          this.api().column(3).visible(false);    //
        }                                         //
      };
```

src/table/tableview/table-view.component.js
dabeng: There is no onToggleColvis in table-view.component.js file. It's in toolbar-component.js file. Additionally, I just followed the inputs/outputs styles described in [angular guide](https://docs.angularjs.org/guide/component#component-based-application-architecture). Is it necessary to customize the callback mechanism by ourselves ?

src/table/tableview/table-view.component.js
dabeng: got it

src/table/tableview/table-view.component.js
dabeng: What's the "this field"?

dabeng:
With respect, jquery datatables, angular-datatables and pfTableView are in three different levels of abstraction. But it seems that our current code do a lot of work which jquery datatables or angular-datatables should take on. This is exactly the matter that I worry most because this situation will probably become the first big hindrance we add new features or fix bugs for pfTableView in the future.

dtaylor113:
> Also, while we are still in alpha, it might be a good time to put the table view component with list view and card view components in src/views.

Hi Jeff, 'table view' requires dtables.net and jQuery so I intentionally set it in it's own module so that users could have list and card view without having to also pull in dtables/jQuery.

src/table/tableview/table-view.component.js
dtaylor113: > (2) However, current code use the hybrid mode to render datatables -- pre-populated html dom table drived by angular + core api of jquery datatables. Such an approach directly causes the lightweight solution as shown below not working.

Why doesn't this work?  dtOptions are passed directly to angular-datatables, unchanged by our code.  If angular-datatables supports dtOptions as you have outlined, then it should work.

dtaylor113: http://l-lin.github.io/angular-datatables/archives/#!/withColVis

src/toolbars/examples/toolbar.js
```diff
@@ -50,7 +50,7 @@
   <file name="index.html">
     <div ng-controller="ViewCtrl" class="row example-container">
       <div class="col-md-12">
-        <pf-toolbar id="exampleToolbar" config="toolbarConfig">
+        <pf-toolbar id="exampleToolbar" config="toolbarConfig" on-toggle-colvis="toggleColvis(column)">
          <actions>
```
dtaylor113: Up until now pfToolbar has been independent of any 'view' (list, card).  It seems wrong to add a table-view specific attribute here.  Especially,  since (and I agree with @jeff-phillips-18) it's name is very cryptic.   Could we rename it to make it clearer?  How about 'toggleTableColsVisibility'.  I know 'colvis' is the name the plug-in uses, but it has been depreciated, and soon will be forgotten.

src/toolbars/examples/toolbar.js
```diff
@@ -119,6 +119,7 @@
       <div class="col-md-12" ng-show="viewType == 'tableView'">
         <pf-table-view config="tableConfig"
                        colummns="colummns"
+                       colvis="colvis"
                        items="items"
```
dtaylor113: Couldn't we add the properties of 'colvis' to the already in use 'columns' object? Seems logical to add 'index' and 'visible' to columns obj. which already has other column attributes.

src/table/tableview/table-view.component.js
dabeng: Because of the following sentence:
```js
 <td ng-repeat="(key, value) in item" ng-if="$ctrl.isColItemFld(key)">{{ value }}</td>
```

src/toolbars/examples/toolbar.js
dabeng: First, $scope.colummns is special for defining table head. Second, "colvis" is just an intermediate variable which is in charge of telling pfTableView which column's visibility is toggled. Any time, there is only one colvis variable.

src/toolbars/examples/toolbar.js
dabeng: Hi @dtaylor113 , I just followed the inputs/outputs styles described in [angular guide](https://docs.angularjs.org/guide/component#component-based-application-architecture). Here, we expose the "column" variable to its parent controller scope in order to communicate with sibling component pfTableView.

bleathem:
+1 for using `column-visibility` over `colvis`.  I imagine the majority of our users will come to using the table view without having previously used datatables.net, and so providing a more descriptive name will make it more clear how to consume the API.  We've also decided not to use the `buttons` library, as it doesn't allow for sufficient customisation to match PatternFly patterns.

jeff-phillips-18:
@dabeng Sorry, I am consumed with other work at the moment. I will get back to this when I free up a bit.

dabeng:
Thanks a lot 😊

bleathem:
Apparently I closed this when I deleted the 4.0-dev branch.

@dabeng can you please rebase this PR against master (master is now 4.0) and reopen it?
@jeff-phillips-18 are we prepared to resume our review of this PR?

PR #434, opened 2017-03-10
Fixes #433 listview actionButton class not applied properly

### LINKED ISSUE #433: listview config for actionButton class not displaying in some cases
When trying to pass a different class for a listview action button like `btn-danger` the default class of `btn-default` takes precedent and renders as such. `btn-primary` however seems to override `btn-default` so this seems to have gone unnoticed. I will submit a PR that fixes this by doing `btn {{actionButton.class = actionButton.class || 'btn-default'}}` in the template. Tests for the fix are included as well.

derekmwright:


src/views/listview/list-view.html
```diff
@@ -24,7 +24,7 @@
 
         <div class="list-view-pf-actions"
              ng-if="(actionButtons && actionButtons.length > 0) || (menuActions && menuActions.length > 0)">
-          <button class="btn btn-default {{actionButton.class}}" ng-repeat="actionButton in actionButtons"
+          <button class="btn {{actionButton.class = actionButton.class || 'btn-default'}}" ng-repeat="actionButton in actionButtons"
```
jeff-phillips-18: I don't think we should assign actionButton.class here. Just using {{actionButton.class || 'btn-default'}} should be sufficient.

jeff-phillips-18:
A bit concerned about removing btn-default when adding other classes. I understand it is necessary when adding certain classes, but any current uses that rely on btn-default but are still adding other classes would break.

derekmwright:
Yea I had that in mind when I authored the change but was thinking this config accepted an array, going back and looking and it appears to not be the case. Perhaps a better update would be to convert class to be an array instead of string and then join them together. Then a default assignment would be `btn-default` otherwise assign the joined array. This would allow you to pass `btn-default` as well as additional classes. Let me know what you think and I can revise this.

jeff-phillips-18:
Currently you can pass "class1 class2 class3" and all will be added to the button. So the use of an array is not necessary to achieve this. However, either way, btn-default would need to be included if it was necessary in either case. 

This could be in in version 4 (still in alpha) with release notes with less concern for backwards compatibility. Would that make more sense?

derekmwright:
Yea I think that makes sense. Updating my PR to remove the assignment and just have the conditional return.

dtaylor113:
@jeff-phillips-18 are all your concerned addressed?  -thanks

jeff-phillips-18:
I am still concerned about removing the btn-default class being removed. This is a backwards compatibility issue. I would rather see this go into version 4 before it releases and make sure it is part of the upgrade notes.

dtaylor113:
Moved this PR to v4.0 branch, new PR #455

PR #451, opened 2017-04-18
Fixes #385 - filter input causes a toolbar view change.

### LINKED ISSUE #385: pfToolbar: current view not sticking after hitting enter on filter value
Reproducible at:

https://www.patternfly.org/angular-patternfly/#/api/patternfly.toolbars.directive:pfToolbar#example_demo

1. In toolbar, click on 'card view' icon on far right
2. In Name filter input field enter 'Frank' and hit [Enter]
3. Items are filtered to 'Frank', but view switches back to List view

I've debugged into this, cannot figure out why after hitting [Enter] key:

`Line 451 in src/toolbars/toolbar-directive.js "$scope.viewSelected = function (viewId)" `

...is called with the 'List View', even though the last view was 'Card View'

Note:
- Adding a filter via dropdown (Birth Month) doesn't result in switching back to List View; Card View remains current
- This also occurs in the a-pf 1.5 toolbar/filter components.

jglogan:
Entering a value into the pf-toolbar filter input box causes the toolbar viewConfig.viewSelected() handler to be invoked.  

This causes the view to change if a filter is keyed in while in any other view.

This appears to happen because the return keypress is also interpreted as a click event, which triggers the ng-click handler for the view selection button.

### DIFF
--- a/src/filters/filter-fields-directive.js
+++ b/src/filters/filter-fields-directive.js
@@ -62,6 +62,7 @@ angular.module('patternfly.filters').directive('pfFilterFields', function () {
 
       scope.onValueKeyPress = function (keyEvent) {
         if (keyEvent.which === 13) {
+          keyEvent.preventDefault();
           scope.addFilterFn(scope.currentField, scope.config.currentValue);
           scope.config.currentValue = undefined;
         }


dtaylor113:
Thanks @jglogan!  Appreciate the fix/PR :-)

PR #457, opened 2017-05-01
Add ability to hide sidebar and/or header in pfWizard

jeff-phillips-18:
This PR also:
    
-     Allows the prevCallback not to be set.
-     Adds disabled class to step indicators when the user cannot nav directly to the step
-     Allows the wizard to be included in typescript applications:
         The use of require ^ fails when being included in typescript so that
         has been removed and replaced by using scope to find the required

@cdcabrera

### DIFF
--- a/src/wizard/wizard-buttons.js
+++ b/src/wizard/wizard-buttons.js
@@ -5,17 +5,32 @@
       .directive(action, function () {
         return {
           restrict: 'A',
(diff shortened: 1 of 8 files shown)


jeff-phillips-18:
@jwforres

jeff-phillips-18:
Closing in favor of a new PR with some additional changes.

PR #458, opened 2017-05-02
Wizard Enhancements

jeff-phillips-18:
- allow hiding header
- allow hiding the sidebar navigation panel
- allow hiding the back button (useful in 2 page wizards)
- add disabled class to step indicators when disabled
- allow adding a class to the sidebar and step panels rather than setting a height
- allow use of the wizard from typescript bases applications:
  The use of require ^ fails when being included in typescript so that has been removed and replaced by using scope to find the required Wizard/WizardStep controllers.

@jwforres @spadgett @rhamilto

### DIFF
--- a/src/wizard/wizard-directive.js
+++ b/src/wizard/wizard-directive.js
@@ -20,6 +20,12 @@
   *
   * @param {string} title The wizard title displayed in the header
   * @param {boolean=} hideIndicators  Hides the step indicators in the header of the wizard
+  * @param {boolean=} hideSidebar  Hides page navigation sidebar on the wizard pages
+  * @param {boolean=} hideHeader Optional value to hide the title bar. Default is false.
+  * @param {boolean=} hideBackButton Optional value to hide the back button, useful in 2 step wizards. Default is false.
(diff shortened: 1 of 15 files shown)


src/wizard/wizard-directive.js
```diff
@@ -20,6 +20,11 @@
   *
   * @param {string} title The wizard title displayed in the header
   * @param {boolean=} hideIndicators  Hides the step indicators in the header of the wizard
+  * @param {boolean=} hideSidebar  Hides page navigation sidebar on the wizard pages
+  * @param {boolean=} hideHeader Optional value to hide the title bar. Default is false.
+  * @param {string=} stepClass Optional CSS class to be given to the steps page container. Used for the sidebar panel as well unless a sidebarClass is provided.
+  * @param {string=} aidebarClass Optional CSS class to be give to the aidebar panel. Only used if the stepClass is also provided.
```
jwforres: noticed typo in two places aidebar -> sidebar

jeff-phillips-18: Thanks. Fixed.

jeff-phillips-18:
Thanks @cdcabrera, I have updated the review page as well.

PR #460, opened 2017-05-08
Wizard Enhancements (4.x)

jeff-phillips-18:
- allow hiding header
- allow hiding the sidebar navigation panel
- allow hiding the back button (useful in 2 page wizards)
- add disabled class to step indicators when disabled
- allow adding a class to the sidebar and step panels rather than setting a height
- allow use of the wizard from typescript bases applications:   
  The use of require ^ fails when being included in typescript so that has been removed and replaced by using scope to find the required Wizard/WizardStep controllers.

This is a port of #458

### DIFF
--- a/src/wizard/wizard.component.js
+++ b/src/wizard/wizard.component.js
@@ -21,6 +21,13 @@
   *
   * @param {string} title The wizard title displayed in the header
   * @param {boolean=} hideIndicators  Hides the step indicators in the header of the wizard
+  * @param {boolean=} hideSidebar  Hides page navigation sidebar on the wizard pages
+  * @param {boolean=} hideHeader Optional value to hide the title bar. Default is false.
+  * @param {boolean=} hideBackButton Optional value to hide the back button, useful in 2 step wizards. Default is false.
+  * @param {string=} stepClass Optional CSS class to be given to the steps page container. Used for the sidebar panel as well unless a sidebarClass is provided.
+  * @param {string=} sidebarClass Optional CSS class to be give to the sidebar panel. Only used if the stepClass is also provided.
+  * @param {string=} contentHeight The height the wizard content should be set to. This is used ONLY if the stepClass is not given. This defaults to 300px if the property is not supplied.
+  * @param {boolean=} hideIndicators  Hides the step indicators in the header of the wizard
   * @param {string=} currentStep The current step can be changed externally - this is the title of the step to switch the wizard to
   * @param {string=} cancelTitle The text to display on the cancel button
   * @param {string=} backTitle The text to display on the back button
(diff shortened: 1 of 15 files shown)


test/wizard/wizard.spec.js
```diff
@@ -92,18 +97,21 @@ describe('Component:  pfWizard', function () {
...
+    var modalHtml = $templateCache.get(wizardHtml);
     element = compileHtml(modalHtml, $scope);
     $scope.$digest();
 
     // there are two dependent timeouts in the wizard that need to be flushed
-    $timeout.flush();
-    $timeout.flush();
-  });
+//    $timeout.flush();
+//    $timeout.flush();
```
dtaylor113: Do we need these anymore?

src/wizard/wizard.component.js
```diff
@@ -305,7 +312,12 @@ angular.module('patternfly.wizard').component('pfWizard', {
   bindings: {
     title: '@',
     hideIndicators: '=?',
-    currentStep: '<?',
+    hideSidebar: '@',
+    hideHeader: '@',
+    hideBackButton: '@',
+    sidebarClass: '@',
+    stepClass: '@',
+    contentHeight: '=?',    currentStep: '<?',
```
dtaylor113: is `currentStep: '<?',` really on the same line as `contentHeight`?

test/wizard/wizard.spec.js
```diff
@@ -92,18 +97,17 @@ describe('Component:  pfWizard', function () {
...
-  beforeEach(function () {
+  var setupWizard = function(wizardHtml) {
     setupWizardScope();
-    var modalHtml = $templateCache.get('test/wizard/wizard-container.html');
+
+    var modalHtml = $templateCache.get(wizardHtml);
     element = compileHtml(modalHtml, $scope);
     $scope.$digest();
-
-    // there are two dependent timeouts in the wizard that need to be flushed
```
cdcabrera: The comment used here relates a nice bit of info about the need for running ```concurrent $timeout.flush()``` Should we look at keeping at least one of them?

dtaylor113: Sure, as long as there is a reason to :-)

cdcabrera: Indeed, could be labeled as... simply related to overall testing against multiple dependent components.

jeff-phillips-18: They actually should be flushed here, rather that in each test only when critical for test passage. I've put them back with the comment. Also added a comment where we flush again when navigating to a specific page.

cdcabrera: Sure thing

PR #461, opened 2017-05-08
Fix to allow previous from substeps when there is no backCallback

jeff-phillips-18:
(already fixed in 4.x)


jeff-phillips-18:
@cdcabrera It ought to be, I will update.

jeff-phillips-18:
@cdcabrera added the same optimization to wizard directive and next function of wizard step directive. Please re-review. Thanks!

bleathem:
Does this PR need to be merged for the 3.24.0 release?

jeff-phillips-18:
@bleathem yes

PR #462, opened 2017-05-09
Removed the jquery resolution from bower.json

bleathem:
This resolution is no longer needed, as patternfly has downgraded the dependency (bootstrap-switch) to a version that no longer brings in this old version of jquery.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -48,8 +48,5 @@
   "devDependencies": {
     "angular-mocks": "1.3.0 - 1.5.*",
     "angular-ui-router": "1.0.0-beta.3"
-  },
-  "resolutions": {
-    "jquery": "~2.1.4"
   }
 }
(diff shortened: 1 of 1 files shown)


bleathem:
Travis is failing with the message:
> bower ECONFLICT     Unable to find suitable version for jquery

I tested this locally by using the master branch for patternfly as my dependency, and it worked well.  I can time this tomorrow to merge this PR after the patternfly release, but before the angular-patternfly release.  Or we can merge it now with the broken bower install.

dtaylor113:
I think there are some dependencies in the angular-patternfly repo which use jQuery (datatables, bootstrap, bootstrap-datepicker, etc..) which may require the jQuery resolution.

bleathem:
@dtaylor113  This PR undoes the temporary fix I introduced with https://github.com/patternfly/angular-patternfly/commit/c12385e78538019cbb76ad402936d2aaecf351ab.  It is no longer necessary as PatternFly 3.24.0 has fixed the underlying problem.

PR #459, opened 2017-05-08
WIP, NPM & ReadMe Updates

cdcabrera:
Work in Progress

Description
NPM scripts & Read Me updates.

Proposed changes
- NPM scripts pointed at Grunt tasks. Shortcuts/aliases for forward planning.
  - Original Grunt tasks can still be accessed
- Read Me updates related to NPM scripts.
- Read Me updates related to initial unit testing guidelines

@dtaylor113 @jeff-phillips-18 input needed on how much unit testing Read Me updates should encompass... high level vs detailed. Plus if there's anything specific that needs to be called out.

PR #464, opened 2017-05-16
[WIP] Build Package Updated 4x

cdcabrera:
Extension of PR #459 NPM & ReadMe Updates

Changes:
- ReadMe
  - NPM script updates
  - Unit testing 
- NPM
  - Script aliases for Grunt tasks
  - NPM Ignore for cleaner install
- Bower 
  - JSON ignore, for cleaner install
- Grunt
  - Less copy as part of the ```dist```/distribution


bower.json
```diff
@@ -11,22 +11,23 @@
...
-    "grunt-ngdocs-index.tmpl"
+    "/styles",
+    "test",
+    "deploy_key.enc",
+    "eslint.yaml",
+    "grunt-ngdocs-index.tmpl",
+    "Gruntfile.js",
+    "/index.js",
+    "npm-shrinkwrap.json",
+    "package.json",
```
jeff-phillips-18: We should include package.json in the distribution

package.json
```diff
@@ -57,13 +58,24 @@
...
+    "watch":"grunt watch",
+    "ngdocs": "grunt ngdocs",
+    "ngdocs:view": "grunt ngdocs:view",
+    "ngdocs:publish": "grunt ngdocs:publish",
+    "lint": "grunt lint",
+    "test": "grunt test",
+    "check": "grunt check",
+    "help": "grunt help",
+    "server": "grunt server",
+    "start": "grunt server"
```
jeff-phillips-18: This may be a good time to update from 'grunt server' to 'grunt serve' to be more consistent with grunt guidelines.

cdcabrera: taking a look

bower.json
cdcabrera: sure thing, as a heads up this ignore was in there previously

Gruntfile.js
```diff
@@ -333,7 +339,7 @@ module.exports = function (grunt) {
     grunt.registerTask('test', ['karma']);
     grunt.registerTask('check', ['lint', 'test']);
     grunt.registerTask('help', ['availabletasks']);
-    grunt.registerTask('server', ['ngdocs:view']);
+    grunt.registerTask('serve', ['ngdocs:view']);
```
jeff-phillips-18: I don't see the change from server to serve in Gruntfile.js

cdcabrera:
Remaining work, less files need to be properly aggregated into ```dist```

cdcabrera:
@jeff-phillips-18 @dtaylor113 need input on the last piece of the build updates. It's related to how we should be handling [line 1 in ```angular-patternfly.less```](https://github.com/cdcabrera/angular-patternfly/blob/d683169d9c02cbc36ea7e7de3b88fb513b93b4bb/styles/angular-patternfly.less#L1) once it gets distributed...
```css
@import "../node_modules/patternfly/dist/less/color-variables.less";
```

Up for alternatives, especially if there's something like Webpack that may benefit from another layout. A couple of proposals... 

## Either 

We could handle it similar to how base Patternfly handles it by including a ```dist/less/dependencies/patternfly/less/[single related less file here]``` directory and updating ```angular-patterfly.less``` to appear as:
```css
@import "patternfly/less/color-variables.less";
```
Or being a little bit nicer by updating the path for immediate consumption
```css
@import "dependencies/patternfly/less/color-variables.less";
```
Or simply flatten the structure down to 
```css
@import "color-variables.less";
```
- An advantage, immediately able to consume the less 
- A disadvantage, if we ever needed to clean up again we may be unintentionally creating a downstream dependency
- A disadvantage, if we try to include all less dependencies, instead of the single ```color-variables.less```, from Patternfly we may end up with a larger directory than intended

## Or

We could we could simply leave the reference as is, pointing at ```node_modules```:
```css
@import "../node_modules/patternfly/dist/less/color-variables.less";
```
- A disadvantage, this ignores Bower consumption

bleathem:
We should consider using a less import syntax that does not require the consumer to specify complex less build paths when they choose to customise the angular-patternfly CSS.

cdcabrera:
Readme updates for consuming Less to be added.

cdcabrera:
see PR #475

bleathem:
Apparently I closed this when I delete the b.0-dev branch.  I assume it's been superseded by #475, otherwise feel free to reopen it.

PR #470, opened 2017-05-25
Set the patternfly version to 3.25.0

bleathem:
This PR is noteworthy as it brings in jQuery 3.


bleathem:
I had to update some broken jQuery API usage in the tests.  All tests now pass.  A spot check of the API doc reveals components are working as expected.

PR #479, opened 2017-06-02
Update mentions of customScope to specify $ctrl.customScope

### LINKED ISSUE #478: pf-list-view custom-scope attributute no longer passing objects
So [here's the implementation](https://github.com/manageiq/manageiq-ui-service/blob/773c7c0738eaaa96b6308e8ee0dc595fb14d81a1/client/app/services/service-details/service-details.html#L172)

And [here's the pr where we're making the pf4 leap](https://github.com/ManageIQ/manageiq-ui-service/pull/804)

`custom-scope` previously passed the contents of `vm` for ang-pf 3.x but now it doesn't... [looking at the file](https://github.com/patternfly/angular-patternfly/blob/master/src/views/listview/list-view.component.js#L455), I can't for the life of me see anything terribly offensive...

TL;DR  3.x -> 4.x change from attribute to element, all things should work, did that, something doesn't work, thought it might be worth a collective 👀

AllenBW:
closes #478

### DIFF
--- a/src/notification/notification-drawer.component.js
+++ b/src/notification/notification-drawer.component.js
@@ -26,7 +26,7 @@
  * @param {string} subheadingInclude Include src for the sub-heading area for each notification group, access the group via notificationGroup
  * @param {string} notificationBodyInclude Include src for the notification body for each notification, access the notification via notification
  * @param {string} notificationFooterInclude Include src for the notification footer for each notification, access the notification via notification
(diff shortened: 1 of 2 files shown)


AllenBW:
An epic shout-out thanks to @jeff-phillips-18 for the 👀 on this 😍 🙇‍♀️

PR #481, opened 2017-06-06
Hides selected filter results unless selected&totalCount !== undefined

### LINKED ISSUE #480: When no items can be selected we still display the "x of x results" text in the filter results
*The issue*: when no items _can_ be selected we still display the "x of x results" text in the filter results unless a `totalCount = 0` is supplied https://github.com/patternfly/angular-patternfly/blob/0a9afa136bf9bdb60bc3c9c96c9ec23c515cfc7b/src/filters/filter-results.html#L15

*Fix One*: allow `selectedCount` to be undefined, in which case hide the results (making the assumption that there are no items to select). This would make a change here https://github.com/patternfly/angular-patternfly/blob/0a9afa136bf9bdb60bc3c9c96c9ec23c515cfc7b/src/filters/filter-results-component.js#L66
*Fix Two*: pass a new param that is the truthy/falsey of `selectItems` found in each of the view types https://github.com/patternfly/angular-patternfly/blob/0a9afa136bf9bdb60bc3c9c96c9ec23c515cfc7b/src/views/listview/list-view.component.js#L17
*Fix Three* update documentation to instruct user to provide `totalCount = 0` when `selectItems` is false

I like fix two the most as its the most explicit, but its also the most invasive (more changes most places).

AllenBW:
Unless selectedcount and total count are explicitly defined, will not display the "x of x selected" filter results text.  Prevents unwanted text for instance, when there is no items to select

This changes behavior from being shown by default, to being shown when explicitly invoked, aligns closer with pf 3.x behavior

closes #480

### DIFF
--- a/src/filters/filter-results.html
+++ b/src/filters/filter-results.html
@@ -12,7 +12,7 @@
         </li>
       </ul>
       <p><a class="clear-filters" ng-click="$ctrl.clearAllFilters()" ng-if="$ctrl.config.appliedFilters.length > 0">Clear All Filters</a></p>
-      <div ng-if="$ctrl.config.totalCount !== 0" class="pf-table-view-selected-label">
+      <div ng-if="$ctrl.config.selectedCount !== undefined && $ctrl.config.totalCount !== undefined" class="pf-table-view-selected-label">
         <strong>{{$ctrl.config.selectedCount}}</strong> of <strong>{{$ctrl.config.totalCount}}</strong> selected
       </div>
     </div><!-- /col -->
--- a/src/filters/filter-results-component.js
+++ b/src/filters/filter-results-component.js
@@ -63,12 +63,6 @@ angular.module('patternfly.filters').component('pfFilterResults', {
       if (ctrl.config.resultsCount === undefined) {
         ctrl.config.resultsCount = 0;
       }
-      if (ctrl.config.selectedCount === undefined) {
-        ctrl.config.selectedCount = 0;
-      }
-      if (ctrl.config.totalCount === undefined) {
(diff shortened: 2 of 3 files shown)


src/filters/filter-results.html
```diff
@@ -12,7 +12,7 @@
         </li>
       </ul>
       <p><a class="clear-filters" ng-click="$ctrl.clearAllFilters()" ng-if="$ctrl.config.appliedFilters.length > 0">Clear All Filters</a></p>
-      <div ng-if="$ctrl.config.totalCount !== 0" class="pf-table-view-selected-label">
+      <div ng-if="$ctrl.config.selectedCount !== undefined && $ctrl.config.totalCount !== undefined" class="pf-table-view-selected-label">
         <strong>{{$ctrl.config.selectedCount}}</strong> of <strong>{{$ctrl.config.totalCount}}</strong> selected
```
dtaylor113: I think this can be rewritten as:

```ng-if="$ctrl.config.selectedCount && $ctrl.config.totalCount```

dtaylor113: Also, could you please add a unit test for this?
Thanks,
- Dave

AllenBW: @dtaylor113 would it make more sense to add the tests `angular-patternflytest/filters/filter.spec.js`? (sure as crap is easier 😆 )

dtaylor113: Sure, that would be fine. I just didn't see any 'selection' type tests in filter.spec, but there are some in list-view.spec.  Whatever's easier is fine. -thanks

PR #489, opened 2017-06-14
Renames colummn to column

### LINKED ISSUE #488: Table component binding misspelling
https://github.com/patternfly/angular-patternfly/blob/master/src/table/tableview/table-view.component.js#L5

Should be `column`

AllenBW:
This is a non-breaking change for when colummn is present, it will be copied to column
The only thing we lose here is making column a required binding for the table component, kinda unavoidable though

closes #488

### DIFF
--- a/src/table/tableview/table-view.component.js
+++ b/src/table/tableview/table-view.component.js
@@ -2,7 +2,8 @@ angular.module('patternfly.table').component('pfTableView', {
   bindings: {
     config: '<?',
     dtOptions: '<?',
-    colummns: '<',
+    colummns: '<?',
+    columns: '<?',
     items: '<',
     actionButtons: '<?',
     menuActions: '<?',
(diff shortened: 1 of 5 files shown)


src/table/tableview/table-view.component.js
```diff
@@ -41,6 +42,9 @@ angular.module('patternfly.table').component('pfTableView', {
         $log.debug("$onInit");
       }
 
+      if (angular.isDefined(ctrl.colummns)) {
+        ctrl.columns = ctrl.colummns;
+      }
```
jeff-phillips-18: should probably check if columns is undefined. If both are set (for some strange reason, we should use columns over columms)

dtaylor113:
Thanks @AllenBW,  can't believe that spelling error was so prevalent and I didn't see it :-)

src/table/tableview/table-view.component.js
AllenBW: NICE 🍨 new commit 🤘

AllenBW:
@dtaylor113 anytime!! 🙇 💃 🌮

PR #498, opened 2017-06-19
Add support for List View compound expansion

### LINKED ISSUE #467: pfListView: always shows pointer cursor
I believe pfListView should show a pointer cursor only when the list items are selectable by clicking on them or when they can be dragged and dropped. In all other scenarios, the default cursor seems more appropriate and more in line with PatternFly's design.

jeff-phillips-18:
Adding config.compoundExpansion when config.useExpanding rows is set will not show the row expand icon/indicator at the beginning of the row. The consumer will need to set the items' isExpanded property and determine which value is being expanded.

Also changes the cursor for the row to only show the pointer for non-compound expansion or when row selection is set.

Fixes #467


src/views/listview/list-view.html
```diff
@@ -13,12 +13,12 @@
...
          dnd-dragend="$ctrl.dragEnd()"
          ng-class="{'drag-original': $ctrl.isDragOriginal(item), 'pf-selectable': $ctrl.selectItems, 'active': $ctrl.isSelected(item), 'disabled': $ctrl.checkDisabled(item), 'list-view-pf-expand-active': item.isExpanded}">
-      <div class="list-group-item-header">
+      <div class="list-group-item-header" ng-class="{'list-group-item-not-selectable' : !$ctrl.config.selectItems && (!$ctrl.config.useExpandingRows || $ctrl.config.compoundExpansion)}">
         <div class="list-view-pf-dnd-drag-items" ng-if="$ctrl.config.dragEnabled === true">
           <div pf-transclude="parent" class="list-view-pf-main-info"></div>
         </div>
         <div ng-class="{'list-view-pf-dnd-original-items': $ctrl.config.dragEnabled === true}">
-          <div class="list-view-pf-expand" ng-if="$ctrl.config.useExpandingRows">
+          <div class="list-view-pf-expand" ng-if="$ctrl.config.useExpandingRows && !$ctrl.config.compoundExpansion">
```
dlabrecq: At first glance, it appears that row and compound expansion are mutually exclusive? I spoke with Cat, Serena, etc. and we don't want to prevent both from being used at the same time. In fact, there is an openshift.io design that will use both.

Perhaps this property can be renamed something like useCompoundExpansionOnly, just to make it clear?

jeff-phillips-18: updated to 'compoundExpansionOnly

PR #503, opened 2017-06-23
Build Package Update

cdcabrera:
The PatternFly references are now disconnected from the main build process, so they've been updated independently. 

Using the ```caret``` to indicate latest release of PatternFly up to, but not including the next major so we can avoid having to update the references on every release of AngularPF until the next major of PatternFly.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -47,7 +47,7 @@
     "datatables.net": "^1.10.12",
     "datatables.net-select": "~1.2.0",
     "lodash": "4.x",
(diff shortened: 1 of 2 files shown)


bleathem:
I suggest we go wtih minimum version we know we are compatible with.  Eg. `>=3.25.1`.  that way users can be sure the prerequisite styles are in place.  Thoughts @jeff-phillips-18?

jeff-phillips-18:
Fixes #430

PR #504, opened 2017-06-26
Fix #502 - pfTableView: does not work with items retrieved via AJAX call

### LINKED ISSUE #502: pfTableView: does not work with items retrieved via AJAX call
When I initialize the items array with an empty array and invoke a service to retrieve the items, the table component does not show the retrieved items when they are available. It shows a message saying "No Items Available".
Plunker code: http://plnkr.co/edit/UxZ0TizJJr7Dx5hi4TMO?p=preview

alexkieling:
Fixed Table View so I can initialize the items array with a null value, fetch the data from an external source, and update the items array with the fetched data. Without this fix, initializing the items array with null throws an error and initializing the items array with empty array works but you see the 'No Items Available' message and updating the items data doesn't remove the message.

### DIFF
--- a/src/table/tableview/examples/table-view-basic.js
+++ b/src/table/tableview/examples/table-view-basic.js
@@ -34,180 +34,201 @@
   *   </ul>
   * @param {object} emptyStateConfig Optional configuration settings for the empty state component.  See the {@link patternfly.views.component:pfEmptyState Empty State} component
   * @example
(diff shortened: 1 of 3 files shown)


alexkieling:
When will the code be merged?

dtaylor113:
Hi @alexkieling, nice work!  Really appreciate the unit tests!

PR #506, opened 2017-06-27
Fix #494 - pfTableView: Need to be able to not show checkboxes

### LINKED ISSUE #494: pfTableView:  Need to be able to not show checkboxes
Need to add a 'showCheckboxes' Boolean attribute to the config object (default is 'true') which will control showing/hiding the initial column of checkboxes used for row selection.

alexkieling:
Added a new configuration property called `showCheckboxes` to show/hide the column with checkboxes. The default is true.


alexkieling:
Amended my commit with the changes to the example. Note that the configuration property is used during component initialization. Therefore, the component is removed and added to the DOM when you click on the 'Show Checkboxes' checkbox.

dtaylor113:
Hi, @alexkieling , looks good except I notice that when I check the 'Show Checkbox' checkbox it does hide the checkboxes, but remains 'checked'.  If the checkboxes aren't shown, then the 'Show Checkboxes' checkbox should not be checked  -man, a lot of 'check' talk :-)

alexkieling:
My bad. Fixed the issue.

Issue #510, opened 2017-07-02
Unable to add action buttons to empty state in view.


rhwood:
The list/card/table views have a parameter to set the configuration of an empty state view, but do not have a parameter to set the action buttons for the configured empty state view.

cdcabrera:
Merged

PR #511, opened 2017-07-03
Allow action buttons in empty state for a card/list/table view

rhwood:
Add ability to put action buttons in the empty state view for card, list, or table view.
Addresses #510.

---
This change is [Reviewable](https://reviewable.io/reviews/patternfly/angular-patternfly/511)

### DIFF
--- a/src/views/listview/examples/list-view.js
+++ b/src/views/listview/examples/list-view.js
@@ -58,7 +58,7 @@
  * @param {function (action, item))} updateMenuActionForItemFn function(action, item) Used to update a menu action based on the current item
  * @param {object} customScope Object containing any variables/functions used by the transcluded html, access via $ctrl.customScope.<xxx>
  * @param {object} emptyStateConfig Optional configuration settings for the empty state component.  See the {@link patternfly.views.component:pfEmptyState Empty State} component
-* @param {array} emptyStateActionButtons Optional buttons to display under the icon, title, and informational paragraph in the empty state component.  See the {@link patternfly.views.component:pfEmptyState Empty State} component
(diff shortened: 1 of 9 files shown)


src/views/listview/examples/list-view.js
```diff
@@ -318,6 +320,30 @@
           }
         };
 
+        $scope.emptyStateActionButtons = [
+          {
+            name: 'Main Action',
+            title: 'Perform an action',
+            actionFn: performAction,
```
dtaylor113: Hi, I noticed in these examples the performAction doesn't update the Events text area.  I think because the callback function for emptyState action buttons only return an 'action' and not a 'item', so you can't reuse the existing performatAction(action, item) function already there.

dtaylor113:
Other than the one issue I raised, LGTM!

rhwood:
For some reason, which I can't figure out, the actionFn property is becoming undefined for the empty state buttons in any/all examples. I have examples that use a different function for the actionFn property, but when the button is clicked in a browser, that property is undefined.

dtaylor113:
> For some reason, which I can't figure out, the actionFn property is becoming undefined

Hi, you need to put the actionFn above/before the emptyStateActionButtons array/obj definition.  This worked for me:

```
          var performEmptyStateAction = function (action) {
            $scope.eventText = action.name + " executed\r\n" + $scope.eventText;
          };

          $scope.emptyStateActionButtons = [
            {
              name: 'Main Action',
              title: 'Perform an action',
              actionFn: performEmptyStateAction,
              type: 'main'
            },
```

PR #512, opened 2017-07-05
pfTableView: Allow use of custom templates to render cells.  Added icons, links, and truncated text

dtaylor113:
In the screenshot below:

- Icons added to the Status column
- Links added to the Name column

image

- Truncated Text added to the Address column
cell_tooltips

@akieling , @serenamarie125 

JIRA: https://patternfly.atlassian.net/browse/OSUX-28

### DIFF
--- a/src/table/tableview/examples/table-view-basic.js
+++ b/src/table/tableview/examples/table-view-basic.js
@@ -16,10 +16,20 @@
   * </ul>
   * @param {object} dtOptions Optional angular-datatables DTOptionsBuilder configuration object.  See {@link http://l-lin.github.io/angular-datatables/archives/#/api angular-datatables: DTOptionsBuilder}
   * @param {array} items Array of items to display in the table view.
-  * @param {array} columns Array of table column information to display in the table's header row
+  * @param {array} columns Array of table column information to display in the table's header row and optionaly render the cells of a column.
   * <ul style='list-style-type: none'>
   *   <li>.header     - (string) Text label for a column header
   *   <li>.itemField    - (string) Item field to associate with a particular column.
+  *   <li>.htmlTemplate - (string) (optional) id/name of an embedded ng/html template. Ex: htmlTemplate="name_template.html".  The template will be used to render each cell of the column.
+  *        Use <code>handleColAction(key, value)</code> in the template to call the <code>colActionFn</code> callback function you specify. 'key' is the item attribute name; which should equal the itemFld of a column.
+  *       'value' is the item[key] value.
+  *   <pre>
+  *     <script type="text/ng-template" id="name_template.html">
+  *       <a href="" ng-click="$ctrl.handleColAction(key, value)">{{value}}</a>
+  *     </script>
+  *   </pre>
+  *   <li>.colActionFn - (function) (optional) Callback function used for the column. 'value' is passed as a paramenter to the
+  *        callback function.
   * </ul>
   * @param {array} actionButtons List of action buttons in each row
   *   <ul style='list-style-type: none'>
@@ -70,6 +80,15 @@
             <textarea rows="10" class="col-md-12">{{eventText}}</textarea>
           </div>
     </div>
+    <script type="text/ng-template" id="status_template.html">
+      <span ng-if="value === 'error'" class="pficon pficon-error-circle-o"></span>
+      <span ng-if="value === 'warning'" class="pficon pficon-warning-triangle-o"></span>
+      <span ng-if="value === 'ok'" class="pficon pficon-ok"></span>
+      {{value}}
+    </script>
+    <script type="text/ng-template" id="name_template.html">
+      <a href="" ng-click="$ctrl.handleColAction(key, value)">{{value}}</a>
+    </script>
   </file>
 
   <file name="module.js">
@@ -84,7 +103,8 @@
           };
 
           $scope.columns = [
-            { header: "Name", itemField: "name" },
+            { header: "Status", itemField: "status", htmlTemplate: "status_template.html" },
+            { header: "Name", itemField: "name", htmlTemplate: "name_template.html", colActionFn: onNameClick },
             { header: "Address", itemField: "address"},
             { header: "City", itemField: "city" },
             { header: "State", itemField: "state"}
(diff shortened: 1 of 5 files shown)


dlabrecq:
So datatables.net allows users to provide the HTML table layout if they choose. In fact, we have an example of this in the patternfly test pages. Does the Angular implementation prevent that?

dtaylor113:
Hi @dlabrecq, Do you mean this:  https://github.com/patternfly/patternfly/blob/master/tests/pages/_includes/widgets/table-view/js/config.js

Unfortunately, the angular-datatables plug-in cannot use the dt-column directive:  "When using the angular way, you CANNOT use the dt-column directive. ...If you need to provide some options to your column, your must provide the dt-column-defs"

The dt-column-defs API does have a renderWith(r) call, but examples do not show embedded HTML.  Also, the datatables HTML is embedded in the template of the a-pf pfTableView component and we don't expose it to the end users/devs.   

But you raise a good point, let me try out the renderWith(r) call.

dlabrecq:
Nope, I'm speaking about HTML (DOM) sourced data as shown below. 

https://github.com/patternfly/patternfly/blob/master/tests/pages/_includes/widgets/table-view/tmpl/table-all.html

The datatables.net allows a few ways of providing data:

- [HTML (DOM) sourced data](https://datatables.net/examples/data_sources/dom.html)
- [Ajax sourced data](https://datatables.net/examples/data_sources/ajax.html)
- [Javascript sourced data](https://datatables.net/examples/data_sources/js_array.html)
- [Server-side processing](https://datatables.net/examples/data_sources/server_side.html)

I'm asking because that would eliminate the need for templates. If don't prevent users of providing data in other ways, they could hook to a database, for example.

dtaylor113:
Hi, in angular-patternfly, we do not expose the underlying HTML to the end-user, it is in the component's HTML template.  This allows an end-user to simply provide an array of item objects (the data) and an array of columns (defines column headers, etc..).   While this allows for quick/easy setup, only one angular element/component, instead of a large amount of HTML.  The trade-off is how does one customize a column/cell?   This PR attempts to provide this.

I did try the renderWith API call:
```
ctrl.dtColumnDefs[3].renderWith (function (data, type, full) {
      return '<i>' + full.address + '</i>';
});
```
But unfortunately, the 'full' attribute is returning the HTML for the column, which includes a lot of commented out angular ng-if statements, not something like 'full.address" to reflect the underlying data, so I think HTML templates is the best way to go.
Thanks,
- Dave

src/table/tableview/examples/table-view-basic.js
```diff
@@ -16,10 +16,20 @@
   * </ul>
   * @param {object} dtOptions Optional angular-datatables DTOptionsBuilder configuration object.  See {@link http://l-lin.github.io/angular-datatables/archives/#/api angular-datatables: DTOptionsBuilder}
   * @param {array} items Array of items to display in the table view.
-  * @param {array} columns Array of table column information to display in the table's header row
+  * @param {array} columns Array of table column information to display in the table's header row and optionaly render the cells of a column.
```
cdcabrera: Minor... ```optionally```

src/table/tableview/examples/table-view-with-toolbar.js
```diff
@@ -12,13 +12,23 @@
  *   <li>.onCheckBoxChange    - ( function(item) ) Called to notify when a checkbox selection changes, default is none
  *   <li>.itemsAvailable      - (boolean) If 'false', displays the {@link patternfly.views.component:pfEmptyState Empty State} component.
  *   <li>.showCheckboxes      - (boolean) Show checkboxes for row selection, default is true
-* </ul>
+ * </ul>
  * @param {object} dtOptions Optional angular-datatables DTOptionsBuilder configuration object.  See {@link http://l-lin.github.io/angular-datatables/archives/#/api angular-datatables: DTOptionsBuilder}
  * @param {array} items Array of items to display in the table view.
- * @param {array} columns Array of table column information to display in the table's header row
+ * @param {array} columns Array of table column information to display in the table's header row and optionaly render the cells of a column.
```
cdcabrera: Same as above, ```optionally```

akieling: @cdcabrera I think $filter('filter') is being used incorrectly in some cases where we need an exact match. I submitted a PR a while ago using lodash but it could have been pure javascript as well.

cdcabrera: good good @akieling  I'll add another inline callout on this PR

src/table/tableview/table-view.component.js
```diff
@@ -339,6 +339,29 @@ angular.module('patternfly.table').component('pfTableView', {
...
     };
 
+    ctrl.hasHTMLTemplate = function (key) {
+      var htmlTemplate = this.getHTMLTemplate(key);
+      return htmlTemplate.length > 0;
+    };
+
+    ctrl.getHTMLTemplate = function (key) {
+      var retVal = '';
+      var tableCol = $filter('filter')(ctrl.columns, {itemField: key});
```
cdcabrera: Preference of LoDash vs $filter ... looks like acceptable behavior based on the documentation, personal preference, or something Angular specific?

src/table/tableview/table-view.component.js
```diff
@@ -339,6 +339,29 @@ angular.module('patternfly.table').component('pfTableView', {
...
+
+      if (tableCol && tableCol.length === 1 && tableCol[0].hasOwnProperty('htmlTemplate')) {
+        retVal = tableCol[0].htmlTemplate;
+      }
+      return retVal;
+    };
+
+    ctrl.handleColAction = function (key, value) {
+      var tableCol = $filter('filter')(ctrl.columns, {itemField: key});
```
cdcabrera: Same LoDash vs $filter comment...

src/table/tableview/table-view.component.js
dtaylor113: Previously, we tried to limit our use of LoDash just to minimize upgrade concerns.

src/table/tableview/table-view.component.js
dtaylor113: I think since this is an angular repo., it's better to use the angular $filter

PR #514, opened 2017-07-07
pfPagination: pagination controls

dtaylor113:
## Common pagination controls to be used with List, Card, and Table views

image

@akieling, @serenamarie125. @LHinson 

JIRA: https://patternfly.atlassian.net/browse/OSUX-25

### DIFF
--- a/src/pagination/pagination.component.js
+++ b/src/pagination/pagination.component.js
@@ -1,0 +1,147 @@
+/**
+ * @ngdoc directive
+ * @name patternfly.pagination.component:pfPagination
+ * @restrict E
+ *
+ * @param {number} pageNumber The current page number
+ * @param {number} numTotalItems The total number of items in the data set.  When a filter is applied, update the <code>numTotalItems</code>
+ * accordingly.
+ * @param {Array<Number>} pageSizeIncrements (optional) Page size increments for the 'per page' dropdown.  If not
+ * specified, the default values are: [5, 10, 20, 40, 80, 100]
+ * @param {number} pageSize (optional) The initial page size to use.  If not specified, the default will be
+ * the first increment in the <code>pageSizeIncrements</code> array.
+ * @description
+ * Component for pagination controls used in various views (list, card, table)
+ *
+ * @example
+ <example module="patternfly.pagination">
+
+ <file name="index.html">
+   <div ng-controller="PageCtrl">
+     <div class="col-md-12">
+       <div ng-repeat="item in items | startFrom:(pageNumber - 1)*pageSize | limitTo:pageSize" class="col-md-12">
+         <div class="row">
+           <div class="col-md-3">
+             <span>{{item.id}}</span>
+           </div>
+           <div class="col-md-7">
+             <span>{{item.status}}</span>
+           </div>
+           <div class="col-md-2">
+             <span>{{item.value}}</span>
+           </div>
+         </div>
+       </div>
+     </div>
+     <pf-pagination
+         page-size="pageSize",
+         page-number="pageNumber"
+         num-total-items="numTotalItems">
+     </pf-pagination>
+   </div>
+ </file>
+ <file name="script.js">
+ angular.module( 'patternfly.pagination').controller( 'PageCtrl', function( $scope ) {
+   $scope.pageSize = 10;
+   $scope.pageNumber = 1;
+
+   $scope.items = [];
+   for(i = 1; i <= 126; i++) {
+      $scope.items.push({id: i, status: 'Ok', value: Math.floor(Math.random() * (1000 - 1 + 1)) + 1});
+   }
+
+   $scope.numTotalItems = $scope.items.length;
+ });
+ </file>
+ </example>
+ */
+angular.module('patternfly.pagination').component('pfPagination', {
+  transclude: true,
+  templateUrl: 'pagination/pagination.html',
+  bindings: {
+    pageNumber: '=',
+    numTotalItems: "<",
+    pageSizeIncrements: '<?',
+    pageSize: "=?"
+  },
+  controller: function ($scope, $log) {
+    'use strict';
+    var ctrl = this;
+
+    var defaultPageSizeIncrements = [5, 10, 20, 40, 80, 100];
+
+    ctrl.$onInit = function () {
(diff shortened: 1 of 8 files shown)


src/pagination/pagination.component.js
```diff
@@ -0,0 +1,142 @@
+/**
+ * @ngdoc directive
+ * @name patternfly.pagination.component:pfPagination
+ * @restrict E
+ *
+ * @param {number} pageNumber The current page number
+ * @param {number} numTotalItems The total number of items in the data set.  When a filter is applied, update the <code>numTotalItems</code>
+ * accordingly.
+ * @param {number[]} pageSizeIncrements (optional) Page size increments for the 'per page' dropdown.  If not
```
cdcabrera: Just a question here... Do we aim to describe typed arrays?

src/pagination/pagination.html
```diff
@@ -0,0 +1,30 @@
...
+    </div>
+    <span class="per-page-label">per page</span>
+  </div>
+  <div class="form-group">
+    <span><span class="pagination-pf-items-current">{{$ctrl.getStartIndex()}}-{{$ctrl.getEndIndex()}}</span> of <span class="pagination-pf-items-total">{{$ctrl.numTotalItems}}</span></span>
+    <ul class="pagination pagination-pf-back">
+      <li ng-class="{'disabled': $ctrl.pageNumber === 1}"><a title="First Page" ng-click="$ctrl.gotoFirstPage()" class="goto-first-page"><span class="i fa fa-angle-double-left"></span></a></li>
+      <li ng-class="{'disabled': $ctrl.pageNumber === 1}"><a title="Previous Page" ng-click="$ctrl.gotoPreviousPage()" class="goto-prev-page"><span class="i fa fa-angle-left"></span></a></li>
+    </ul>
+    <input class="pagination-pf-page" type="text" ng-model="$ctrl.pageNumber" ng-change="$ctrl.onPageNumberChange()"/>
```
cdcabrera: Appears the field lets you continue to a blank page if you manually enter pages that don't exist, maybe add a check to ignore higher than available pages.

src/pagination/pagination.component.js
```diff
@@ -0,0 +1,142 @@
...
+    pageNumber: '=',
+    numTotalItems: "<",
+    pageSizeIncrements: '<?',
+    pageSize: "=?"
+  },
+  controller: function ($scope, $log) {
+    'use strict';
+    var ctrl = this;
+
+    var defaultPageSizeIncrements = [5, 10, 20, 40, 80, 100];
```
cdcabrera: Are we concerned about displaying large increments when the number of results is exceedingly low? Like say 2...

src/pagination/pagination.service.js
```diff
@@ -0,0 +1,24 @@
...
+        // TODO: may not need this service
+      };
+
+      return {
+        setPages: setPages,
+      };
+    })
+    .filter('startFrom', function () {
+      return function (input, start) {
+        start = +start; //parse to int
```
cdcabrera: Hadn't seen that shortcut in awhile, we could remove the explanation by using ```parseInt(start)```.

src/pagination/pagination.service.js
```diff
@@ -0,0 +1,24 @@
+(function () {
```
dlabrecq: It appears that the pagination service isn't currently used by the pagination component? (And the setPages function isn't implemented in the service itself.) Perhaps we should omit this until we can show an example that works with table and list view, for example?

dtaylor113: Hmm..yeah, as the comment in there says it's a placeholder because we may need a service to hold variables to hold the current pagination configs (pageSize, pageNum, etc..) when switching between List, Card, and Table views, but I won't know until I integrate them.

I'll remove the service for now and add it later if needed -thanks

src/pagination/pagination.component.js
dtaylor113: I don't see the harm in providing this info., just to let them know we're expecting an array of numbers.  I can trim it to just 'Array' if you feel that's better.

src/pagination/pagination.component.js
dtaylor113: I'm not quite following, if you only have two results/items, but a large pageSize, like 50, it would still only show one page with the two results.

src/pagination/pagination.html
dtaylor113: Ah..good catch!

src/pagination/pagination.service.js
dtaylor113: Yeah, this was cut&paste code, I'll adjust it.  -thanks

src/pagination/pagination.service.js
dtaylor113: Fixed

src/pagination/pagination.component.js
dtaylor113: Fixed to: Array<Number> to match other ngDoc examples in a-pf repo.

src/pagination/pagination.html
dtaylor113: Fixed, defaults to last page if number > last page.  Hmm..I should probably do the same for before first page (0, -1, etc).

src/pagination/pagination.component.js
cdcabrera: Ah yes, not so clear. Had to do with me reducing the number of table entries to "2" then still seeing the page results dropdown showing increments up to 100.

src/pagination/pagination.html
dtaylor113: Fixed

src/pagination/pagination.component.js
dtaylor113: Ok, yeah, still think it's valid to always show all the increments, regardless of results, keeps it more consistent. -thanks

src/pagination/pagination.html
cdcabrera: We may have inadvertently caused an additional issue, blanking out the field does this... ```NaN-NaN```

screen shot 2017-07-10 at 12 17 16 pm

dtaylor113: Hi, fixed issue, but now you have to enter the number and then hit Tab or Enter to apply the new page number.

PR #515, opened 2017-07-07
pfTableView: Truly turn off pagination by default

### LINKED ISSUE #513: pfTableView: paging is enabled by default, even though paging controls are not shown by default
Need to add paging=false to the default dt options in pfTableView:

```
ctrl.defaultDtOptions = {
      ...
      paging: false;
    };
```

dtaylor113:
By default pagination controls are not displayed in the pfTableView.  Unfortunately, items where still being paginated; as a result only the first 10 items would be displayed and with no pagination controls, there was no way to get to the rest of the items.

This PR sets 'pagination=false' in the default dt options, and thus shows all items.

Fixes #513 

This will all go away when we integrate the pfPagination component into pfTableView.
@akieling

### DIFF
--- a/src/table/tableview/examples/table-view-basic.js
+++ b/src/table/tableview/examples/table-view-basic.js
@@ -235,6 +235,30 @@
               city: "New York",
               state: "New York"
               },
+              {
+              name: "Mike Bird",
+              address: "50 Forth Street",
+              city: "New York",
+              state: "New York"
+              },
+              {
+              name: "Cheryl Taylor",
+              address: "2 Main Street",
+              city: "New York",
+              state: "New York"
+              },
+              {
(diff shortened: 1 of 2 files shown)


src/table/tableview/examples/table-view-basic.js
```diff
@@ -235,6 +235,30 @@
...
+              address: "10 Chase Lane",
+              city: "Boston",
+              state: "Massacusetts"
+              },
+              {
+              name: "Kim Livingston",
+              address: "5 Tree Hill Lane",
+              city: "Boston",
+              state: "Massacusetts"
+              }
```
dtaylor113: Added more items to the 'Basic' example in order to show that all 12 items are being displayed when pagination is turned off by default.

PR #527, opened 2017-07-26
pfTableView with pfPagination

dtaylor113:
Integration of pfPagination controls into pfTableView

- [Forked](https://github.com/patternfly/angularjs-datatables) angular-datatables due to [issue](https://github.com/l-lin/angular-datatables/issues/262).  Angular-datatables has moved on to supporting angular 2 & 4, not sure if they will ever get around to addressing this issue in angularjs. 
- Former method of specifying dtOptions for pagination supported for backwards compatibility
- Tested in pfTableView - Basic, pfTable, pfTableView - w/ Toolbar, and pfToolbar examples
- Tested with Filtering, Sorting, Select All, Empty State, show/hide checkboxes, and Action buttons/menus

image

**Updated ngDocs:**
image

@akieling


jeff-phillips-18:
The pagination bar doesn't seem to right align correctly. Though, neither does the toolbar.

test/table/tableview/table-view.spec.js
```diff
@@ -108,10 +141,25 @@ describe('Component: pfTableView', function () {
...
+
+  it('should show pagination controls when configured', function () {
+    paginationSetup();
+    expect(element.find('pf-pagination').length).toBe(1);
+
+    expect(angular.element(element.find('.pagination-pf-items-current')).text().trim()).toBe('3-4');  // page # 2
+    expect(angular.element(element.find('.pagination-pf-items-total')).text().trim()).toBe('7');
+    expect(angular.element(element.find('.pagination-pf-page')).val().trim()).toBe('2');
+    expect(angular.element(element.find('.pagination-pf-pages')).text().trim()).toBe('4');
+  });
```
jeff-phillips-18: Please add tests to make sure updatePageSize and updatePageNumber functions are called correctly when appropriate.

src/table/table.less
```diff
@@ -27,6 +27,11 @@ table.dataTable {
     }
   }
   tbody {
+    tr.selected {
+      span a {
+        color: white;
```
jeff-phillips-18: use @color-pf-white

src/pagination/pagination.component.js
```diff
@@ -131,11 +146,22 @@ angular.module('patternfly.pagination').component('pfPagination', {
 
     ctrl.getEndIndex = function () {
       var numFullPages = Math.floor(ctrl.numTotalItems / ctrl.pageSize);
-      var numItemsOnLastPage = ctrl.numTotalItems - (numFullPages * ctrl.pageSize);
+      var numItemsOnLastPage = ctrl.numTotalItems - (numFullPages * ctrl.pageSize) || ctrl.pageSize;
       var numItemsOnPage = isLastPage() ? numItemsOnLastPage : ctrl.pageSize;
       return ctrl.getStartIndex() + numItemsOnPage - 1;
```
jeff-phillips-18: If there are no items, this value would incorrectly be pageSize - 1.

test/table/tableview/table-view.spec.js
dtaylor113: Hi, do you mean with spy functions, or test changing the pageSize and pageNumber controls and getting the expected HTML?

jeff-phillips-18: Spy functions work, or just by saving the parameters and checking the values after. Similar to

https://github.com/patternfly/angular-patternfly/blob/master/test/notification/notification-drawer.spec.js#L332-L337

and 

https://github.com/patternfly/angular-patternfly/blob/master/test/notification/notification-drawer.spec.js#L382-L389

dtaylor113: Did both, fixed

src/table/table.less
dtaylor113: Fixed

src/pagination/pagination.component.js
dtaylor113: Fixed

dtaylor113:
Hi @jeff-phillips-18, @cdcabrera,  Does this PR warrant a Major version change since it has breaking changes?  pfTableView won't work after this PR until end-users update their <scripts> to point to "angularjs-datatables" from "angular-datatables".

cdcabrera:
@Dtaylor113 I would concur, breaking change equals major increment

dtaylor113:
> are we going to need to release package this change immediately, or is it something we can package with bug fixes?

@cdcabrera This is an enhancement which OpenShift will "eventually" need, but it is not immediate.  @serenamarie125 do you know when OpenShift will want pfTableView w/ pagination?

jeff-phillips-18:
I’d say keep it to a minor release. It won’t be worse than now if the application doesn't update the dependency, it will just use the old dependency and still have the bug. If the application is using wiredep, then all should be good.

dtaylor113:
> I’d say keep it to a minor release.

So, v4.3.0 ?

dtaylor113:
Hi, so can this be merged?

cdcabrera:
lgtm

PR #536, opened 2017-07-31
Adds complex filter select

### LINKED ISSUE #521: pfFilter: RFE - Multi-dropdown filter, filterAlias, input typeahead
As per: https://github.com/ManageIQ/manageiq-ui-service/pull/840#issuecomment-315496383 being able to display a second dropdown that further breaks up filter content grouping would be aweeeesommeee

ALSO if [filter fields](https://github.com/patternfly/angular-patternfly/blob/master/src/filters/filter-component.js#L29-L35) supported value aliases or in other words an additional filter value field available for passing a value around, for when the consumer of the filter requires drastically different formatting than what is pleasant for users to view

ALSO if filter inputs were typeaheads

Apologies if this comes across as needy, sometimes a fella just wanta extra special filtering fun! 🌮 💃 

ps: mostly ordered by level of how awesome each would be <3

pss:  the highest impact easiest win here would be `filterAlias` could temporarily alienate need for multi filter dropdowns

AllenBW:
closes #521 

Gonna keep description here skinny, force ya to read updated docs :trollface: (its the only way to make em better)

[design mockup for this work lives here](https://github.com/ManageIQ/manageiq-ui-service/pull/840#issuecomment-315496383)

### ~~Our Current State~~ See gif below
complex-filter

~~**WIP**ping this just for posterity, so visibility on the direction exists, holding off on reviewers until there's something real pretty to review.~~

TODO:
* ~~tests tests tests~~
* ~~code cleanup~~
* ~~ensure category filterValue selection highlight in dropdown~~
* ~~documentation documentation documentation~~

🌮 💃 @chriskacerguis

### DIFF
--- a/src/filters/filters.less
+++ b/src/filters/filters.less
@@ -2,6 +2,21 @@
   a {
     cursor: pointer;
   }
+  .input-group {
+    .input-group-btn {
+      .dropdown-menu>.selected>a {
+        background-color: @color-pf-blue !important;
+        border-color: #0076b7 !important;
+        color: @color-pf-white !important;
+      }
+    }
+  }
+  .category-select{
+    display: flex;
+  }
+  .category-select-value{
+    border-left-width: 0 !important;
+  }
 }
 .filter-pf.filter-fields {
   .form-group {
@@ -17,15 +32,6 @@
     font-size: 12px;
     font-style: italic;
     font-weight: 400;
-  }
-}
-.input-group {
-  .input-group-btn {
-    .dropdown-menu>.selected>a {
-      background-color: @color-pf-blue !important;
-      border-color: #0076b7 !important;
-      color: @color-pf-white !important;
-    }
   }
 }
 
--- a/src/filters/simple-filter/filter-component.js
+++ b/src/filters/simple-filter/filter-component.js
@@ -18,8 +18,17 @@ angular.module('patternfly.filters').component('pfFilter', {
     };
 
     function filterExists (filter) {
-      var foundFilter = _.find(ctrl.config.appliedFilters, {title: filter.title, value: filter.value});
-      return foundFilter !== undefined;
+      return angular.isDefined(_.find(ctrl.config.appliedFilters, {title: filter.title, value: filter.value}));
+    }
+
+    function findDuplicateCategory (field, value) {
+      var duplicateValue;
+
+      function searchAppliedFilters (item) {
+        return _.includes(item.value, _.split(value, field.filterDelimiter, 1)) ? duplicateValue = item : null;
+      }
+
+      return _.some(ctrl.config.appliedFilters, searchAppliedFilters) ? duplicateValue : null;
     }
 
     function enforceSingleSelect (filter) {
@@ -33,10 +42,15 @@ angular.module('patternfly.filters').component('pfFilter', {
         type: field.filterType,
         value: value
       };
+
       if (!filterExists(newFilter)) {
 
         if (newFilter.type === 'select') {
           enforceSingleSelect(newFilter);
+        }
+
+        if (field.filterType === 'complex-select' && !field.filterMultiselect && findDuplicateCategory(field, value)) {
+          _.remove(ctrl.config.appliedFilters, findDuplicateCategory(field, value));
         }
 
         ctrl.config.appliedFilters.push(newFilter);
--- a/src/filters/examples/filter.js
+++ b/src/filters/examples/filter.js
@@ -14,8 +14,12 @@
  * <li>.id          - (String) Optional unique Id for the filter field, useful for comparisons
  * <li>.title       - (String) The title to display for the filter field
  * <li>.placeholder - (String) Text to display when no filter value has been entered
- * <li>.filterType  - (String) The filter input field type (any html input type, or 'select' for a single select box)
- * <li>.filterValues - (Array) List of valid select values used when filterType is 'select'
+ * <li>.filterMultiselect - (Boolean) In `complex-select`, allow selection of multiple values per category. Optional, default is `false`
+ * <li>.filterType  - (String) The filter input field type (any html input type, or 'select' for a single select box or 'complex-select' for a category select box)
+ * <li>.filterValues - (Array) List of valid select values used when filterType is 'select' or 'complex-select' (in where these values serve as case insensitve keys for .filterCategories objects)
+ * <li>.filterCategories - (Array of (Objects)) For 'complex-select' only, array of objects whoes keys (case insensitive) match the .filterValues, these objects include each of the filter fields above (sans .placeholder)
+ * <li>.filterCategoriesPlaceholder - (String) Text to display in `complex-select` category value select when no filter value has been entered, Optional
+ * <li>.filterDelimiter - (String) Delimiter separating 'complex-select' category and value. Optional, default is a space, ' '
  * </ul>
  * <li>.appliedFilters - (Array) List of the currently applied filters
  * <li>.resultsCount   - (int) The number of results returned after the current applied filters have been applied
@@ -45,6 +49,9 @@
             <div class="col-md-2">
               <span>{{item.birthMonth}}</span>
             </div>
+            <div class="col-md-4">
+              <span>{{item.car}}</span>
+            </div>
           </div>
         </div>
       </div>
@@ -67,27 +74,33 @@
           {
             name: "Fred Flintstone",
             address: "20 Dinosaur Way, Bedrock, Washingstone",
-            birthMonth: 'February'
+            birthMonth: 'February',
+            car: 'Toyota-Echo'
           },
           {
             name: "John Smith",
             address: "415 East Main Street, Norfolk, Virginia",
-            birthMonth: 'October'
+            birthMonth: 'October',
+            car: 'Subaru-Outback'
+
           },
           {
             name: "Frank Livingston",
             address: "234 Elm Street, Pittsburgh, Pennsylvania",
-            birthMonth: 'March'
+            birthMonth: 'March',
+            car: 'Toyota-Prius'
           },
           {
             name: "Judy Green",
             address: "2 Apple Boulevard, Cincinatti, Ohio",
-            birthMonth: 'December'
+            birthMonth: 'December',
+            car: 'Subaru-Impreza'
           },
           {
             name: "Pat Thomas",
             address: "50 Second Street, New York, New York",
-            birthMonth: 'February'
+            birthMonth: 'February',
+            car: 'Subaru-Outback'
           }
         ];
         $scope.items = $scope.allItems;
@@ -102,6 +115,8 @@
             match = item.address.match(re) !== null;
           } else if (filter.id === 'birthMonth') {
             match = item.birthMonth === filter.value;
+          } else if (filter.id === 'car') {
+            match = item.car === filter.value;
           }
           return match;
         };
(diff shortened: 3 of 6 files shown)


serenamarie125:
@AllenBW thanks for taking this on!  Just a reminder that both of those components should use the typeahead dialog select, as you did here: https://github.com/ManageIQ/manageiq-ui-service/pull/821

AllenBW:
@serenamarie125 this work belongs in another pr, especially since it touches legacy stuff (existing dropdown select)

chriskacerguis:
@serenamarie125 for posterity I've setup this story to track the change you wanted: https://www.pivotaltracker.com/story/show/149809336

src/filters/filters.less
```diff
@@ -28,3 +28,6 @@
     }
   }
 }
+.category-select{
+  display: flex;
+}
```
jeff-phillips-18: Prefix this with filter-pf or child it to the filter-pf selector.

Not due to this PR, but I see in the same file there is .input-group... could you child that to the filter-pf while you are in there?

jeff-phillips-18:
Currently it adds a filter when you select the filter value, nothing happens when selecting a category. There is no indication of a category selection. 

IMO, when I select a filter on the category, it should add that filter. So, I should see something like:

Active Filters:   Car/Make: Suburu  

Then if I select a model:

Active Filters:   Car/Make: Suburu    Car/Model: Outback

serenamarie125:
@AllenBW when the first drop down is selected, the value should be shown

src/filters/filter-fields.html
```diff
@@ -39,5 +39,45 @@
...
+
+      <div class="btn-group bootstrap-select form-control filter-select" uib-dropdown >
+        <button type="button" uib-dropdown-toggle class="btn btn-default dropdown-toggle">
+          <span class="filter-option pull-left">{{$ctrl.currentValue || $ctrl.currentField.filterCategoriesPlaceholder}}</span>
+          <span class="caret"></span>
+        </button>
+        <ul uib-dropdown-menu class="dropdown-menu-right" role="menu">
+          <li ng-if="$ctrl.currentField.placeholder">
+            <a role="menuitem" tabindex="-1" ng-click="$ctrl.selectValue()">
+              {{$ctrl.currentField.placeholder}}
```
jeff-phillips-18: this should be $ctrl.currentField. filterCategoriesPlaceholder

serenamarie125:
@jennyhaines can you take a look at this screenshot? Should the dropdown be a bit closer to one another?
screen shot 2017-08-01 at 1 51 40 pm

jennyhaines:
@serenamarie125 Yes! Just one pixel closer would look nicer. This would mean just one pixel line between each, rather than 2 pixels between.

src/filters/filter-fields.html
AllenBW: DOH yes it should, made a change, forgot to put it back.

AllenBW:
All, thank you IMMENSELY for the pr feedback!  Ok pleasantries aside gonna go down the line: 

@jeff-phillips-18  hoping you'll find (in the updated gif above) that updating placeholder value is sufficient for informing the user something has happened.  I'm mild to strongly opposed about adding a filter result prematurely as the case of a complex filter requires both parts to be valid, (category and value) imply a category alone is sufficient input on which to filter feels misleading.

@serenamarie125 @jennyhaines ok get ready for some ss!
### here you'll notice, when the right select is moved a pixel to the left, we now have a super odd box highlight when selecting the left box
screen shot 2017-08-01 at 5 47 32 pm

### but this box shadow looks fine on the right select
screen shot 2017-08-01 at 5 47 40 pm

### I propose we eliminate the left-border of the right select (instead of moving it -1 pixels to the left)  as this maintains the original box shadow (what this ss shows)
screen shot 2017-08-01 at 5 49 46 pm

### tada! normal ish looking!
screen shot 2017-08-01 at 6 01 34 pm

AllenBW:
complex-filter

@jeff-phillips-18 @serenamarie125 @jennyhaines  updated gif!

serenamarie125:
@jeff-phillips-18 brought up a good point to me earlier in that if you select Toyota/Prius & then Toyota/Corolla, the filters will be ANDed rather than ORed, and that this is different than our drop down filter where they replace an existing chip with the new value.

I initially didn't think this was a problem, but just looked back at this documentation : https://manageiq.github.io/manageiq-design/UX/common/Tags/design and see an alternate suggestion from @beanh66 a while back which allows multi-selection in the 2nd drop down

@beanh66 what are your thoughts here?  Should we enable multi select and then use the new Filter Result Tags which may contain multiple values which Dave Taylor has up https://github.com/patternfly/angular-patternfly/pull/540?

@AllenBW this is looking great, but wondering if we should make this small tweak.

jennyhaines:
@AllenBW - I like your solution of eliminating the left-border of the right select! This looks good.

beanh66:
@AllenBW I think what @jeff-phillips-18 and @serenamarie125 are describing is that the current implementation would allow a user to pick Toyota-Corolla AND Toyota-Prius as filter conditions which is impossible and would get zero results. Ideally, if a user selected "Car:Toyota-Corolla" and then made a selection of "Car:Toyota-Prius" it would replace the original selection since a car cannot be a toyota prius and corolla at the same time. Is it possible to make this change using the current styling you have to keep things simple?

beanh66:
@AllenBW Options (2) is if we wanted to get fancy and add the OR functionality. In that case it would be worth changing the visuals to match what Dave did for #540 as @serenamarie125 mentioned. The visuals would look like the following to allow a user to say I want to see Cars that are either Toyota-Corolla OR Toyota-Prius:
screen shot 2017-08-02 at 8 54 26 am

IMO this could be a separate PR to add this functionality in the future so we could stick with Option (1) for now if that makes sense to you!

AllenBW:
@beanh66 gaaaah I really wanta speak this over 👖 buuuuuut will give it a try here. So maybe the example is a bit misleading, but real world and also in the specific use case I have for building this, totally possible for an item to have different unique values from a same category.  I really would be a shame to lock this feature down to only one behavior, after all, filter is configurable enough that if a dev desires replace instead of add they could that functionality in the filter/results configs

AllenBW:
@beanh66 not opposed to that design, is not in the repo yet, if we want to use it, should wait until it is in, would be kinda insane to duplicate the code :-/ 

so YESSSSSS down to do that in a separate pr 🤘

serenamarie125:
@AllenBW appreciate your concern on locking down the behavior.  I have a suggestion.  For this filter type, could we have an "Replace/Single Value per category" option which is either:
- true (default setting) which is Colleen's Option1.  This would be consistent with the existing behavior when there is a single select ( see example with birth month ).  
- false, thus the user can add as many values per category as desired ( thus being ANDed, which is your current behavior )

And then create an issue covering the Option2  requirement

Thoughts on the above proposal @jeff-phillips-18 @AllenBW @beanh66 ?

One more nit @AllenBW - the separator between the category & value should be a ":" not a "-"

AllenBW:
@serenamarie125 not responding (yet, will tomorrow morn) but had to mention real quick that delimiter is customizable!  default of it is just a space  (unless we don't want it to be, but the case for it being customizable is for instance, in the sui, tag filtering is a string with items separated by `/`)

AllenBW:
@serenamarie125 ok reading and understanding da proposal, whole heartily down with more flexibility!  if this is the consensus (aka the thing that will get this pr in) will do 🤘

beanh66:
@serenamarie125 That sounds like a good proposal to me 👍  Thanks Allen!

jeff-phillips-18:
@serenamarie125 like the proposal, need to look into the existing code to see how it is effected to be sure there are no backward compatibility issues.

AllenBW:
@jeff-phillips-18  soooo being that result reporting and filtering are two separated components  in order to not have duplicates results, would have to track whats pushed to results, kinda give complex filtering a state (it is presently stateless, pushes results onwards never looks back)

If you're ok with me giving it a (mild) state i'll write this up and add the commit.

AllenBW:
@serenamarie125 @jeff-phillips-18 @beanh66 heeeeeey check it out, now ya can specify filter selection behavior! `.filterMultiselect` is the param name
## default is false, not allowed to filter multiselect, but wait for it, wanted to show all other functionality remained as is (didn't conflict in any odd ways)
complex-filter
## but you can specify true!
complex-filter

AllenBW:
@serenamarie125 heeeyoooooo never heard back on the latest updates to the behavior were solid? or are we gonna drop this feature, should I close the pr?

@jeff-phillips-18 should i reword last commit to follow the feature commit wording guidelines for this to make it in?

beanh66:
@AllenBW This looks great to me! I know @jeff-phillips-18 wanted to check on the back end to be sure there were no backward compatibility issues, so I'll leave that to him but the functionality looks great.

serenamarie125:
oh my @AllenBW 🤗 ❤️ this looks beautiful!!!

src/filters/simple-filter/filter-component.js
```diff
@@ -18,8 +18,7 @@ angular.module('patternfly.filters').component('pfFilter', {
     };
 
     function filterExists (filter) {
-      var foundFilter = _.find(ctrl.config.appliedFilters, {title: filter.title, value: filter.value});
-      return foundFilter !== undefined;
+      return angular.isUndefined(_.find(ctrl.config.appliedFilters, {title: filter.title, value: filter.value}));
```
jeff-phillips-18: I believe this should be angular.isDefined(...

src/filters/simple-filter/filter-component.js
```diff
@@ -33,18 +32,28 @@ angular.module('patternfly.filters').component('pfFilter', {
...
         ctrl.config.appliedFilters.push(newFilter);
 
         if (ctrl.config.onFilterChange) {
           ctrl.config.onFilterChange(ctrl.config.appliedFilters);
         }
       }
+
+      function complexFilterExists (item) {
+        return _.includes(item.value, _.split(value, field.filterDelimiter, 1)) ? matchedFilter = item : null;
+      }
```
jeff-phillips-18: I wouldn't expect the side effect of setting matchedFilter to be part of complexFilterExists.

src/filters/simple-filter/filter-component.js
AllenBW: @jeff-phillips-18 the logic is a bit 🐻 , but `isUndefined` maintains existing logic, you'll notice on line 37, the `!` is removed  my thought here was to wholly encapsulate the check in the function rather than spreading the logic in two places (though maybe its less intuitive now?)

(if changed to `isDefined`, filter does not work unless line 37 is restored to be negated, the first gif shows it working just fine)

SO I can do whatever ya want here, restore it to original orrr change to isundefined and restore line 37...

src/filters/simple-filter/filter-component.js
AllenBW: @jeff-phillips-18 unsure of what you're asking for here... can we do a 👖 to talk it out?

src/filters/simple-filter/filter-component.js
jeff-phillips-18: The function is 'filterExists'. This logic should return true if the filter exists. Line 37 below changed to be filterExists from !filterExists.  The maintenance of logic is due to two opposite checks to make it back to where it was.

AllenBW: Say no more, updated and pushed 🙇‍♀️

src/filters/simple-filter/filter-component.js
jeff-phillips-18: The logic of this in kinda janky.  A function called 'complexFilterExists' should return true/false without a side effect of setting the matchedFilter variable which is then used.

dtaylor113: I agree with @jeff-phillips-18 , 'complexFilterExists' should just return a boolean, and 'matchedFilter' should be handled in a diff function, perhaps 'getComplexFilter'.  'complexFilterExists' just seems overloaded and difficult to 'grok' :-)  Pls consider breaking it up for better readability.  -thanks

AllenBW: @jeff-phillips-18 Updated the pr to address the changes we discussed in slack, 

* rework of duplicate category logic to be more explicit! 
* wait, crap, i can't remember if there was anything else

jeff-phillips-18:
@AllenBW Could you squash your commits please?

AllenBW:
@jeff-phillips-18 all of em into that one that has (feat)?

jeff-phillips-18:
Yes, please

AllenBW:
@jeff-phillips-18 updated! 🙇‍♀️

PR #538, opened 2017-07-31
Issue Standalone Angular PatternFly

### LINKED ISSUE #534: "patternfly" & "c3ChartDefaults" undefined when running Angular Patternfly standalone
Unable to run Angular Patternfly JS as a standalone script reference without including core Patternfly JS because of of multiple references and calls to the global ```patternfly``` object and subsequent method call to ```c3ChartDefaults``` ... 

```javascript
patternfly.c3ChartDefaults()
```

Purposing a quick fix within ```c3-chart-defaults.js``` similar to...
```javascript
(function (window, patternfly) {
  'use strict';

  window.patternfly = patternfly || {};
  window.patternfly.c3ChartDefaults = patternfly.c3ChartDefaults || function(){ console.error('patternfly.c3ChartDefaults unavailable.\nhttps://github.com/patternfly/angular-patternfly'); return{} };

  var patternflyDefaults = patternfly.c3ChartDefaults();
```

cdcabrera:
## Description
Allows Angular Patternfly JS  to run as a standalone script reference without including core Patternfly JS. Unit test updates related to globals "d3" and "c3ChartDefaults"
- Closes #534 

#### Side-Note
  Had to relax the linter around "console statements", converted to warnings. If there's another way to handle it more than willing to convert it

## PR Checklist

- [x] Unit tests are included
- [n/a ] Screenshots are attached (if there are visual changes in the UI)
- [n/a] A Designer (@beanh66) is assigned as a reviewer (if there are visual changes in the UI)
- [n/a] A CSS rep (@cshinn) is assigned as a reviewer (if there are visual changes in the UI)

### DIFF
--- a/eslint.yaml
+++ b/eslint.yaml
@@ -35,7 +35,7 @@ rules:
   vars-on-top: 2
   no-debugger: 2
   no-cond-assign: 2
-  no-console: 2
+  no-console: 1
   no-extra-semi: 2
   no-irregular-whitespace: 2
   dot-notation:
--- a/src/charts/c3/c3-chart-defaults.js
+++ b/src/charts/c3/c3-chart-defaults.js
@@ -1,7 +1,17 @@
-(function () {
+(function (window, patternfly) {
   'use strict';
 
-  var patternflyDefaults = patternfly.c3ChartDefaults();
+  var patternflyDefaults;
+
+  var fallbackC3ChartDefaults = function () {
+    console.error('patternfly.c3ChartDefaults unavailable.\nhttps://github.com/patternfly/angular-patternfly');
+    return {};
+  };
+
+  window.patternfly = patternfly || {};
+  window.patternfly.c3ChartDefaults = patternfly.c3ChartDefaults || fallbackC3ChartDefaults;
+
+  patternflyDefaults = patternfly.c3ChartDefaults();
 
   angular.module('patternfly.charts').constant('c3ChartDefaults', {
     getDefaultColors: patternflyDefaults.getDefaultColors,
@@ -20,4 +30,6 @@
     getDefaultSparklineConfig: patternflyDefaults.getDefaultSparklineConfig,
     getDefaultLineConfig: patternflyDefaults.getDefaultLineConfig
   });
-})();
+
+})(window, patternfly);
+
--- a/test/charts/c3/c3.spec.js
+++ b/test/charts/c3/c3.spec.js
@@ -1,4 +1,4 @@
-describe('Directive: pfC3Chart', function() {
+describe('Component: pfC3Chart', function() {
   var $scope, $compile, element;
 
   beforeEach(module(
@@ -16,6 +16,11 @@ describe('Directive: pfC3Chart', function() {
     element = '<pf-c3-chart id="myChart" config="chartConfig"></pf-c3-chart>';
     element = $compile(element)($scope);
     $scope.$digest();
+  });
+
+  it("should have access to a global PatternFly object & method", function() {
+    expect(window.patternfly).toBeDefined();
+    expect(window.patternfly.c3ChartDefaults()).toBeDefined();
   });
 
   it("chart should find empty template", function() {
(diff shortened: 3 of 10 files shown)


eslint.yaml
```diff
@@ -35,7 +35,7 @@ rules:
   vars-on-top: 2
   no-debugger: 2
   no-cond-assign: 2
-  no-console: 2
+  no-console: 1
```
jeff-phillips-18: I would rather see instances were console must be used be eslint ignored than change this global setting

jeff-phillips-18:
I'm not sure I am in agreement with the premise of the PR. Angular patternfly charts should use the default settings from patternfly. I see that the README file does not specify that when using the charts package, one must include the patternfly settings javascript file which is necessary. Would it be a better option just to include that?

There are other places in the charts module that depend on the patternfly C3 defaults to be setup, like https://github.com/patternfly/angular-patternfly/blob/master/src/charts/sparkline/sparkline-chart.component.js#L189-L208

eslint.yaml
cdcabrera: Sure thing

cdcabrera:
Modifying the README totally agree. Adding in the PatternFly settings would definitely accomplish this in a more graceful modular way without an override. 

So based on that seems like this could go towards a couple of options
- Currently we have ```angular-patternfly.js``` and ```angular-patternfly.min.js```...
	- as @jeff-phillips-18  mentioned, we could bundle everything within the same files while also doing a check to see if the globals exist first, to avoid overwriting
	- or to avoid conflicts entirely we could create another distributed version, like ```angular-patternfly-[withsettings].js```
- And/Or we could just look at going back and possibly wrapping IFFEs with global checks, similar to what's currently in this PR for ```d3```, ```patternfly```, ```c3ChartDefaults``` around the components/directives that utilize them...

dtaylor113:
Hi, I asked a question in #534, I'll ask it here.  What problem is this PR trying to solve?  I don't think angular-patternfly can/should be run without the core patternfly c3 defaults.

jeff-phillips-18:
IMO, just an update to the README would do. It should be OK to depend on patternfly-settings.js since there are no further dependencies (jquery, lodash, anything...).

cdcabrera:
Cool on the Readme input, @jeff-phillips-18 and good question(s) @dtaylor113 ...

>What problem is this PR trying to solve? I don't think angular-patternfly can/should be run without the core patternfly c3 defaults.

The issue is a non-blocking ease of integration piece, aimed at first time integration and not necessarily requiring a project build like Grunt/Gulp or Webpack. 

Probably going to counter myself with this explanation ...

### Settings, Globals, and the Sans-Build

If we were to bundle the ```patternfly-settings.js``` into an additional JS dist file, that would allow a bit more flexibility for users integrating our framework enhancements without taking down the rest of AngularPf when someone either purposefully didn't need/want charting components, or inadvertently forgot the settings reference.

And agreed, including the Bower or NPM reference pulls in the required resources, also integrating AngularPf into a Webpack based app probably resolves any issue. This does highlight that AngularPf is build oriented. This opens towards "what about lower forms of adoption" by newer community members and possible CDN use. Where sometimes "[I just want to place a simple script tag, maybe a CSS reference in my HTML]". We have an opportunity, with minimal effort, to make it that way, either through this proposal, or another form of it, perhaps in the form of our ```dist``` containing a copy of the ```patternfly-settings.js```.

### Unit testing

Calling out our use of globals/frameworks/settings, in addition to what Karma Config contains (since that's not always the first things someone reads) within our unit tests potentially pushes us forward in documentation despite some minor redundancy. 

The counter to these unit test additions is, "where does this stop?" Do we also handle this in our use of all outside frameworks; lodash, Bootstrap, etc. or do we focus on the ones that potentially take down all of AngularPf instead of individual modules. It may not be as tedious of an update since we could also look at making sure something similar to ```beforeEach(module('lodash'));``` or whatever the module is, is defined.

cdcabrera:
Closing this down until can work through a more defined way of handling. Will re-open accordingly

PR #548, opened 2017-08-03
Fix notification drawer accordion panel sizing

### LINKED ISSUE #545: Notification drawer accordion does not work consistently
Was talking with @jeff-phillips-18 about some issues, he noted height calculation.  

I've been experiencing issues where items are truncated.  For example, I have 15 items in this list, but only 7 are visible & it will not scroll consistently:

screen shot 2017-08-03 at 12 21 19 am

jeff-phillips-18:
## Description
Update sizes when elements in the notification drawer are updated.
Update collapse children to not have a intermediate parent.
Fix for example notification drawer body html indentation error.

Fixes #545 

@benjaminapetersen

### DIFF
--- a/src/notification/notification-drawer.component.js
+++ b/src/notification/notification-drawer.component.js
@@ -51,6 +51,12 @@ angular.module('patternfly.notification').component('pfNotificationDrawer', {
       });
     };
 
+    var updateAccordionSizing = function () {
+      $timeout(function () {
+        angular.element($window).triggerHandler('resize');
+      }, 100);
+    };
+
     ctrl.toggleCollapse = function (selectedGroup) {
       if (selectedGroup.open) {
         selectedGroup.open = false;
@@ -59,6 +65,7 @@ angular.module('patternfly.notification').component('pfNotificationDrawer', {
           group.open = false;
         });
         selectedGroup.open = true;
+        updateAccordionSizing();
       }
     };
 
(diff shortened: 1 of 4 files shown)


src/utils/fixed-accordion.directive.js
```diff
@@ -78,24 +85,21 @@ angular.module('patternfly.utils').directive('pfFixedAccordion', function ($wind
 
           if (angular.isDefined($scope.scrollSelector)) {
             scroller = angular.element(collapsePanel[0].querySelector($scope.scrollSelector));
-            if (scroller.length === 1) {
+            if (scroller.length > 0) {
               scrollElement = angular.element(scroller[0]);
```
dtaylor113: Nit. could remove "> 0"

jeff-phillips-18: fixed

src/notification/notification-drawer.component.js
```diff
@@ -51,6 +51,12 @@ angular.module('patternfly.notification').component('pfNotificationDrawer', {
       });
     };
 
+    var updateAccordionSizing = function () {
+      $timeout(function () {
```
benjaminapetersen: Any concerns about performance here?  if `$onChanges` is triggered frequently, do you want to be triggering resize often, or would a `debounce`/`throttle` be in order?  I'm also curious about `100`, vs (`0` or any other small #) if its arbitrary or standardized across patternfly angular for this kind of thing.

src/notification/notification-drawer.html
```diff
@@ -19,16 +19,15 @@ <h4 class="panel-title">
           <span class="panel-counter" ng-include src="$ctrl.subheadingInclude"></span>
         </div>
         <div class="panel-collapse collapse" ng-class="{in: notificationGroup.open || $ctrl.notificationGroups.length === 1}">
```
cdcabrera: Nit: truthy check instead...

jeff-phillips-18: huh?

src/notification/notification-drawer.component.js
jeff-phillips-18: I don't believe onChanges will get call that frequently. Typically most of these won't change, but should they, the size needs to be recalculated. Also, its only done if the draw is shown and one of these values change.

We've typically used 100 just to get past any follow up changes and get thru a digest cycle or two. Main idea is to get past the current digest cycle anyhow.

src/notification/notification-drawer.html
jeff-phillips-18: But it must be exactly 1.

benjaminapetersen: Right, this is the single panel option. Sorry, just saw the truthy "huh" :)

benjaminapetersen:
@jeff-phillips-18 I've built this & pulled it into the console to test.  Works much better, big improvement, thx!

PR #557, opened 2017-08-08
Restores Build Package Update from Release

cdcabrera:
## Description
Corrects current build fail due to incorrect "-eng" repo version... switches back from an exact NPM package version to a ">=" version.

## PR Checklist

- [n/a] Unit tests are included
- [n/a] Screenshots are attached (if there are visual changes in the UI)
- [n/a] A Designer is assigned as a reviewer (if there are visual changes in the UI)
- [n/a] A CSS rep is assigned as a reviewer (if there are visual changes in the UI)

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "karma-phantomjs-launcher": "^1.0.0",
     "matchdep": "0.3.0",
     "nsp": "^2.6.1",
-    "patternfly-eng-release": "3.26.2"
+    "patternfly-eng-release": "3.26.6"
   },
   "optionalDependencies": {
     "angularjs-datatables": "^0.5.7",


cdcabrera:
Dunno, I remember it but now the change just isn't there

jeff-phillips-18:
It got updated by patternfly-build. Will this re-occur?

cdcabrera:
per @jeff-phillips-18 
> git blame... https://github.com/patternfly/angular-patternfly/commit/4ccc15bebec4f77e03422eb83de3d0c817798d8e
>  Will this re-occur?

Indeed, this will probably happen again since the release scripts are trying to increment the version and we're customizing it... might need an exception where if it sees anything in addition to version number it ignores it. 

That being said this may be a non-issue, little bit more testing. 

Based on everything this PR, as is, provides an immediate fix, not something long-term.

PR #559, opened 2017-08-09
Remove early ‘semantic release pre’ step from Travis build

dlabrecq:
This step was in place so that we could add the generated version number from package.json to bower.json. We just need to get the semantic release going for now, then revisit bower.json if necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,6 @@ before_install:
 install: true
 
 script:
-  - if [[ -n "$NPM_TOKEN" && -n "$GH_TOKEN" ]]; then npm run semantic-release-pre; fi
   - sh -x ./node_modules/patternfly-eng-release/scripts/_build.sh -a
 
 after_success:
--- a/package.json
+++ b/package.json
@@ -88,8 +88,7 @@
     "help": "grunt help",
     "serve": "grunt serve",
     "start": "grunt serve",
-    "semantic-release": "npm publish && semantic-release post",
-    "semantic-release-pre": "semantic-release pre"
+    "semantic-release": "semantic-release pre && npm publish && semantic-release post"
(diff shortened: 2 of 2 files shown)


dlabrecq:
There are a couple things that the semantic release won't do. For example, it won't bump the version number in bower.json.

I wanted to call "semantic-release pre" to generate the version number in package.json first, so I could take that and bump bower.json via a script. 

I'm not sure if calling "semantic-release pre" is why Travis failed to build, but thought I should focus on getting the release going for now. Then, revisit bower.json if we need to bump that number, too.

This may take an iteration or two to get right. The problem is we cannot test this fully without actually generating a release.

jeff-phillips-18:
Not updating bower.json is likely a show stopper. It will lead to confusion for projects pulling angular-patternfly. I know I often look at packages bower.json file to verify the version.

Can we clone the repo and test releasing from there?

dlabrecq:
The semantic release creates a Github tag, which bower uses to install from. There is no change there, so bower will install normally.

The only potential issue is that the version in bower.json will read "0.0.0-semantically-released", which is required by the semantic release command.

jeff-phillips-18:
And IMO that is an issue. The version in bower.json should be accurate.

dlabrecq:
I have not been able to confirm it is an issue, just something I mean to watch. If it is a problem, I planed to simply move the 'semantic pre' command earlier in Travis -- just trying to get the build going for this repo right now -- patternfly-eng-release and patternfly-ng are already up and running.

That said, I can create another PR to disable the semantic release? However, I don't have much more time I can spend on this -- setup semantic release with npm/github tokens, setup Travis, publish to a fake npm package, etc. for yet another repo.

PR #560, opened 2017-08-09
semantic release: publish webjar

dlabrecq:
Added a patternfly-eng-release script to publish web jar after semantic release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,6 @@ before_install:
 install: true
 
 script:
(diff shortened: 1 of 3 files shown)


jeff-phillips-18:
How do we get bower.json to be updated in the release process? I don't want to see us release with the wrong version in bower.json. I believe that was removed in #559

dlabrecq:
My plan was to re-evaluate that after the semantic release runs successfully. If it turns out that the bower.json file is not updated by the semantic release, then I will move the 'semantic pre' command earlier in the Travis build.

We can always delete the generated tag associated with a release.

jeff-phillips-18:
OK @dlabrecq I can buy that.

PR #570, opened 2017-08-12
Utilization Donut Chart Component

cdcabrera:
## Description
Implementation for utilization donut chart component. 

Create a CPU Utilization Component, which is not embedded in a Card:
- color changes based on thresholds
- optional details are shown
- optional details can be stacked horizontally, after the chart
- optional details can be stacked vertically 

## In Progress Checklist
- [x] Unit tests
- [x] CSS styling for labels
- [x] Clean up and line endings
- [x] Evaluate focusing on use of a C3 chart config instead of individual attributes, similar to existing chart components
- [x] Evaluate transcluded label instead of "parameterizing" and "concatenating" them

## PR Checklist

- [x] Unit tests are included
- [x] A Designer is assigned as a reviewer (if there are visual changes in the UI)
- [x] A CSS rep is assigned as a reviewer (if there are visual changes in the UI)
- [x] Amend commit for continuous integration 

### Updated
screen shot 2017-08-14 at 4 13 55 pm

### Previous 
screen shot 2017-08-12 at 1 26 38 am

### DIFF
--- a/src/charts/donut/donut-pct-chart.html
+++ b/src/charts/donut/donut-pct-chart.html
@@ -1,5 +1,22 @@
-<span>
-  <pf-c3-chart ng-if="$ctrl.data.dataAvailable !== false" id="{{$ctrl.donutChartId}}" config="$ctrl.config"
-               get-chart-callback="$ctrl.setChart"></pf-c3-chart>
-  <pf-empty-chart ng-if="$ctrl.data.dataAvailable === false" chart-height="$ctrl.chartHeight"></pf-empty-chart>
+<span class="pct-donut-chart-pf">
+  <span ng-class="{'pct-donut-chart-pf-left': $ctrl.config.labelConfig.orientation === 'left', 'pct-donut-chart-pf-right': $ctrl.config.labelConfig.orientation === 'right'}">
+    <span class="pct-donut-chart-pf-chart">
+      <pf-c3-chart ng-if="$ctrl.data.dataAvailable !== false" id="{{$ctrl.donutChartId}}" config="$ctrl.config"
+                   get-chart-callback="$ctrl.setChart"></pf-c3-chart>
+      <pf-empty-chart ng-if="$ctrl.data.dataAvailable === false" chart-height="$ctrl.chartHeight"></pf-empty-chart>
+    </span>
+    <span ng-if="$ctrl.data.dataAvailable !== false && $ctrl.config.labelConfig && !$ctrl.config.labelConfig.labelFn()" class="pct-donut-chart-pf-label">
+      {{$ctrl.config.labelConfig.title}}
+      <span ng-if="$ctrl.data" ng-switch="$ctrl.config.labelConfig.label">
+        <span ng-switch-when="none"></span>
+        <span ng-switch-when="available">{{$ctrl.data.available}} {{$ctrl.config.labelConfig.units}} available</span>
+        <span ng-switch-when="percent">{{$ctrl.data.percent}}&#37; used</span>
+        <span ng-switch-default="">{{$ctrl.data.used}} {{$ctrl.config.labelConfig.units}} of {{$ctrl.data.total}} {{$ctrl.config.labelConfig.units}} used</span>
+      </span>
+    </span>
+    <span ng-if="$ctrl.data.dataAvailable !== false && $ctrl.config.labelConfig && $ctrl.config.labelConfig.labelFn()" class="pct-donut-chart-pf-label"
+          ng-bind-html="$ctrl.config.labelConfig.labelFn()">
+    </span>
+  </span>
 </span>
+
--- a/src/charts/donut/examples/donut-pct-chart.js
+++ b/src/charts/donut/examples/donut-pct-chart.js
@@ -26,11 +26,21 @@
  * <li>.tooltipFn(d)   - user defined function to customize the tool tip (optional)
  * <li>.centerLabelFn  - user defined function to customize the text of the center label (optional)
  * <li>.onClickFn(d,i) - user defined function to handle when donut arc is clicked upon.
+ * <li>.labelConfig    - object containing properties for external label (optional) - default: undefined
+ *   <ul>
+ *       <li>.orientation - string with possible values: 'left', 'right' (optional) - default: 'center'
+ *       <li>.title       - string representing a prefix or title (optional) - default: empty string
+ *       <li>.label       - the wording format to display, possible values: 'used', 'available', 'percent', 'none' (optional) - default: 'used'
+ *       <li>.units       - unit label for values, ex: 'MHz','GB', etc.. (optional) - default: empty string
+ *       <li>.labelFn     - function to customize the text of the external label. This callback returns no data. Updated display data can be accessed through the passed and updated parameter 'data'. (optional) - default: undefined
+ *   </ul>
+ * </li>
  * </ul>
  *
  * @param {object} data the Total and Used values for the donut chart.  Available is calculated as Total - Used.<br/>
  * <ul style='list-style-type: none'>
  * <li>.used          - number representing the amount used
+ * <li>.percent       - number representing the percentage used
  * <li>.total         - number representing the total amount
  * <li>.dataAvailable - Flag if there is data available - default: true
  * </ul>
@@ -62,19 +72,23 @@
          <div class="row">
            <div class="col-md-3 text-center">
              <label>Error Threshold</label>
-             <pf-donut-pct-chart config="configErr" data="dataErr" chart="chartErr"></pf-donut-pct-chart>
+             <p class="text-right">
+               <pf-donut-pct-chart config="configErr" data="dataErr" chart="chartErr"></pf-donut-pct-chart>
+             </p>
            </div>
-           <div class="col-md-3 text-center"">
+           <div class="col-md-3 text-center">
              <label>Warning Threshold</label>
              <pf-donut-pct-chart config="configWarn" data="dataWarn"></pf-donut-pct-chart>
            </div>
-           <div class="col-md-3 text-center"">
+           <div class="col-md-3 text-center">
              <label class="camelcase">{{threshLabel}} Threshold</label>
-             <pf-donut-pct-chart config="configDynamic" data="dataDynamic" center-label="labelDynamic"
-                                 on-threshold-change="thresholdChanged(threshold)">
-             </pf-donut-pct-chart>
+             <p class="text-left">
+               <pf-donut-pct-chart config="configDynamic" data="dataDynamic" center-label="labelDynamic"
+                                   on-threshold-change="thresholdChanged(threshold)">
+               </pf-donut-pct-chart>
+             </p>
            </div>
-           <div class="col-md-3 text-center"">
+           <div class="col-md-3 text-center">
              <label>No Threshold</label>
              <pf-donut-pct-chart config="configNoThresh" data="dataNoThresh"></pf-donut-pct-chart>
            </div>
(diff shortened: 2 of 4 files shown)


dtaylor113:
Hi David, how is this component different from the current donut chart?   It seems very similar, just wondering why we could enhance the already existing component?
Thanks,
Dave

cdcabrera:
@Dtaylor that's correct most of this component is a pass through directly for the existing donut charts. 

Funny enough have an initial version that just adds a couple of additional attributes on top of the pct donut but the sizing was awkward. 

Overall the difference are around label placement, minor sizing, and center label customizing... all things that a user can potentially do with the config, but with short-cuts/settings/properties instead. 

Sizing and placement were/are the more bothersome properties to deal with and one of the reasons for separating it out. Now that understand how sizing functions I can look at what converting this version would take, since it sounds like we're good with the add on approach.

cdcabrera:
@dtaylor113 have part of your input, @jeff-phillips-18 do you have input on...

- Adding to the existing pct-donunt chart component?

And both...

- The idea that a large percentage of this component, as is, is just structuring the "config" in a customized way, and that perhaps just placing an updated demo might be enough? 
  Here's a quick snapshot of the intention behind it... 
  cpuutilization-horizontallystacked

serenamarie125:
@cdcabrera agree with the approach of updating the existing donut chart rather than create a new component

cdcabrera:
good good @serenamarie125 , @beanh66 @cshinn
- aspects of the initial design layout can start to be viewed, used flex to layout the columns. There's an initial screen shot, which when compared to the request is missing some bold and sized text

@beanh66 @cshinn your input on
- adding to the existing component vs writing a new one, that's really really close to the old one.

dtaylor113:
@cdcabrera, you are right that we already have an attribute called 'center-label' which can have values 'used', 'available', 'percent', and 'none', so it sounds like setting it to 'percent' will correctly set the center label.  The issue is how can we bundle the 'side labels' so that it's easy for app. dev to customize/specify.
I would suggest adding a config option, something like 'side-label': {
position: 'left' / 'right',
title: 'Storage',    
type: percent, used, available  (?),
format: full ('of m GB used'), idk, just spitballing here :-)
}

jeff-phillips-18:
@cdcabrera I agree with enhancing the existing component.

cdcabrera:
@amarie401 for reference

jeff-phillips-18:
Is this still WIP or ready for final reviews?

cdcabrera:
@jeff-phillips-18 et all, it's close to final can remove the WIP just...
- need some for/against input on how the additional config option is structured on top of the existing donut
- any minor design concerns

Then i'll rebase and amend the commit message

src/charts/charts.less
```diff
@@ -11,6 +11,32 @@ pf-c3-chart {
...
+    flex-direction: row;
+    justify-content: center;
+    align-items: center;
+  }
+
+  .pct-donut-chart-pf-left {
+    flex-direction: row-reverse;
+  }
+}
+
```
jeff-phillips-18: Should this really be put into Patternfly? Is this a Patternfly pattern?

src/charts/donut/examples/donut-pct-chart.js
```diff
@@ -26,6 +26,15 @@
  * <li>.tooltipFn(d)   - user defined function to customize the tool tip (optional)
  * <li>.centerLabelFn  - user defined function to customize the text of the center label (optional)
  * <li>.onClickFn(d,i) - user defined function to handle when donut arc is clicked upon.
+ * <li>.label          - object containing properties for external label (optional)
+ *   <ul>
+ *       <li>.orientation - string with possible values: 'left', 'right' (optional)
```
jeff-phillips-18: This is optional, what is the default?

src/charts/donut/examples/donut-pct-chart.js
```diff
@@ -26,6 +26,15 @@
...
+ *       <li>.title       - string representing a prefix or title (optional)
+ *       <li>.label       - similar in to the center-label parameter, which specifies the contents of the donut's external label, values: 'used', 'available', 'percent', 'none' (optional)
```
jeff-phillips-18: I don't understand the description here.

src/charts/donut/examples/donut-pct-chart.js
```diff
@@ -62,17 +71,21 @@
...
-             <pf-donut-pct-chart config="configDynamic" data="dataDynamic" center-label="labelDynamic"
-                                 on-threshold-change="thresholdChanged(threshold)">
-             </pf-donut-pct-chart>
+             <p class="text-left">
+               <pf-donut-pct-chart config="configDynamic" data="dataDynamic" center-label="labelDynamic"
+                                   on-threshold-change="thresholdChanged(threshold)">
+               </pf-donut-pct-chart>
+             </p>
            </div>
            <div class="col-md-3 text-center"">
```
jeff-phillips-18: Not this PR but coul you remove the extra "  character.

src/charts/donut/examples/donut-pct-chart.js
```diff
@@ -26,6 +26,15 @@
  * <li>.tooltipFn(d)   - user defined function to customize the tool tip (optional)
  * <li>.centerLabelFn  - user defined function to customize the text of the center label (optional)
  * <li>.onClickFn(d,i) - user defined function to handle when donut arc is clicked upon.
+ * <li>.label          - object containing properties for external label (optional)
```
jeff-phillips-18: Maybe this should be labelConfig or something more descriptive so that it is understood that this is not just a simple label.

dtaylor113: I had suggested ```sideLabel``` with a left and right attribute as well.

dtaylor113: Hi, I don't think ```.label``` should be under config, rather it should be a separate param like ```center-label``` is, and called ```side-label``` with properties orientation, title, and perhaps something like 'amount-type' or 'units' for available, percent, used.

src/charts/donut/examples/donut-pct-chart.js
cdcabrera: Yeah the wording is a little funky reading back on it, can correct it...

src/charts/donut/examples/donut-pct-chart.js
cdcabrera: Good catch can update the comments/annotation. For our purposes, the optional default, is it simply doesn't exist, don't have to use it

src/charts/charts.less
cdcabrera: @jeff-phillips-18  can move it over to core for sure np. There are aspects of charts both within this repository's less and core, just need a direction we want to head?

src/charts/donut/examples/donut-pct-chart.js
cdcabrera: Took using ```side-label``` into consideration, but when the label can be used on the right, left, and bottom it becomes a slight misnomer, so changed it to ```label```. 

Was also going to create it as a separate attribute similar to ```center-label```, but then we blend the use of how ```center-label``` functions. We have both a configuration aspect, which takes a function, and the attribute, which takes a string. Based on that took the configuration route all in one, instead of a blended approach, which for myself felt confusing. 

I do agree the ```label``` label should probably be changed to something else, I like ```labelConfig```, had also thought maybe ```outsideLabel```, but the ```Config``` suffix seems more appropriate. 

As a quick fix, I can go ahead and change the label to ```labelConfig``` while we continue to discuss.

src/charts/donut/donut-pct-chart.html
```diff
@@ -1,5 +1,22 @@
...
+<span class="pct-donut-chart-pf">
+  <span ng-class="{'pct-donut-chart-pf-block': $ctrl.config.label.orientation !== 'left' && $ctrl.config.label.orientation !== 'right', 'pct-donut-chart-pf-left': $ctrl.config.label.orientation === 'left', 'pct-donut-chart-pf-right': $ctrl.config.label.orientation === 'right'}">
+    <span class="pct-donut-chart-pf-chart">
+      <pf-c3-chart ng-if="$ctrl.data.dataAvailable !== false" id="{{$ctrl.donutChartId}}" config="$ctrl.config"
+                   get-chart-callback="$ctrl.setChart"></pf-c3-chart>
+      <pf-empty-chart ng-if="$ctrl.data.dataAvailable === false" chart-height="$ctrl.chartHeight"></pf-empty-chart>
+    </span>
+    <span ng-if="$ctrl.data.dataAvailable !== false && $ctrl.config.label && !$ctrl.config.label.labelFn()" class="pct-donut-chart-pf-label">
+      {{$ctrl.config.label.title}}
+      <span ng-if="$ctrl.data" ng-switch="$ctrl.config.label.label">
```
dtaylor113: ```$ctrl.config.label.label``` I think we do need 'labelConfig' or something clearer.  Currently, 'label.label' is tough to read.

src/charts/charts.less
cdcabrera: After looking a little bit more on this, looks like we have a choice. The appearance is, it looks like we attempted to start a PF core aspect to the charts, here:
- https://github.com/cdcabrera/patternfly/blob/master/src/less/charts.less
- https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/donut-charts.html

With a distinction about only how we modify SVG aspect of the Charts within AngularPF. However it looks like we slightly diverged a little and started placing a few more bits inside AngularPF instead of core. 

I can go back and move this newer piece over to core now, or hold off based on what's already in AngularPf? But in addition, outside of this PR, we probably need to go through the rest of charts.less and separate them out a bit more.

cdcabrera:
@dtaylor113 @jeff-phillips-18 since both commented on the Less, I'll go ahead and move it over to PF core for this update. If there are any additional PCT Donut styles that don't fall into the SVG/C3 manipulate I can try to move those as well.

src/charts/charts.less
cdcabrera: @jeff-phillips-18 @dtaylor113  the less has been moved over to core Patternfly PR  720

cdcabrera:
Went ahead and [separated out the Pct Donut Less towards core Pf, see this PR 720](https://github.com/patternfly/patternfly/pull/720)

src/charts/charts.less
```diff
@@ -11,6 +11,7 @@ pf-c3-chart {
     width: 100%;
   }
 }
+
```
jeff-phillips-18: Looks like this file's changes could be discarded. No point updating to include an empty line 😉

src/charts/donut/examples/donut-pct-chart.js
```diff
@@ -26,6 +26,15 @@
  * <li>.tooltipFn(d)   - user defined function to customize the tool tip (optional)
  * <li>.centerLabelFn  - user defined function to customize the text of the center label (optional)
  * <li>.onClickFn(d,i) - user defined function to handle when donut arc is clicked upon.
+ * <li>.labelConfig    - object containing properties for external label (optional) - default: undefined
+ *   <ul>
+ *       <li>.orientation - string with possible values: 'left', 'right' (optional) - default: 'center'
+ *       <li>.title       - string representing a prefix or title (optional) - default: empty string
+ *       <li>.label       - the wording format to display, possible values: 'used', 'available', 'percent', 'none' (optional) - default: 'used'
+ *       <li>.units       - unit label for values, ex: 'MHz','GB', etc.. (optional) - default: empty string
+ *       <li>.labelFn     - function to customize the text of the external label (optional) - default: undefined
```
jeff-phillips-18: Should this function take some parameters? How is the function to know what the label ought to be?

dtaylor113:
Hi @cdcabrera, I feel like the ngDoc has somewhat gotten somewhat munged (if that is a word :-).  Previously, the top row was showing the different Thresholds, the middle row was the different center-labels, the last row was some custom stuff:
image

With your current changes, the first row is now showing Thresholds and some custom label placements, and I think different donut sizes(?).  However, there is no ngDoc help text or labels to denote what is being displayed in the first row:

image

I would suggest leaving the first row to denote the different Thresholds and add a new row which clearly describes the different/new label customizations and placements you are introducing.
Thanks,
Dave

cdcabrera:
@dtaylor113 can add a dedicated row for sure

cdcabrera:
This PR is dependent on [core Patternfly PR 720](https://github.com/patternfly/patternfly/pull/720)

src/charts/donut/examples/donut-pct-chart.js
cdcabrera: Can add a note/annotation on there.

The behavior is similar to the existing ```centerLabelFn``` callback. That one also seems to be lacking in the documentation/annotation department

cdcabrera:
@dtaylor113 the revised layout linked to Patternfly core PR 720

screen shot 2017-08-28 at 5 00 40 pm

cdcabrera:
The associated core PatternFly PR 720 has been merged, you'll need to update your packages to see the CSS update

dtaylor113:
@cdcabrera Perfect!  Exactly what I was envisioning.

PR #572, opened 2017-08-15
Fix #571 - pfTableView: ng-include inside ng-repeat hurts performance

### LINKED ISSUE #571: pfTableView: ng-include inside ng-repeat hurts performance
The following `ng-include` is causing a big performance hit in the table view:
https://github.com/patternfly/angular-patternfly/blob/master/src/table/tableview/table-view.html#L22

See blog posts about the issue:
https://www.bennadel.com/blog/2738-using-ngrepeat-with-nginclude-hurts-performance-in-angularjs.htm
http://www.codelord.net/2016/11/17/avoiding-ng-include-for-elegenace-and-performance/
http://eric.stolten.net/posts/2015-04-10-prefer-directives-to-ng-include/index.html

I'll try to fix it.

akieling:
## Add new configuration property to `columns` object to define a template function

The goal of this PR is to provide a fix to performance issues caused by the use of ng-include inside ng-repeat to parse and include the content of each table cell that has a template.
This change is backward compatible and doesn't affect existing code that depends on the `htmlTemplate` property of the `columns` object. This change adds a new property called `tempateFn` as a potential replacement to `htmlTemplate`. 

> **Note:** This new property is light and simple but doesn't support AngularJS directives in the template, therefore it doesn't support things like ng-click.

List of changes:
- Add new `templateFn` property to documentation
- Inject `$sce` into controller to tell AngularJS to trust the template HTML
- Remove `hasHTMLTemplate` and `getHTMLTemplate` functions from controller because we can check whether the columns have a template by just checking the existence of the `htmlTemplate` and `templateFn` properties
- Use `ng-bind-html` to invoke the template function and inject the results in the cells
- Add a new unit test for this feature

## PR Checklist

- [ X ] Unit tests are included
- [ ] Screenshots are attached (if there are visual changes in the UI)

### DIFF
--- a/src/table/tableview/examples/table-view-basic.js
+++ b/src/table/tableview/examples/table-view-basic.js
@@ -20,7 +20,8 @@
   * <ul style='list-style-type: none'>
   *   <li>.header     - (string) Text label for a column header
   *   <li>.itemField    - (string) Item field to associate with a particular column.
-  *   <li>.htmlTemplate - (string) (optional) id/name of an embedded ng/html template. Ex: htmlTemplate="name_template.html".  The template will be used to render each cell of the column.
+  *   <li>.templateFn - (function) (optional) Template function used to render each cell of the column. Pro: more performant than `htmlTemplate`. Con: doesn't support AngularJS directives in the template, therefore it doesn't support things like ng-click. Example: <pre>templateFn: value => `<span class="text-danger">${value}</span>`</pre>
+  *   <li>.htmlTemplate - (string) (optional) id/name of an embedded ng/html template. Pro: supports AngularJS directives in the template. Con: poor performance on large tables. Ex: htmlTemplate="name_template.html".  The template will be used to render each cell of the column.
   *        Use <code>handleColAction(key, value)</code> in the template to call the <code>colActionFn</code> callback function you specify. 'key' is the item attribute name; which should equal the itemFld of a column.
   *       'value' is the item[key] value.
   *   <pre>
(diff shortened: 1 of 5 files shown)


src/table/tableview/examples/table-view-basic.js
```diff
@@ -20,6 +20,7 @@
   * <ul style='list-style-type: none'>
   *   <li>.header     - (string) Text label for a column header
   *   <li>.itemField    - (string) Item field to associate with a particular column.
+  *   <li>.templateFn - (function) (optional) Template function used to render each cell of the column. Example: <pre>templateFn: value => `<span class="text-danger">${value}</span>`</pre>
```
dtaylor113: Please add some info. on the pros and cons of using templateFn as opposed to htmlTemplate.  Pro: Is more performant than ```htmlTemplate```.  Con: Doesn't support AngularJS directives in the template, therefore it doesn't support things like ng-click.

src/table/tableview/examples/table-view-basic.js
```diff
@@ -20,6 +20,7 @@
...
   *   <li>.htmlTemplate - (string) (optional) id/name of an embedded ng/html template. Ex: htmlTemplate="name_template.html".  The template will be used to render each cell of the column.
```
dtaylor113: Same comment as above.  Please include some info as to why an app dev might choose htmlTemplate over templateFn, or visa-versa.  -thanks

dtaylor113:
Thanks @akieling , LGTM :+1:

Issue #542, opened 2017-08-01
Customizing chart transitions/animations in angular-patternfly version > 4.*


aptmac:
Hi,

After updating to angular-pf 4 from 3 I've noticed that my charts (donut and line specifically) have instant transitions and don't retain the smooth animations that they had in version 3. I used to be able to add a 'transition' attribute to the chart config, but this no longer works. An example is in the API for [donut-pct-chart](http://www.patternfly.org/angular-patternfly/#/api/patternfly.charts.component:pfDonutPctChart), the updating 'Ok/Warning/Error Threshold' animations are instant, whereas before I was able to customize the animation time to look similar to [this](http://jsfiddle.net/prewiseDevelopers/uoLfv79b/). Any recommendations?

Cheers,

aptmac

aptmac:
I'm working on a fix for this.

aptmac:
Fixed in https://github.com/patternfly/angular-patternfly/pull/578

Issue #541, opened 2017-08-01
Adding data points to a chart causes the page to automatically scroll upwards


aptmac:
**Issue:**

When a chart refreshes after updating/adding a new point to a chart, the page automatically scrolls upwards. 

**To reproduce:**
I found several instances of this behaviour when browsing the [angular-patternfly documentation](http://www.patternfly.org/angular-patternfly/#/api), including:

1. [donut chart](http://www.patternfly.org/angular-patternfly/#/api/patternfly.charts.component:pfDonutPctChart): medium screen size and smaller, scroll to the bottom of the page and when the "Ok Threshold" donut chart updates, the screen will scroll upwards.
2. [line chart](http://www.patternfly.org/angular-patternfly/#/api/patternfly.charts.component:pfLineChart) & [sparkline chart](http://www.patternfly.org/angular-patternfly/#/api/patternfly.charts.component:pfSparklineChart): medium screen size or smaller, scroll to the bottom of the page and click the "Add Data Point" button; the screen should scroll up a small amount.
3. [trends chart](http://www.patternfly.org/angular-patternfly/#/api/patternfly.charts.directive:pfTrendsChart): all screen sizes, scroll to the bottom of the page and click the "Add Data Point" button; the screen should scroll up a small amount.

I originally saw this behaviour with some charts I'm currently working on, here are two gifs that show how the chart behaved in 3.23.3 (top) and in the latest release. 4.2.0 (bottom).

3.23.3:
![3.23.3](https://user-images.githubusercontent.com/10425301/28844800-da883b1a-76d3-11e7-8ddf-9fab1c7d694f.gif)

4.2.0 (in this gif I'm scrolling to the bottom of the page after each chart refresh):
![4.2.0](https://user-images.githubusercontent.com/10425301/28844793-d33c666a-76d3-11e7-9001-487f0fe5c282.gif)

aptmac:
I'm working on a fix for this.

beanh66:
Thanks @aptmac! Would you mind assigning yourself to this issue as long as you are working on it, just so nobody else picks it up?

aptmac:
Fixed in https://github.com/patternfly/angular-patternfly/pull/578

PR #578, opened 2017-08-17
Fix transitions in c3-charts and non-threshold donutPctCharts

aptmac:
## Description
This PR addresses issues #541 and #542.

The first short commit corrects #541, where updating charts would cause a re-generation of the chart, and as a result, potentially shift screen orientation. This was caused by 
`if (!chart || $attrs.id.indexOf('donutPctChart')) {`
because indexOf() returns a number, and will not cause the condition to fail if the result is '-1'. As a result, charts that are not donutPctCharts (e.g., line, sparkline, and trends) will pass this condition and re-generate. My proposed solution here is to change indexOf() to includes(), which will return a boolean instead, and ensure intended behaviour.

The second short commit addresses #542, and fixes the 'transition' behaviour in donutPctCharts that do not contain a 'thresholds' attribute. Not all donutPctCharts will use thresholds, and only those that do will need to be re-generated to change colours. The transition attribute gets nullified by the chart being re-generated, because instead of using an animation when updating the chart will just re-draw itself to it's new value. My proposed solution is to add a check to the condition which makes the chart re-generate only if:
1. the chart doesn't exist, OR
2. the chart is a donutPctChart AND contains a threshold object
This doesn't fix transitions in thresholds utilizing donutPctCharts, but will fix the behaviour in charts that don't use thresholds.

Thoughts?

Cheers,
aptmac

## PR Checklist

- [ ] Unit tests are included
- [ ] Screenshots are attached (if there are visual changes in the UI)

### DIFF
--- a/src/charts/c3/c3-chart.component.js
+++ b/src/charts/c3/c3-chart.component.js
@@ -97,9 +97,9 @@
           chartData = ctrl.config;
           if (chartData) {
             chartData.bindto = '#' + $attrs.id;
-            // always re-generate donut pct chart because it's colors
-            // change based on data and thresholds
-            if (!chart || $attrs.id.includes('donutPctChart')) {
+            // only re-generate donut pct chart if it has a threshold object
+            // because it's colors will change based on data and thresholds
+            if (!chart || ($attrs.id.includes('donutPctChart') && chartData.thresholds)) {
               chart = c3.generate(chartData);
             } else {
               //if chart is already created, then we only need to re-load data
--- a/src/charts/c3/c3-chart.component.js
+++ b/src/charts/c3/c3-chart.component.js
@@ -99,7 +99,7 @@
             chartData.bindto = '#' + $attrs.id;
             // only re-generate donut pct chart if it has a threshold object
             // because it's colors will change based on data and thresholds
-            if (!chart || ($attrs.id.includes('donutPctChart') && chartData.thresholds)) {
+            if (!chart || ($attrs.id.indexOf('donutPctChart') !== -1 && chartData.thresholds)) {
               chart = c3.generate(chartData);
             } else {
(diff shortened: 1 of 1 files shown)


src/charts/c3/c3-chart.component.js
```diff
@@ -97,9 +97,9 @@
           chartData = ctrl.config;
           if (chartData) {
             chartData.bindto = '#' + $attrs.id;
-            // always re-generate donut pct chart because it's colors
-            // change based on data and thresholds
-            if (!chart || $attrs.id.indexOf('donutPctChart')) {
+            // only re-generate donut pct chart if it has a threshold object
+            // because it's colors will change based on data and thresholds
+            if (!chart || ($attrs.id.includes('donutPctChart') && chartData.thresholds)) {
```
cdcabrera: Love to use this bit of newer ECMA, but the IE support appears like it's not there... an alternative, you could change it to use ```test``` and accomplish something similar...
```javascript
if (!chart || ( /donutPctChart/.test($attrs.id) && chartData.thresholds)) ...
```

dtaylor113: or if the goal is to return a boolean from indexOf: ```$attrs.id.indexOf('donutPctChart') !== -1```

aptmac: Sure, I have reversed the usage of 'includes', and will instead will use the indexOf() !== -1.

For reference, is there a particular reason why indexOf() would be more advantageous than includes()?

dtaylor113: ```.includes()``` is not supported in all browsers, especially older ones.

aptmac: Ah that's good to know, thanks!

PR #587, opened 2017-08-22
Fixes issue 575: change h1 to h4 styled as h1

benjaminapetersen:
- Update `h1` to `h4.h1`
- Update id's to be unique via `id="empty-state-title-{{$id}}"`
- Update tests
- Fix typos

@jeff-phillips-18 for review, might be some conventions I should conform to.
@spadgett 

Re: update needed for [origin-web-console PR 1326](https://github.com/openshift/origin-web-console/pull/1326) via [this comment](https://github.com/openshift/origin-web-console/pull/1326#issuecomment-323076310).

### DIFF
--- a/src/views/empty-state.html
+++ b/src/views/empty-state.html
@@ -2,13 +2,13 @@
   <div ng-if="$ctrl.config.icon" class="blank-slate-pf-icon">
     <span class="{{$ctrl.config.icon}}"></span>
   </div>
-  <h1 id="title">
+  <h4 id="blank-state-pf-title-{{$id}}" class="h1 blank-state-pf-title">
     {{$ctrl.config.title}}
-  </h1>
-  <p id="info" ng-if="$ctrl.config.info">
+  </h4>
+  <p id="blank-state-pf-info-{{$id}}" class="blank-state-pf-info" ng-if="$ctrl.config.info">
     {{$ctrl.config.info}}
   </p>
-  <p id="helpLink" ng-if="$ctrl.config.helpLink">
+  <p id="blank-state-pf-helpLink-{{$id}}" class="blank-state-pf-helpLink" ng-if="$ctrl.config.helpLink">
     {{$ctrl.config.helpLink.label}} <a href="{{$ctrl.config.helpLink.url}}">{{$ctrl.config.helpLink.urlLabel}}</a>.
   </p>
   <div ng-if="$ctrl.hasMainActions()" class="blank-slate-pf-main-action">
--- a/test/notification/notification-drawer.spec.js
+++ b/test/notification/notification-drawer.spec.js
@@ -593,7 +593,7 @@ describe('Component:  pfNotificationDrawer', function () {
     var emptyStates = element.find('.blank-slate-pf');
     expect(emptyStates.length).toBe(1);
 
-    var title = angular.element(emptyStates[0]).find('#title').html();
+    var title = angular.element(emptyStates[0]).find('.blank-state-pf-title').html();
     expect(_.trim(title)).toBe('There are no notifications to display.');
   });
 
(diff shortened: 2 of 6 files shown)


src/views/empty-state.html
```diff
@@ -2,13 +2,13 @@
   <div ng-if="$ctrl.config.icon" class="blank-slate-pf-icon">
     <span class="{{$ctrl.config.icon}}"></span>
   </div>
-  <h1 id="title">
+  <h4 id="empty-state-title-{{$id}}" class="h1 title">
```
dtaylor113: Hi, it's a little confusing having a <h4> with a class of 'h1 title'?

benjaminapetersen: `h1` is just [the bootstrap way](https://getbootstrap.com/docs/3.3/css/#type-headings) to style a different heading as an `h1`, thats the only reason i kept it.  I don't know that we need to keep the class `.title` either, i just kept it so that the test updates are changes from `#title` to `.title`.  Def open to any suggestions, was just going for the most minimal changes to make this work!

jeff-phillips-18: Can we avoid using title as a class and just use this new id as a selector in the tests?

src/views/empty-state.html
```diff
@@ -2,13 +2,13 @@
...
     {{$ctrl.config.title}}
-  </h1>
-  <p id="info" ng-if="$ctrl.config.info">
+  </h4>
+  <p id="empty-state-info-{{$id}}" class="info" ng-if="$ctrl.config.info">
```
jeff-phillips-18: Can we use this new ID in the tests rather than adding info as a class?

src/views/empty-state.html
```diff
@@ -2,13 +2,13 @@
...
     {{$ctrl.config.info}}
   </p>
-  <p id="helpLink" ng-if="$ctrl.config.helpLink">
+  <p id="empty-state-helpLink-{{$id}}" class="helpLink" ng-if="$ctrl.config.helpLink">
```
jeff-phillips-18: Again...

jeff-phillips-18:
Also, we are using semantic versioning now. Please see the [Commit Guidelines](https://github.com/patternfly/angular-patternfly#git-commit-guidelines).

src/views/empty-state.html
benjaminapetersen: Works for me, I'll update.

src/views/empty-state.html
benjaminapetersen: I can update this in the tests, though want to bring up that someone targeting with `css` will have a lousy time trying to get at these without a class hook at all (and such a specific id as `empty-state-info-54`.  Are we sure we don't want a hook at all?  (I tend to come from the camp that believes you should never write `css` targeting dom nodes).

src/views/empty-state.html
benjaminapetersen: Same reply as above.  Just to confirm, I'm ok with the change if y'all are certain something like `empty-state-title` & `empty-state-helplink` isn't helpful for those who may need to tweak styles (though I'd prob shorten classes and go for `.empty-state .helplink` by preference, but I'll update to whatever is current convention).

src/views/empty-state.html
jeff-phillips-18: OK, I can buy into that. How about using pf-empty-state-xxx for the classes though?

benjaminapetersen:
Awesome, I'll update to:

`fix(empty state): Change h1 to h4 styled as h1, update to use unique id's generated by view`

If that works.  Is there a convention for scope in `<type>(<scope>): <subject>`, like `emptyState` vs `Empty State` vs `Empty State View`?  I'll look @ past commit history & see if there is anything that stands out & maybe PR the contrib file with a couple contextual examples.

jeff-phillips-18:
the <scope> just becomes the highlighted word for the fix/enhancement. We should use the component name being updated/added as a convention, so pfEmptyState in this case.

benjaminapetersen:
Updated, though I matched classes/ids to `blank-state-pf-*` as this was already used within the view (rather than `pf-empty-state-*`).

PR #593, opened 2017-08-23
fix(pfFilter): do not allow empty filters

### LINKED ISSUE #509: pfFilter: Should not allow empty filters
If I hit the [Enter] key with no filter criteria specified in the filter input textbox, the pfFilter component adds an 'empty' applied filter.  This should be prevented.

dtaylor113:
## Description
Updated so that hitting [Enter] on an empty input filter field doesn't create an empty filter.

Fixes #509

### DIFF
--- a/src/filters/simple-filter/filter-fields-component.js
+++ b/src/filters/simple-filter/filter-fields-component.js
@@ -89,7 +89,7 @@ angular.module('patternfly.filters').component('pfFilterFields', {
     }
 
     function onValueKeyPress (keyEvent) {
-      if (keyEvent.which === 13) {
+      if (keyEvent.which === 13 && ctrl.currentValue && ctrl.currentValue.length > 0) {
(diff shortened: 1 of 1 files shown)


src/filters/simple-filter/filter-fields-component.js
```diff
@@ -89,7 +89,7 @@ angular.module('patternfly.filters').component('pfFilterFields', {
     }
 
     function onValueKeyPress (keyEvent) {
-      if (keyEvent.which === 13) {
+      if (keyEvent.which === 13 && ctrl.currentValue && ctrl.currentValue.length > 0) {
```
jeff-phillips-18: Simplify to:
` if (keyEvent.which === 13 && ctrl.currentValue.length) {`

dtaylor113: ```ctrl.currentValue``` is null

Issue #543, opened 2017-08-02
pfFilterPanel: Responsiveness needs to be addressed/overhauled


dtaylor113:
Attempted to import Bootstrap's variables.less file and use `@media(max-width: @screen-md-min)` to address responsiveness layout in the pfFilterPanel.

A concern was:  _"if someone overrides bootstrap variables then includes a-pf less files, they will lose their overrides"_

We need to figure out, in general, how to address responsiveness in angular-patternfly.   
In addition, the pfFilterPanel needs to updated to handle resposiveness.

dtaylor113:
Merging this issue with #600

PR #601, opened 2017-08-29
feat(pfFilter): Add filter options - inline, results count, items label

jeff-phillips-18:
## Description

Provides options to:

- display filter and result on a single line
- show the total count in the results label (i.e. 5 of 10 Results)
- change the label from Results to a passed in string

## PR Checklist

- [x] Unit tests are included

## Rawgit Link (in lieu of screen shots)
https://rawgit.com/jeff-phillips-18/angular-patternfly/filter-docs/dist/docs/index.html#/api/patternfly.filters.component:pfFilter

### DIFF
--- a/src/filters/filters.less
+++ b/src/filters/filters.less
@@ -98,3 +98,53 @@ pf-filter-panel {
     padding-left: 4px;
   }
 }
+
+.filter-pf {
+  &.inline-filter-pf {
+    flex: 1 1 100%;
+    margin: 15px 15px 7px 0;
+
+    pf-filter-fields,
+    pf-filter-results {
+      display: inline-block;
+    }
+
+    .form-group {
+      margin-bottom: 0;
+      margin-right: 15px;
+    }
+
+    .toolbar-pf-results {
+      border-top: none;
+      margin: 0;
+
+      .col-sm-12 {
+        float: left;
+        padding: 0;
+      }
+
+      h5,
+      p,
+      ul {
+        line-height: 1.43;
+        padding-bottom: 6px;
+        padding-top: 6px;
+      }
+
+      .list-inline {
+        margin-bottom: -5px;
+        padding-right: 5px;
+
+        > li {
+          margin-bottom: 5px;
+        }
+      }
+    }
+  }
+
+  .toolbar-pf-results {
+    .list-inline {
+      margin-left: 0;
+    }
+  }
+}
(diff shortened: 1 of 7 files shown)


styles/angular-patternfly.less
```diff
@@ -1,3 +1,4 @@
+@import "../node_modules/patternfly/node_modules/bootstrap/less/variables.less";
```
dtaylor113: What?!  Didn't you stop me from adding bootstrap variables?  #543

jeff-phillips-18: LOL, right... I will remove.

test/filters/filter.spec.js
```diff
@@ -207,4 +207,51 @@ describe('Directive:  pfFilter', function () {
 
     expect(element.find('.pf-table-view-selected-label').text()).toContain('0 of 10 selected');
   });
+
+  it('should show results inline when specified', function() {
```
dtaylor113: To be truly well rounded, each of these three tests need a negative test.  Ex:  "should not show results inline when not specified"  ```expect(element.find('.filter-pf.inline-filter-pf').length).toBe(0);```

src/filters/filters.less
```diff
@@ -98,3 +98,55 @@ pf-filter-panel {
...
+    margin: (@grid-gutter-width / 2) (@grid-gutter-width / 2) (@grid-gutter-width / 4) 0;
+
+    pf-filter-fields,
+    pf-filter-results {
+      display: inline-block;
+    }
+
+    .form-group {
+      margin-bottom: 0;
+      @media (min-width: 345px) {
```
dtaylor113: Is there a variable for this?  Thought this would be why you @imported bootstrap variables.less?

dtaylor113:
Other than the minor issues I raised, LGTM

src/filters/filters.less
dtaylor113: Are there no patternfly variables which denote the min-widths at diff screen sizes?  If not, we should file an issue and add our own, then we don't have to worry about bootstrap variables being overwritten.

jeff-phillips-18: Removing as this is unnecessary. Thanks.

test/filters/filter.spec.js
jeff-phillips-18: Added negative tests

dtaylor113: You did?  I don't see them?

dtaylor113:
@cdcabrera, ok if I merge this or do you want a chance to review?  -thanks

cdcabrera:
@dtaylor113 taking a look

PR #602, opened 2017-08-29
fix(pfNotificationDrawer): Allow setting of the field to use to track notifications

jeff-phillips-18:
## Description
This will set the ng-repeat’s ‘track by’ value to the given field in the notification.

No visual changes.

Reported by https://github.com/openshift/origin-web-console/issues/1981

cc @spadgett @benjaminapetersen

### DIFF
--- a/src/notification/examples/notification-drawer.js
+++ b/src/notification/examples/notification-drawer.js
@@ -35,6 +35,7 @@
  * @param {boolean} drawerExpanded Flag if the drawer is expanded (only valid if allowExpand is true). Optional, default: false
  * @param {string}  drawerTitle  Title to display for the drawer (leaving this blank will remove the provided expand capability)
  * @param {object} notificationGroups Collection notification groups to add to the drawer. Alternatively, a single group object can be given if categorization is not used.
+ * @param {string} notificationTrackField Optional field from the notifications to use to track by in the notifications listing ($index used otherwise).
  * @param {function} onClose function() Callback for the close button. Close button is shown if this callback is supplied. Callback should set drawerHidden to true to close the drawer.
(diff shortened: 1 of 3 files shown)


src/notification/examples/notification-drawer.js
```diff
@@ -35,6 +35,7 @@
  * @param {boolean} drawerExpanded Flag if the drawer is expanded (only valid if allowExpand is true). Optional, default: false
  * @param {string}  drawerTitle  Title to display for the drawer (leaving this blank will remove the provided expand capability)
  * @param {object} notificationGroups Collection notification groups to add to the drawer. Alternatively, a single group object can be given if categorization is not used.
+ * @param {string} notificationTrackField Optional field from the notifications to use to track by in the notifications listing ($index used otherwise).
```
benjaminapetersen: I don't know if it matters a ton, but perhaps `key` instead of `field`, just for consistency with the lingo Angular uses re: `track by`.  Example:

```
Duplicates in a repeater are not allowed. Use 'track by' expression to specify unique keys. Repeater: {0}, Duplicate key: {1}, Duplicate value: {2}
```
From [the docs](https://docs.angularjs.org/error/ngRepeat/dupes).

benjaminapetersen:
I see I missed the merge by 1 minute 😄

jeff-phillips-18:
Yeah, sorry. If you feel strongly I can update and re-release

benjaminapetersen:
I don't think I feel super strong about it, is there anything similar in other components that makes it worth pressing for consistency?

jeff-phillips-18:
Not that comes to mind.

benjaminapetersen:
ok, we prob good to go then!  Will be good to pick this up in console +1

PR #556, opened 2017-08-08
chore(API docs): update header of angular-patternfly API Docs site

dabeng:
## Description
This is the sub-task of [PTNFLY-2467](https://patternfly.atlassian.net/browse/PTNFLY-2467).

@dlabrecq @dtaylor113 @jeff-phillips-18 @bleathem @LHinson 
Based on the PatternFly Showcase Design[1], I've created a PR for updating the header.
The demo link is here: https://rawgit.com/dabeng/angular-patternfly/new-header-dist/docs/index.html#/api

PR #608, opened 2017-09-01
Revert "chore(API docs): update header of angular-patternfly API Docs site"

### LINKED ISSUE #607: Drawer title appears to have fallen behind the accordion
I suspect with the latest changes to upstream [patternfly drawer](https://github.com/patternfly/patternfly/pull/717) updates.

screen shot 2017-09-01 at 2 38 03 pm

I've pulled master & `npm install && bower install` 2x, don't think it is local.

@jeff-phillips-18 @matthewcarleton

jeff-phillips-18:
Reverts patternfly/angular-patternfly#556

Related to #607 

Fixes #607


jeff-phillips-18:
@dabeng I am reverting this because it messes up the styling of the components in the examples. The CSS is lengthy (I assume it comes from a less file somewhere) and uses '.showcase h3' and other such things. That selector is far too broad and should be limited only to the header (ie. .showcase > header h3).

dtaylor113:
Does commit msg need to start with "chore(....)"

jeff-phillips-18:
Updated the commit wording

dabeng:
@jeff-phillips-18 , could you tell me which the components showcase.css messes up?

jeff-phillips-18:
@dabeng I would have to reload it to find out. I know there are H3's and such that are affected. The gist of it is the .showcase selectors are broad and should be limited to the header, footer, etc.

PR #611, opened 2017-09-07
fix(dependencies): Update version of angularjs-datatables

jeff-phillips-18:
This removes a false nsp security check issue caused by a devDependency only.

@chriskacerguis @LHinson @bleathem


chriskacerguis:
my hero!  Thanks @jeff-phillips-18

PR #612, opened 2017-09-07
fix(pfToastNotificationList): HTML Content for Toast Notifications

cdcabrera:
## Description
Minor adjustment to allow HTML content for Toast Notifications. Minor clean up around spec/tests and example.

***Side notes:*** 
- ~~left the ```@``` bindings as is for ```header``` and ```message``` instead of converting them to one-way-binds. Intended backwards compatibility, but can adjust accordingly~~
- ~~nested a ```span``` underneath the original display ```message``` span, there are no intended visual changes~~

***What remains the same:***
- Existing implementations of this component that make use of Angular syntax/markup related to ```strings```, ```one-way binding```, and ```two-way binding``` should remain untouched, meaning your consumer app shouldn't break functionally (i.e Angular & ECMA/JS errors, etc)

***What could cause an issue***
- Nested a ```span``` underneath the original display ```message``` span, there are no intended visual changes. Custom styling on a consumers part could cause alterations
- Activating the **opt-in** ```html-content``` Boolean could have unintended consequences if your toast notifications contain any form of XML/HTML markup/syntax.

## PR Checklist

- [x] Unit tests are included
- [ ] ~~Screenshots are attached (if there are visual changes in the UI)~~
- [ ] ~~A Designer (@beanh66) is assigned as a reviewer (if there are visual changes in the UI)~~
- [ ] ~~A CSS rep (@cshinn) is assigned as a reviewer (if there are visual changes in the UI)~~

### DIFF
--- a/src/notification/toast-notification.component.js
+++ b/src/notification/toast-notification.component.js
@@ -5,9 +5,10 @@
  * @scope
  *
  * @param {string} notificationType The type of the notification message. Allowed value is one of these: 'success','info','danger', 'warning'
- * @param {string} header The header text of the notification. Accepts HTML content.
- * @param {string} message The main text message of the notification. Accepts HTML content.
+ * @param {string} header The header text of the notification. Accepts HTML content when allowed.
+ * @param {string} message The main text message of the notification. Accepts HTML content when allowed.
  * @param {boolean} showClose Flag to show the close button, default: true
+ * @param {boolean} htmlContent Flag to allow HTML content within the header and message options.
  * @param {function} closeCallback (function(data)) Function to invoke when close action is selected, optional
  * @param {string} actionTitle Text to show for the primary action, optional.
  * @param {function} actionCallback (function(data)) Function to invoke when primary action is selected, optional
@@ -37,9 +38,9 @@
      <div ng-controller="ToastNotificationDemoCtrl" class="row example-container">
        <div class="col-md-12">
          <pf-toast-notification notification-type="{{type}}" header="{{header}}" message="{{message}}"
-              show-close="{{showClose}}" close-callback="closeCallback"
-              action-title="{{primaryAction}}" action-callback="handleAction"
-              menu-actions="menuActions">
+              show-close="{{showClose}}"  html-content="htmlContent"
+              close-callback="closeCallback" action-title="{{primaryAction}}"
+              action-callback="handleAction" menu-actions="menuActions">
          </pf-toast-notification>
 
          <form class="form-horizontal">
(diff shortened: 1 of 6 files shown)


src/notification/toast-notification.component.js
```diff
@@ -190,7 +190,7 @@ angular.module( 'patternfly.notification' ).component('pfToastNotification', {
     'data': '=?'
   },
   templateUrl: 'notification/toast-notification.html',
-  controller: function () {
+  controller: function ($sce) {
```
jeff-phillips-18: Can we add an example of adding HTML to the message?

jeff-phillips-18: How about something more like "Default <strong>notification</strong> message." ? I think we need to make it obvious so that folks don't think the toast message is always italicized.

cdcabrera: Didn't catch your response earlier, gotcha... could also place a CSS class for color in there and maybe limit the italics to a word instead of the whole message

src/notification/toast-notification.component.js
```diff
@@ -249,5 +249,10 @@ angular.module( 'patternfly.notification' ).component('pfToastNotification', {
         ctrl.updateShowClose();
       }
     };
+
+    ctrl.trustAsHtml = function (html) {
+      return $sce.trustAsHtml(html);
```
spadgett: This is kind of scary. Is it really needed for most basic markup? It bypasses all sanitation of things like script tags.

spadgett: I'd feel a lot better about this if it was opt-in. For instance, using a `as-html="true"` attribute. @jeff-phillips-18 @dtaylor113 opinion?

jeff-phillips-18: I would agree with that @spadgett  (and your reviews are always welcomed)

cdcabrera: Sounds good, opt in it is. 

The other option was to take a look at transclusion in some form, though this may have a similar outcome.

cdcabrera:
And for clarity around breaking and downstream consumption...

>  the parameters/values/properties being passed in for ```header``` and ```message``` weren't altered on purpose. They still make use of ```@``` to purposefully not break anything downstream.

spadgett:
>  Edit: aside from the incorporation of displaying HTML

But that is the breaking change :)  I realize it's the point of the PR, but it's also API breaking.

I could pass `message="<foo>"` before, and it displayed fine. Now it will be empty since it's treated as markup. Opt-in fixes that as well as avoiding potential XSS problems. Clients have to explicitly say they want HTML, so they know they have to escape unsafe content.

cdcabrera:
Acknowledge @spadgett, your last example is more spot on than the prior one regarding the use of "foo & bar" XML/HTML markup syntax intended as an escaped display string. 

In part of being responsible towards other downstream consumers we've gone ahead and updated the PR notes to make a distinction between implementation and potential functional/behavioral breaks instead of rolling them together.

jeff-phillips-18:
@spadgett Are you ok with this now that it is opt in?

PR #615, opened 2017-09-11
feat(wizard): Support for wizard at mobile resolutions

jeff-phillips-18:
Show only the selected step and substep titles at mobile resolution
Add the activeStepTitleOnly option to show only the active step title
in the steps indicator (header) when set to true.


jeff-phillips-18:
@cdcabrera Yes, there are PF changes required for this (https://github.com/patternfly/patternfly/pull/735)

PR #616, opened 2017-09-12
fix(pfWizard): Make wizard buttons keyboard accessible

### LINKED ISSUE #603: Accessibility: pf wizard buttons are inaccessible via keyboard navigation
via [issue 406](https://github.com/openshift/origin-web-catalog/issues/406) in web console.

You cannot tab through the wizard content & then access the `cancel`, `next` buttons.

- [ ] button is not actually a button, it is just a node styled like a button w/o a `tabindex`.
- [ ] you can also tab through content below the modal (on the page) while the modal is open.

jeff-phillips-18:
Fixes #603

## Description
Changes the buttons in the pfWizard component to ```<button>```'s making them keyboard accessible. No longer using the pfWizardNext, pfWizardPrevious, or pfWizardCancel components.

### DIFF
--- a/src/wizard/wizard.html
+++ b/src/wizard/wizard.html
@@ -29,25 +29,34 @@
     <div class="wizard-pf-position-override" ng-transclude ></div>
   </div>
   <div class="modal-footer wizard-pf-footer wizard-pf-position-override" ng-class="{'wizard-pf-footer-inline': $ctrl.embedInPage}">
-    <pf-wiz-cancel class="btn btn-default btn-cancel wizard-pf-cancel"
-                   ng-class="{'wizard-pf-cancel-no-back': $ctrl.hideBackButton}"
-                   ng-disabled="$ctrl.wizardDone" ng-click="$ctrl.onCancel()"
-                   ng-if="!$ctrl.embedInPage">{{$ctrl.cancelTitle}}</pf-wiz-cancel>
+    <button ng-if="!$ctrl.embedInPage"
+            class="btn btn-default btn-cancel wizard-pf-cancel"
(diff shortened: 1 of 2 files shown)


jeff-phillips-18:
Yeah, it's just changing the components to buttons, all classes and events stayed the same.

src/wizard/wizard.html
```diff
@@ -29,25 +29,34 @@ <h3 class="blank-slate-pf-main-action">{{$ctrl.loadingWizardTitle}}</h3>
     <div class="wizard-pf-position-override" ng-transclude ></div>
   </div>
   <div class="modal-footer wizard-pf-footer wizard-pf-position-override" ng-class="{'wizard-pf-footer-inline': $ctrl.embedInPage}">
-    <pf-wiz-cancel class="btn btn-default btn-cancel wizard-pf-cancel"
-                   ng-class="{'wizard-pf-cancel-no-back': $ctrl.hideBackButton}"
-                   ng-disabled="$ctrl.wizardDone" ng-click="$ctrl.onCancel()"
-                   ng-if="!$ctrl.embedInPage">{{$ctrl.cancelTitle}}</pf-wiz-cancel>
+    <button ng-if="!$ctrl.embedInPage"
```
benjaminapetersen: Looks good. My only question is, if there is a wizard with more than 2 panels, what happens when the "next" button is clicked?  Does it reset focus on the first item in the next panel?

jeff-phillips-18: This does not handle setting the focus when a page is shown. That would be left to the application. Barring the application setting the focus, the next button would still have the focus.

PR #619, opened 2017-09-12
chore(pfNotificationDrawer): Document how to embedd HTML content in messages

dtaylor113:
## Description
Very similar to #612 but this allows HTML content into notification draw messages.

image

## PR Checklist

- [x] Unit tests are included

### DIFF
--- a/src/notification/examples/notification-drawer.js
+++ b/src/notification/examples/notification-drawer.js
@@ -115,7 +115,7 @@
    <span ng-if="notification.status" class="{{'pull-left ' + $ctrl.customScope.getNotficationStatusIconClass(notification)}}" ng-click="$ctrl.customScope.markRead(notification)"></span>
    <div class="drawer-pf-notification-content" ng-click="$ctrl.customScope.markRead(notification)">
      <span class="drawer-pf-notification-message" tooltip-append-to-body="true" tooltip-popup-delay="500" tooltip-placement="bottom" tooltip="{{notification.message}}">
-      {{notification.message}}
+       <span ng-bind-html="notification.message"></span>
      </span>
      <div class="drawer-pf-notification-info" ng-click="$ctrl.customScope.markRead(notification)">
        <span class="date">{{notification.timeStamp | date:'MM/dd/yyyy'}}</span>
@@ -187,7 +187,7 @@
              {
                uid: 1,
                unread: true,
-               message: "A New Event! Huzzah! Bold.",
+               message: "A New Event! Huzzah! Bold. <a href='http://www.google.com'>Goto Google!</a>",
                status: 'info',
                actions: menuActions,
                timeStamp: currentTime - (1 * 60 * 60 * 1000)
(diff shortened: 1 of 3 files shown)


src/notification/examples/notification-drawer.js
```diff
@@ -114,8 +115,11 @@
    </div>
    <span ng-if="notification.status" class="{{'pull-left ' + $ctrl.customScope.getNotficationStatusIconClass(notification)}}" ng-click="$ctrl.customScope.markRead(notification)"></span>
    <div class="drawer-pf-notification-content" ng-click="$ctrl.customScope.markRead(notification)">
-     <span class="drawer-pf-notification-message" tooltip-append-to-body="true" tooltip-popup-delay="500" tooltip-placement="bottom" tooltip="{{notification.message}}">
-      {{notification.message}}
+     <span ng-if="!$ctrl.htmlContent" class="drawer-pf-notification-message" tooltip-append-to-body="true" tooltip-popup-delay="500" tooltip-placement="bottom" tooltip="{{notification.message}}">
```
dtaylor113: My only concern is that we put the entire notification message in the tooltip and I don't know how/if it works when there is HTML in there.  I can't get tooltips to appear for the pfNotificationDrawer ngDoc example?

jeff-phillips-18:
I'm a bit confused here. Why are we adding this to the component? The message HTML is all defined in the transcluded content.

dtaylor113:
> I'm a bit confused here. Why are we adding this to the component? The message HTML is all defined in the transcluded content.

The actual message is/was just text.  This PR uses $sce similar to #612.  I did initially try to add HTML content into the notification message string but it rendered as text.  Am I missing something?

jeff-phillips-18:
The entire message area is handled by the application. There is no need for the notification drawer component to be involved. You can see that you did not modify any of the notification drawer dist html, just the example and the tests. Its a bit of a utility to be able to use $ctrl.htmlContent flag and the $ctrl.trustAsHtml function, but that could be handled by the applications controller.

dtaylor113:
Ok, I hear what you are saying.  Any value in keeping this PR just as an example on how to do this?  If not, I'll close this PR -thanks

jeff-phillips-18:
I'm ok with keeping the example if we can remove the changes to the component and keep the changes to the example.

dtaylor113:
Ok, will do.  I'm not sure why we do ```$sce.trustAsHtml(html)``` when ```ng-bind-html``` will still render the html and do the ```$sanitize``` checks as well.

cdcabrera:
Yeah, its one of those screwy ones... kind of makes me want to go back and play around with includes or transclusion on the toast notifications. Kudos though @dtaylor113 on avoiding most of the ```$sce``` comments for downstream =)

dtaylor113:
Hmm...not much of a PR after removing the unnecessary stuff :-)

jeff-phillips-18:
yeah well, the tests don't really belong either since you are really only testing test code ;)

dtaylor113:
Should I just kill this PR :-) ?

jeff-phillips-18:
I think that might be best

PR #621, opened 2017-09-13
fix(pfNotificationDrawer): Make expand and collapse links accessible

### LINKED ISSUE #599: Accessibility: Notification drawer expand & collapse links are inaccessible via keyboard navigation
The `a ng-click`s in the header of the notification drawer need an empty `href=""` to be keyboard accessible (expand/collapse links).

Re [this comment](https://github.com/openshift/origin-web-console/pull/1326#issuecomment-325401834) on origin-web-console [PR 1326](https://github.com/openshift/origin-web-console/pull/1326)

Is there accessibility guidelines in upstream patternfly?  This would be a great topic @ that level.

@jeff-phillips-18

amarie401:
## Description
Added ```href```s to the ```ng-click```s in header of the notification drawer to allow keyboard accessibility to the expand/collapse links. This would close #599

### DIFF
--- a/src/notification/notification-drawer.html
+++ b/src/notification/notification-drawer.html
@@ -1,7 +1,7 @@
 <div class="drawer-pf" ng-class="{'hide': $ctrl.drawerHidden, 'drawer-pf-expanded': $ctrl.drawerExpanded}">
-  <div  ng-if="$ctrl.drawerTitle" class="drawer-pf-title">
-    <a ng-if="$ctrl.allowExpand" class="drawer-pf-toggle-expand fa fa-angle-double-left" ng-click="$ctrl.toggleExpandDrawer()"></a>
-    <a  ng-if="$ctrl.onClose" class="drawer-pf-close pficon pficon-close" ng-click="$ctrl.onClose()"></a>
+  <div ng-if="$ctrl.drawerTitle" class="drawer-pf-title">
+    <a href="#0" ng-if="$ctrl.allowExpand" class="drawer-pf-toggle-expand fa fa-angle-double-left" ng-click="$ctrl.toggleExpandDrawer()"></a>
+    <a href="#0" ng-if="$ctrl.onClose" class="drawer-pf-close pficon pficon-close" ng-click="$ctrl.onClose()"></a>
     <h3 class="text-center">{{$ctrl.drawerTitle}}</h3>
   </div>
   <div ng-if="$ctrl.titleInclude" class="drawer-pf-title" ng-include src="$ctrl.titleInclude"></div>


jeff-phillips-18:
This should be a bug fix.

src/notification/notification-drawer.html
```diff
@@ -1,7 +1,7 @@
 <div class="drawer-pf" ng-class="{'hide': $ctrl.drawerHidden, 'drawer-pf-expanded': $ctrl.drawerExpanded}">
   <div  ng-if="$ctrl.drawerTitle" class="drawer-pf-title">
-    <a ng-if="$ctrl.allowExpand" class="drawer-pf-toggle-expand fa fa-angle-double-left" ng-click="$ctrl.toggleExpandDrawer()"></a>
-    <a  ng-if="$ctrl.onClose" class="drawer-pf-close pficon pficon-close" ng-click="$ctrl.onClose()"></a>
+    <a href="javascript:void(0);" ng-if="$ctrl.allowExpand" class="drawer-pf-toggle-expand fa fa-angle-double-left" ng-click="$ctrl.toggleExpandDrawer()"></a>
+    <a href="javascript:void(0);" ng-if="$ctrl.onClose" class="drawer-pf-close pficon pficon-close" ng-click="$ctrl.onClose()"></a>
```
jeff-phillips-18: We need to figure out the unit tests and use href="" or, worst case, use href="#0"

amarie401: getting the same broken test with href="#0" as well

amarie401:
When using ```ng-href=""``` I get this error:

 Component: pfTableView should not show the empty state when items is null FAILED
        Expected '' to contain ''.
        test/table/tableview/table-view.spec.js:88:67

dtaylor113:
> Component: pfTableView should not show the empty state when items is null FAILED

What?!  NotificationDrawer change shouldn't have any affect on pfTableview.

amarie401:
Yeah I found it odd that it was affecting pfTableView

amarie401:
Maybe there is an error on my part. I will re-check.

amarie401:
Can anyone reproduce this error?

amarie401:
@dtaylor113 I commented out that specific unit test and the test passed when using ```ng-href=""``` I'm not sure why I am having test failure in pfTableView when I have only touched ```notification-drawer.html```. Is anyone else having this issue?

dtaylor113:
@amarie401, I'l clone this PR and test

amarie401:
@dtaylor113 thank you!

jeff-phillips-18:
Changing to href="#0" seems to work fine for me.

amarie401:
@jeff-phillips-18 href="#0" is also working for me now. I was able to fix the issue. My local wasn't up to date with angular-pf master. Once I did that, it fixed the weird unit test failure issue. Sorry guys!

dtaylor113:
I think the issue was in:  https://github.com/patternfly/angular-patternfly/blob/master/test/notification/notification-drawer.spec.js#L423
with ```href=""``` and clicking on the ```<a>``` probably did cause a page reload :-)

amarie401:
So should we stick with href=#0?

dtaylor113:
> So should we stick with href=#0?

Yes, if that allows accessibility and doesn't break the unit tests

amarie401:
We want the events to fire correct? So we shouldn't use preventDefault? So would it be best to just use href=#0?

amarie401:
@dtaylor113 I agree. This allows for accessibility without breaking the unit tests as well

amarie401:
@jeff-phillips-18 thoughts?

cdcabrera:
Not necessarily a fan of placing a hash into a link within an Angular app if there's routing involved, but as long as we're handling the click event properly seems ok

amarie401:
Any work around @cdcabrera ?

cdcabrera:
I reinstalled/updated through NPM and ran back through the change and testing...


Just running ```$ npm test``` then performed the following changes, most passed, couple of fails...

- placing an empty ```ng-href=""``` **test passed**
- placing an empty ```href=""``` **test failed**
- placing just a hash in both **test passed**
- placing the hash 0 in both **test passed**
- placing the ```ng-href="javascript:void(0)"``` **test failed**
- placing the ```href="javascript:void(0)"``` **test passed**

From there I decided to start the local server, soooo ran ```$ npm start``` and performed the following changes there were 2 failures, the same 2 as before... 

- placing an empty ```ng-href=""``` **test passed**
  - tabbing **focus fail** in Chrome
- placing an empty ```href=""``` **test failed**
- placing just a hash in both **test passed**
  - tabbing **focus success** in Chrome
- placing the hash 0 in both **test passed**
  - tabbing **focus success** in Chrome
- placing the ```ng-href="javascript:void(0)"``` **test failed**
- placing the ```href="javascript:void(0)"``` **test passed**
  - tabbing **focus success** in Chrome

And finally, I quit the server and ran the tests again by themselves... they matched again

- placing an empty ```ng-href=""``` **test passed**
- placing an empty ```href=""``` **test failed**
- placing just a hash in both **test passed**
- placing the hash 0 in both **test passed**
- placing the ```ng-href="javascript:void(0)"``` **test failed**
- placing the ```href="javascript:void(0)"``` **test passed**

The only thing I could come up with is that possibly when running the tests beside the local serve, if we're using any kind of live reload mechanism that's running parallel to the unit tests and someone hasn't quit say multiple Chrome tabs... all running the project... there may be some funkiness around the testing.

At this point I think it just comes down to consistency for us, or a pattern we want to stick with for the near term. Using the hash 0 works, and as long as we're handling the event properly shouldn't mess with any downstream app routing...

If we're looking for extensions in accessibility it might be a good idea to take a peak at highlighting the rows while a user is tabbing through them (similar to hover state), but I'll leave that as a future UX/Design aspect

So to answer your question @amarie401 for workarounds... eh, lets stick with your PR as is, as long as we all agree and stick to the pattern hopefully for the near term. (which coincidentally also makes search and replacing easy =)

jeff-phillips-18:
Agreed @cdcabrera (any thanks for looking into all the options). Recently this came up in Patternfly as well, and the decision was also to go with href="#0" so I think its best to be consistent here and with Patternfly.

benjaminapetersen:
`href="#0"` is interesting, the argument is you should never have `<some-elem id="0">` ?

jeff-phillips-18:
ID's starting with a number are invalid. so there cannot be an id="0"

benjaminapetersen:
Nice. Thats clever.

PR #624, opened 2017-09-18
chore(API docs): update header of angular-patternfly API Docs site

dabeng:
Hi @jeff-phillips-18 , the class "showcase" has been removed from the body tag in this PR.


misc/patternfly-showcase.css
```diff
@@ -0,0 +1,537 @@
...
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  word-wrap: normal;
+}
+header + .container-fluid {
+  padding-bottom: 50px;
+  padding-top: 20px;
+  transition: padding-left 0.2s cubic-bezier(0.35, 0, 0.25, 1);
+}
+header .navbar.navbar-default {
```
dtaylor113: Hi, this will affect every ```<header>``` element.  We need to just limit this to ```h3```'s under ```showcase```.  Ie:
```
.showcase > header {
   ...
}
```

misc/patternfly-showcase.css
```diff
@@ -0,0 +1,537 @@
+.text-overflow-pf {
```
dtaylor113: You are overwritting ```text-overflow-pf``` for the entire application.  This needs to be limited to just ```.showcase```

misc/patternfly-showcase.css
```diff
@@ -0,0 +1,537 @@
...
+  white-space: nowrap;
+  word-wrap: normal;
+}
+header + .container-fluid {
+  padding-bottom: 50px;
+  padding-top: 20px;
+  transition: padding-left 0.2s cubic-bezier(0.35, 0, 0.25, 1);
+}
+header .navbar.navbar-default {
+  background: #040404 !important;
```
jeff-phillips-18: Shouldn't we be using colors from the Patternfly palette?

jeff-phillips-18: Why so many !important's?

junezhang:
@dtaylor113 @jeff-phillips-18 I've updated the CSS part. the main structure of the less is as below:
screen shot 2017-09-22 at 11 07 36 am
The detail less file is here: https://github.com/junezhang/showcase/blob/master/dist/less/showcase.less
The detail css file is here:
https://github.com/junezhang/showcase/blob/master/dist/styles/patternfly-showcase.css

For your above questions, I answered as following:

- If angular-patternfly doesn't need footer, I think we can remove it. 
- the showcase css is using the same color pattern with PatternFly-org.
- Why so many `important`, because some pieces of code are from PatternFly-org, after checking, I removed some of them. thanks for reminder.

jeff-phillips-18:
The structure of this less file is what I was looking for thanks. If the footer can not be made more specific to the showcase components, they I would not include it's CSS here.

IMO, we should be using the Patternfly color palette for the showcase apps (and for Patternfly.org for that matter) but I'm OK with that being done as an overall Patternfly.org/Repo effort if others agree.

dtaylor113:
> The detail less file is here: https://github.com/junezhang/showcase/blob/master/dist/less/showcase.less

Looks good, just needs to be pushed to this PR
Thanks, Dave

dabeng:
@dtaylor113 , please review again.

jeff-phillips-18:
I've reviewed per @dtaylor113 's comments. I believe they are all resolved.

PR #630, opened 2017-09-20
feat(pfInfoStatusCard): Adds Info Status Card and tests

### LINKED ISSUE #629: create pfInfoStatusCard
We have `pfAggregateStatusCard`, but what if you just want to display a few lines of text or whatever!!

Introducing, the `pfInfoStatusCard` 👍 

### Should look something like the two cards in the upper right of this img:
screen shot 2017-09-20 at 3 20 07 pm

### a closer look
screen shot 2017-09-20 at 3 22 10 pm 

As requested by @serenamarie125

AllenBW:
## Description
As per the issue, creates the requested component as to the spec.

### Whatz it look like? 
screen shot 2017-09-20 at 4 02 40 pm

## PR Checklist

- [x] Unit tests are included

closes #629 

@cdcabrera @jeff-phillips-18 @dtaylor113  I can't add any labels to this, hoping ya can help out (enhancement, tagging those who need to review)

cc @serenamarie125 @chriskacerguis

### DIFF
--- a/src/card/info-status/info-status-card.component.js
+++ b/src/card/info-status/info-status-card.component.js
@@ -1,0 +1,95 @@
+/**
+ * @ngdoc directive
+ * @name patternfly.card.component:pfInfoStatusCard
+ * @restrict E
+ *
+ * @param {object} status Status configuration information<br/>
+ * <ul style='list-style-type: none'>
+ * <li>.title         - the main title of the info status card
+ * <li>.href          - the href to navigate to if one clicks on the title or count
+ * <li>.iconClass     - an icon to display to the left of the count
+ * <li>.iconImage     - an image to display to the left of Infrastructure
+ * <li>.info          - an array of strings to display, each element in the array is on a new line, accepts HTML content
+ * </ul>
+ * @param {boolean=} show-top-border Show/hide the top border, true shows top border, false (default) hides top border
+ * @param {boolean} htmlContent Flag to allow HTML content within the info options
+ *
+ * @description
+ * Component for easily displaying textual information
+ *
+ * @example
+ <example module="patternfly.card">
+
+ <file name="index.html">
+   <div ng-controller="CardDemoCtrl" style="display:inline-block;">
+     <div class="col-md-10">
+       <label>With Top Border, Icon Class, Href</label>
+       <pf-info-status-card status="infoStatus" show-top-border="true"></pf-info-status-card>
+       <br/>
+       <label>No Top Border, Icon Image, No Title</label>
+       <pf-info-status-card status="infoStatusTitless"></pf-info-status-card>
+       <br/>
+       <label>With HTML</label>
+       <pf-info-status-card status="infoStatusAlt" html-content="true"></pf-info-status-card>
+     </div>
+   </div>
+ </file>
+
+ <file name="script.js">
+   angular.module( 'patternfly.card' ).controller( 'CardDemoCtrl', function( $scope ) {
+    $scope.infoStatus = {
+      "title":"TinyCore-local",
+      "href":"#",
+      "iconClass": "fa fa-shield",
+      "info":[
+        "VM Name: aapdemo002",
+        "Host Name: localhost.localdomian",
+        "IP Address: 10.9.62.100",
+        "Power status: on"
+      ]
+    };
+
+    $scope.infoStatusTitless = {
+      "iconImage":"img/OpenShift-logo.svg",
+      "info":[
+        "Infastructure: VMware",
+        "Vmware: 1 CPU (1 socket x 1 core), 1024 MB",
+        "12 Snapshots",
+        "Drift History: 1"
+        ]
+    };
+
+    $scope.infoStatusAlt = {
+      "title":"Favorite Things",
+      "iconClass":"fa fa-heart",
+      "info":[
+        "<i class='fa fa-coffee'>",
+        "<i class='fa fa-motorcycle'>",
+        "<b>Tacos</b>"
+      ]
+    };
+   });
+ </file>
+
+ </example>
+ */
+
+angular.module( 'patternfly.card' ).component('pfInfoStatusCard', {
+  bindings: {
+    status: '=',
+    showTopBorder: '@?',
+    htmlContent: '@?'
+  },
+  templateUrl: 'card/info-status/info-status-card.html',
+  controller: function ($sce) {
+    'use strict';
+    var ctrl = this;
+    ctrl.$onInit = function () {
+      ctrl.shouldShowTopBorder = (ctrl.showTopBorder === 'true');
+      ctrl.shouldShowHtmlContent = (ctrl.htmlContent === 'true');
+      ctrl.trustAsHtml = function (html) {
+        return $sce.trustAsHtml(html);
+      };
+    };
+  }
+});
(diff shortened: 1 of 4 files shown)


src/card/info-status/info-status-card.component.js
```diff
@@ -0,0 +1,91 @@
...
+    showTopBorder: '@?'
+  },
+  templateUrl: 'card/info-status/info-status-card.html',
+  controller: function ($sce) {
+    'use strict';
+    var ctrl = this;
+    ctrl.$onInit = function () {
+      ctrl.shouldShowTopBorder = (ctrl.showTopBorder === 'true');
+      ctrl.trustAsHtml = function (html) {
+        return $sce.trustAsHtml(html);
```
dtaylor113: I don't think we need to trustAsHtml, this bypasses $sanitize.

src/card/info-status/info-status-card.html
```diff
@@ -0,0 +1,19 @@
...
+  <div>
+      <h2 class="card-pf-title" ng-if="$ctrl.status.title">
+        <a href="{{$ctrl.status.href}}" ng-if="$ctrl.status.href">
+          <span class="">{{$ctrl.status.title}}</span>
+        </a>
+        <span ng-if="!$ctrl.status.href">
+            <span class="">{{$ctrl.status.title}}</span>
+          </span>
+      </h2>
+    <p ng-bind-html="$ctrl.trustAsHtml(item)" ng-repeat="item in $ctrl.status.info track by $index"></p>
```
dtaylor113: I think you could just have:
```
<p ng-bind-html="item" ...
```

AllenBW: Nah, we need it, otherwise `Error: [$sce:unsafe] Attempting to use an unsafe value in a safe context.`   If you notice, eerily similar to https://github.com/AllenBW/angular-patternfly/blob/71d481d5def61ec759b19c8a1d0188c432dd3697/src/notification/toast-notification.html#L36 😏

jeff-phillips-18: I think we should make this opt in as well. Treat as text unless html is specified.

src/card/info-status/info-status-card.component.js
```diff
@@ -0,0 +1,91 @@
...
+ * @ngdoc directive
+ * @name patternfly.card.component:pfInfoStatusCard
+ * @restrict E
+ *
+ * @param {object} status Status configuration information<br/>
+ * <ul style='list-style-type: none'>
+ * <li>.title         - the main title of the info status card
+ * <li>.href          - the href to navigate to if one clicks on the title or count
+ * <li>.iconClass     - an icon to display to the left of the count
+ * <li>.iconImage     - an image to display to the left of Infrastructure<li>.info          - an array of strings to display, each element in the array is on a new line, accepts HTML content
```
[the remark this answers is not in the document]
AllenBW: awww how did that happen! 👍

AllenBW:
@jeff-phillips-18 updated as per request 🙇‍♀️

src/card/info-status/info-status-card.component.js
```diff
@@ -0,0 +1,95 @@
...
+   });
+ </file>
+
+ </example>
+ */
+
+angular.module( 'patternfly.card' ).component('pfInfoStatusCard', {
+  bindings: {
+    status: '=',
+    showTopBorder: '@?',
```
cdcabrera: Realize we use the ```@``` syntax related to strings in a few places related to Booleans. May have even used it on the referenced component. Not necessarily a fan of this syntax, especially the inevitable check that has to happen later in the controller ```someString === 'true'```. 

Would "one-way bind" syntax be better here? 

@jeff-phillips-18 @dtaylor113

src/card/info-status/info-status-card.component.js
```diff
@@ -0,0 +1,95 @@
...
+    htmlContent: '@?'
+  },
+  templateUrl: 'card/info-status/info-status-card.html',
+  controller: function ($sce) {
+    'use strict';
+    var ctrl = this;
+    ctrl.$onInit = function () {
+      ctrl.shouldShowTopBorder = (ctrl.showTopBorder === 'true');
```
cdcabrera: Related to the first comment, which would clean up this check...

src/card/info-status/info-status-card.component.js
```diff
@@ -0,0 +1,95 @@
...
+      ctrl.shouldShowHtmlContent = (ctrl.htmlContent === 'true');
```
cdcabrera: Related to first comment

src/card/info-status/info-status-card.component.js
AllenBW: Oh snap, not a fan?  Well hot darn, I just did it cuz it was in place, less lines little simpler to just pass in `<?` gonna update pr to include that, 🙇  @cdcabrera

jeff-phillips-18: We should be consistent with other cards. If we ever change it we should change it for all cards.

src/card/info-status/info-status-card.component.js
jeff-phillips-18: Again,  I would leave it this way to be consistent.

src/card/info-status/info-status-card.component.js
cdcabrera: That's an undertaking for sure, possibly breaking too for downstream if they're dumping in things like the curly brace syntax instead of... 

```html
<some-component a-boolean-attr="normalBoolean"
```
If this is a new element though implementation/consumption downstream wouldn't be as affected

cdcabrera:
After having talked through on this with @jeff-phillips-18 ... looks like an agreement. 

@AllenBW I know it's a little bit of a back track, but can you revert towards the "@" instead of the "<" (one-way) syntax. @jeff-phillips-18 brought up a good point for keeping the card components consistent especially if anyone needs to implement multiples.

However, for newer components in general "one-way" syntax for life (or until the framework alters again)!

AllenBW:
no worries! happy to do 🤘

serenamarie125:
We need a quick visual review here as well - @jennyhaines can you take a look?

jennyhaines:
Thanks for tagging me, @serenamarie125 

Hi @AllenBW - It looks like currently that .info-img is set to "margin-right: 10px". I would change this to 15px to give a bit more breathing room between the icon and content. The "No Top Border, Image Icon, No Title" example shows a small top margin between the content and the top of the card. I would increase it, matching the same height that you have between "Favorite Things" title and the top of the card on the "With HTML" example. I think that across the board, the spacing between individual lines of content could be reduced by 5px since they are a little bit too spaced apart in my opinion. The line spacing for wrapping content could be reduced slightly (maybe 2-4px less) as well.

jennyhaines:
@AllenBW I think there should also be a clear distinction between the labels and the values, so I suggest making the labels (for example, "VM Name:" or "Host Name:") semibold, which I think translates to "font-weight: 600"

AllenBW:
@serenamarie125 oh oh! Did I request tagging of wrong individual for review? 😟 

@jennyhaines Thanks for the input! will play with the pixels and post an update.
For simplicity, i'll call em example 1 2 and 3 (html being the third)  The margin you see in 1 but not in two is a result of 2 not having a title, there is no added margin. 

As for the second comment this card doesn't presume to know anything about the desired display content for each examples we're just looking at strings, (there is no title vs contents), for examples 1 and 2, simple text fields (rather than html), it is not possible to to semi bold the contents.  In 3, a user can design any markup necessary.

PR #633, opened 2017-09-27
Topology map component

Hyperkid123:
## Description
Goal is to create new topology graph component. New component will not have any internal logic such as searching or filtering. This should be done in parent component.

Instead of SVG it uses canvas for rendering. This should increase performance for graphs with big amount of nodes.

Please ignore changes to eslint and package.json, before final commit i will revert it back to original state.  I only wanted to make my life easier. 
## PR Checklist

- [x] use canvas for rendering
- [x] zooming
- [x] dragging
    - ~~dragging nodes while zooming resets position to 0, 0. Have to solve that.~~
- [x] select node(s)
- [x] select edge(s)
- [x] minimap
- [x] Specify correct data format
- [x] Handle different nodes styling
   - ~~node font icon~~
   - ~~node image icon~~
   - ~~node custom opacity, fill, border and icon color~~
   - ~~links dashed line option~~
   - ~~utlization indication~~
   - ~~node tooltip on hover~~
- [x] Code styling
## Data format

Component has these parameters: _nodes, edges, select-node, multi-select-nodes, select-edge, multi-select-edges, tooltip-style, show-node-labels, show-edge-labels_

- Nodes is an array of nodes to render.
- Edges is array of links between nodes.
- Select-node is a function which will return clicked node as a parameter if passed function.
- Multi-select-nodes is a function which will return array of selected nodes. Multiple nodes can be selected by shift/ctrl + clicking.
- Select-edge is a function which returns clicked edge.
- Multi-select-edges is a function which returns array of selected edges. Multiple edges can be selected by shift/ctrl + clicking.
- Tooltip-style is a object with styling properties of node titles
- Show-node-label is a boolean. If true, all node titles will be shown without hovering over nodes.
- Show-edge-labels is a boolean. If true, all edge titles will be shown without hovering over nodes.
### Nodes data format
Array of objects. Only `id` is mandatory parameter
```
nodes = [
  //node with all attributes
  {
    id: 'nodeId1',
    title: 'nodeTitle',
    size: 17, //node radius,
    fileicon: this attribute specifies path to image file. eg: '/some/path/to/image/image.png'. File icon has higher priority than fonticon.
    fonticon: css class of node icon eg: 'fa fa-info'
    fill: '#FFFFFF', //color of node background
    borderColor: '#000000', //color of node border
    iconColor: '#FFFFFF', //color of fontIcon
    opacity: 〈0,1〉, //node opacity
    utilization: 〈0,100〉, //border chart around the node
  },
  {
    id: 'nodeId2',
  }, //example of node without any attributes
]
```
### Edges data format
Array of objects. Each edge have two mandatory attributes _source_ and _target_.
```
edges = [
  {
    source: 'nodeId1', //id of source node in nodes array
    target: 'nodeId2', //id of target node in nodes array
    lineStyle: ['dashed'], //style of line
    title: 'edge title', //edge tool tip
  }
]
```
### Select node(s)
User can select one or more nodes. Single node is selected by clicking.
```
selectNode = function(node){
   console.log('selected node from graph: ', node);
 };
```
Multiple nodes are selected by click + ctrl/shift key
```
multiSelect = function(array) {
  console.log('selected nodes from graph: ', array);
}
```
 ### Select edge(s)
User can select one or more nodes. Single node is selected by clicking.
```
selectEdge = function(edge){
   console.log('selected edge from graph: ', edge);
 };
```
Multiple edges are selected by click + ctrl/shift key
```
multiSelectEdge = function(array) {
  console.log('selected edges from graph: ', array);
}
```

### Tooltip style
Object with styling properties for tooltips. Tooltip style is not mandatory attribute. None of the styling properties is mandatory. If some property is not specified, default value will be used.
Tooltips will only be shown for nodes with `title` property.
 ```
tooltip-style = {
  size: 12,
  font: 'Arial',
  textColor: 'white',
  background: 'rgba(0, 0, 0, .5)',
  borderColor: 'red',
  borderWidth: 10
}
```
### Zooming
Component also supports semantic zooming. Only distance between nodes is growing/shrinking, but node size remains the same. Canvas will zoom towards the position of the mouse cursor.

### DIFF
--- a/src/charts/topology-map/examples/topology-map-view.js
+++ b/src/charts/topology-map/examples/topology-map-view.js
@@ -1,0 +1,1129 @@
+/**
+ * @ngdoc directive
+ * @name patternfly.charts.component:pfTopologyMap
+ * @restrict E
+ *
+ * @param {array} nodes array containing objects representing graph nodes. Each node has these attributes. Only node id is mandatory parameter.
+ * <ul style='list-style-type: none'>
+ *   <li>id: unique node identifier</li>
+ *   <li>title: node title</li>
+ *   <li>size: node radius; default value is 17</li>
+ *   <li>fileicon: this attribute specifies path to image file. eg: <b>'/some/path/to/image/image.png'</b>. File icon has higher priority than fonticon.</li>
+ *   <li>fonticon: css class of node icon eg: <b>'fa fa-info'</b> File icon has higher priority than fonticon.</li>
+ *   <li>fill: string containing color code (basic, RGB, HEX) of node background.</li>
+ *   <li>borderColor: string containing color code (basic, RGB, HEX) of node border.</li>
+ *   <li>iconColor: string containing color code (basic, RGB, HEX) of node icon, if iconType is <b>fonticon</b>.</li>
+ *   <li>opacity: number from 〈0,1〉range, representing node opacity</li>
+ *   <li>utilization: number from〈0,100〉range, representing node utilization percentage</li>
+ * </ul>
+ * @param {array} edges array of objects. Each object represents one edge between two nodes. Source and target are mandatory attributes.
+ * <ul style='list-style-type: none'>
+ *  <li>source: id of source node</li>
+ *  <li>target: id of target node</li>
+ *  <li>lineStyle: stroke style of edge; currently only <b>'dashed'</b> is avaliable</li>
+ *  <li>title: label of edge</li>
+ * </ul>
+ *
+ * @param {boolean=} show-node-labels show/hide all node tooltips
+ *
+ * @param {boolean=} show-edge-labels show/hode all edge tooltips
+ *
+ * @param {object=} tooltip-style object used for tooltip styling. This is an optional parameter.
+ * <ul style='list-style-type: none'>
+ *  <li>size: text size in px</li>
+ *  <li>font: font name. eg: <b>'Arial'</b></li>
+ *  <li>textColor: string containing color code (basic, RGB, HEX) of title text.</li>
+ *  <li>background: string containing color code (basic, RGB, HEX) of title background</li>
+ * </ul>
+ * @param {function (node) =} select-node function that return selected(clicked) node from graph
+ * @param {function (array) =} multi-select-nodes function that returns array of selected nodes. Multiple nodes are selected while holding the <b>ctrl/shift</b> key and clicking
+ * @param {function (edge) =} select-edge function that return selected(clicked) edge from graph
+ * @param {function (array) =} multi-select-edges function that returns array of selected edges. Multiple edges are selected while holding the <b>ctrl/shift</b> key and clicking
+ *
+ * @description
+ *   Component for rendering topology chart on Canvas element. This is just a simple component. It has no searching/filtering or other methods. Will only render given data and return data about selected objects.<br/>
+ *   Component also supports semantic zooming. Only distance between nodes is growing/shrinking, but node size remains the same. Canvas will zoom around the mouse cursor.
+ * @example
+   <example module="patternfly.charts">
+    <file name="index.html">
+      <div ng-controller="TopologyMapCtrl" class="container-topology">
+        <pf-topology-map
+          show-node-labels="showNodeLabels"
+          show-edge-labels="showEdgeLabels"
+          nodes="nodes"
+          edges="edges"
+          select-node="selectNode(node)"
+          tooltip-style="tooltipStyle"
+          multi-select-nodes="multiSelect(array)"
+          select-edge="selectEdge(edge)"
+          multi-select-edges="multiSelectEdges(array)"
+        >
+        </pf-topology-map>
+        <h2>Selected node:</h2>
+        <pre>{{selectedNode}}</pre>
+        <h2>Selected nodes:</h2>
+        <pre>{{selectedNodes}}</pre>
+        <h2>Selected edge:</h2>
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by Hyperkid123** [comment]
@skateman bump

File: eslint.yaml [hunk 1]
```diff
@@ -35,7 +35,7 @@ rules:
   vars-on-top: 2
   no-debugger: 2
   no-cond-assign: 2
-  no-console: 2
+  no-console: 0
```
skateman: YAGNI :wink:

skateman:
@Hyperkid123 the zooming is just wrong, it should zoom the distance among the nodes, but not the nodes itself. Check out my PR...

Hyperkid123:
@skateman Yep i know, i'am having some issues zooming in combination with dragging (v3 / v4 inconsistencies). I know how it is supposed to be, but until i have to make i cooperate first :laughing:

cdcabrera:
Not sure how the bug label was added...

Hyperkid123:
@skateman ping

Hyperkid123:
@skateman can you take a look at this please?

src/charts/topology-map/examples/topology-map-view.js
```diff
@@ -0,0 +1,147 @@
...
+ * @restrict E
+ *
+ * @param {array} nodes array containing objects representing graph nodes. Each node has these attributes. Only node id is mandatory parameter
+ * <ul style='list-style-type: none'>
+ *   <li>id: unique node identifier</li>
+ *   <li>title: node title</li>
+ *   <li>size: node radius; default value is 17</li>
+ *   <li>iconType: string that can be one of <b>'fonticon', 'fileicon'</b>. icon is chosen based on iconType value. If no type has been chosen, default icon will be used.</li>
+ *   <li>iconClass: if iconType is <b>fonticon</b>, node should have icon css class. eg: <b>'fa fa-info'</b></li>
+ *   <li>path: if iconType is <b>fileicon</b>, this attribute specifies path to image file. eg: <b>'/some/path/to/image/image.png'</b></li>
```
skateman: Can't you just use `fonticon` or `fileicon` directly? The priority should be `fileicon` first...as in trees.

src/charts/topology-map/topologyMap.component.js
```diff
@@ -0,0 +1,379 @@
+angular.module('patternfly.charts').component('pfTopologyMap', {
+  bindings: {
+    nodes: '<',
+    edges: '<',
+    selectNode: '<',
+    tooltipStyle: '<?',
+  },
+  templateUrl: 'charts/topology-map.html',
```
dtaylor113: The html file should be located at `charts/topology-map/topology-map.html`

dtaylor113:
Hi @Hyperkid123, I find it strange that there is no hover or selection css/treatment.

dtaylor113:
Are there any plans for multi-select?  Ctrl+Right-Click?  or selection by rectangle?

dtaylor113:
I think folks might like to be able to specify edge titles/labels.   And also perhaps to select an edge as well for various operations.

Hyperkid123:
@dtaylor113 @skateman First of all sorry for my lack of response, i had finals at uni, so i was out of touch for a long time. I have finished this week and i should back next week and will resume work on as soon as i can. @dtaylor113 Your suggestions make sense and will definitely look into that, but give a bit of time co catch up :D

skateman:
@Hyperkid123 what does the `usage` parameter do?

Hyperkid123:
@skateman i think that we have discussed this already. Back in the day, when we had a meeting with guys from angular-pattern fly, somebody(can't remember the name) was showing a demo of new topology. There was a feature, that i have noticed. There was a line around the nodes, that represented relative usage of that node. Don't remember what was the purpose exactly, but i remember it was there...

dtaylor113:
Hi @Hyperkid123, please rebase, angular-patternfly was recently updated to use `"patternfly": "^3.37.1"` to fix Node Security warnings.

skateman:
Ah, that' the utilization. I would rename it to something more generic like: "the-line-around-the-nodes" (actually no idea how would I name it) but that's up to the PF guys, @dtaylor113 what do you think?

dtaylor113:
> Ah, that' the utilization. I would rename it to something more generic like: "the-line-around-the-nodes" (actually no idea how would I name it) but that's up to the PF guys, @dtaylor113 what do you think?

Hi, are we talking about:  https://www.patternfly.org/angular-patternfly/#/api/patternfly.charts.component:pfDonutPctChart ?

Having each node have a used .vs available "line-around-the-node"?

skateman:
@dtaylor113 @Hyperkid123 what about `borderChart`?

dtaylor113:
`utilizationGaugeChart` ?

skateman:
@Hyperkid123 I found a solution for the IE11 topology issue. The fonticonService hack we have in ui-components should be enough, you just have to convert the `XXXX` hex code to `YYYY` decimal and pass it further as `&#YYYY;`. Haven't tried it as I'm on my phone now, but it looks promising :wink:

Hyperkid123:
@skateman thanks. I will check it as soon as i can

Hyperkid123:
@dtaylor113 can you review this please?

Hyperkid123:
ping @dtaylor113

Hyperkid123:
ping @dtaylor113 can you merge this please?

PR #637, opened 2017-09-28
Use the window.imagePath global for the base image path in examples if present.

bleathem:
This PR allows that patternfly.org site to override the image path used in some examples by specifying the IMAGE_PATH global variable.

### DIFF
--- a/src/card/aggregate-status/aggregate-status-card.component.js
+++ b/src/card/aggregate-status/aggregate-status-card.component.js
@@ -68,7 +68,8 @@
  </file>
 
  <file name="script.js">
-   angular.module( 'patternfly.card' ).controller( 'CardDemoCtrl', function( $scope ) {
+   angular.module( 'patternfly.card' ).controller( 'CardDemoCtrl', function( $scope, $window ) {
+    var imagePath = $window.IMAGE_PATH || "img";
     $scope.status = {
       "title":"Nodes",
       "count":793,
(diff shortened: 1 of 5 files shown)


src/card/aggregate-status/aggregate-status-card.component.js
```diff
@@ -68,7 +68,8 @@
  </file>
 
  <file name="script.js">
-   angular.module( 'patternfly.card' ).controller( 'CardDemoCtrl', function( $scope ) {
+   angular.module( 'patternfly.card' ).controller( 'CardDemoCtrl', function( $scope, $window ) {
+    var imagePath = $window.IMAGE_PATH || "img";
```
jeff-phillips-18: we should be consistent and use /img

bleathem: I was trying to preserve what was previously present.  Would you say `img` is expected to be correct over '/img'

src/navigation/application-launcher.component.js
```diff
@@ -77,8 +77,9 @@
    </div>
  </file>
  <file name="script.js">
-   angular.module('patternfly.navigation').controller('applicationLauncherController', ['$scope',
-     function ($scope) {
+   angular.module('patternfly.navigation').controller('applicationLauncherController', ['$scope', '$window',
+     function ($scope, $window) {
+       $scope.imagePath = $window.IMAGE_PATH || "img";
```
jeff-phillips-18: and again...

bleathem: ...again, I was preserving what was previously present.  Happy to converge on a single path with this PR though.

src/card/aggregate-status/aggregate-status-card.component.js
jeff-phillips-18: yeah, I would say 'img' is probably better, I'd just rather see consistency.

bleathem: Done.  Standardised on `img`.

PR #638, opened 2017-10-02
chore(KarmaDependencies): Update karma packages to latest versions

### LINKED ISSUE #623: Update Karma Dependencies
How do we want to handle updating existing dependencies for karma and jasmine?

Is there any reason we can't do these updates?

@cdcabrera @cliffpyles @jeff-phillips-18 @dtaylor113

amarie401:
## Description
Update to the karma/jasmine dev dependencies to their latest versions. This would close #623 

Note: [Seems like this one unit test is failing now](https://github.com/patternfly/angular-patternfly/blob/b8c1f4eb0b3250fc5356906aa5cd0f6e2cebe506/test/table/tableview/table-view.spec.js#L84)

All others are passing after update.


dtaylor113:
Looking into it....

dtaylor113:
Hi @amarie401,

Please change that line to be:

`expect(element.find('.blank-state-pf-title').text().length).toBe(0);`

I'm not sure why the char comparison stopped working with the dependency update, but this fixes it.

amarie401:
I was just about to change it to that to see if it fixed the issue lol..thanks @dtaylor113  :)

PR #640, opened 2017-10-05
fix(Canvas): Canvas directive to component implementation

### LINKED ISSUE #636: Canvas is a directive, consider rewriting it as a component
Consider rewriting the canvas directive as a component.

From slack:
>dana [6:27 AM] 
>Canvas is actually a directive internally - it has a link function and binds to several document listeners

amarie401:
## Description
Conversion of the Canvas directive to a component. This fixes #636 

[Rawgit](https://rawgit.com/amarie401/angular-patternfly/canvas-to-component-dist/dist/docs/index.html#/api/patternfly.canvas.component:pfCanvas)


dtaylor113:
Other than the keystroke issue, LGTM :+1: !

amarie401:
Thanks @dtaylor113 going to look into that issue now

src/canvas-view/canvas/canvas-component.js
```diff
@@ -0,0 +1,411 @@
...
+        ctrl.zoomLevel = function () {
+          return ctrl.chart.zoom.getLevel();
+        };
+
+        ctrl.$onInit = function () {
+          var deleteKeyCode = 46;
+          var ctrlKeyCode = 17;
+          var ctrlDown = false;
+          var aKeyCode = 65;
+          var dKeyCode = 68;
```
jeff-phillips-18: this is not used

src/canvas-view/canvas/canvas-component.js
```diff
@@ -0,0 +1,411 @@
...
+          if (ctrl.chart.inConnectingMode && node !== ctrl.connectingModeSourceNode) {
+            ctrl.chart.createNewConnection(ctrl.connectingModeOutputConnector, ctrl.mouseOverConnector);
+            ctrl.stopConnectingMode();
+          }
+        };
+
+        //
+        // zoom.
+        //
+        $scope.$on('zoomIn', function (evt, args) {
```
jeff-phillips-18: arguments are not used

src/canvas-view/canvas/canvas-component.js
```diff
@@ -0,0 +1,411 @@
...
+          ctrl.chart.zoom.in();
+        });
+
+        $scope.$on('zoomOut', function (evt, args) {
```
jeff-phillips-18: arguments are not used

amarie401:
@dtaylor113 keys should be working now. There was an issue where the mac keyboards use the backspace keycode for "delete" so had to account for both.

dtaylor113:
Hi @amarie401. think you have some debugging code in there :-)
```
/home/travis/build/patternfly/angular-patternfly/src/canvas-view/canvas/canvas-component.js
  392:13  error  Unexpected console statement  no-console
```

src/canvas-view/canvas-editor/toolbox-items.component.js
```diff
@@ -0,0 +1,27 @@
...
+        startDragCallback: '=',
+        clickCallback: '=',
+        searchText: '='
+      },
+      controller: function toolboxItemsController () {
+        var ctrl = this;
+
+        ctrl.clickCallbackfmDir = function (item) {
+          if (!item.disableInToolbox) {
+            ctrl.clickCallback(item);
```
cdcabrera: This may have been in the existing directive, but if we're going for consistency...

we do checks for a function within the ```node-toolbar-component.js``` should we be doing that here too?
Similar too...
```js
if (angular.isFunction(ctrl.clickCallback)) {...
```

src/canvas-view/canvas-editor/toolbox-items.component.js
```diff
@@ -0,0 +1,27 @@
...
+          }
+        };
+
+        ctrl.startDragCallbackfmDir = function (event, ui, item) {
+          ctrl.startDragCallback(event, ui, item);
```
cdcabrera: Same call out on function check...
```js
if (angular.isFunction(ctrl.startDragCallback)) {...
```

src/canvas-view/canvas-editor/toolbox-items.component.js
```diff
@@ -0,0 +1,27 @@
...
+
+  angular.module('patternfly.canvas')
+    .component('toolboxItems', {
+      templateUrl: 'canvas-view/canvas-editor/toolbox-items.html',
+      bindings: {
+        items: '=',
+        startDragCallback: '=',
+        clickCallback: '=',
+        searchText: '='
+      },
```
cdcabrera: One-way binds for the ```callbacks``` using ```<``` do they work? Or do we need some form of the function/callback passed back out?

src/canvas-view/canvas/node-toolbar.component.js
```diff
@@ -0,0 +1,32 @@
...
+
+  angular.module('patternfly.canvas')
+    .component('nodeToolbar', {
+      templateUrl: 'canvas-view/canvas/node-toolbar.html',
+      bindings: {
+        node: '=',
+        nodeActions: '=',
+        nodeClickHandler: '=',
+        nodeCloseHandler: '='
+      },
```
cdcabrera: Same comment regarding using one-way binds instead of two-way on the ```-Handlers``` using ```<```, do we need the handlers passed back out?

src/canvas-view/canvas-editor/toolbox-items.component.js
amarie401: The one-way binds seem to work in this case.

src/canvas-view/canvas-editor/toolbox-items.component.js
amarie401: Can do. Makes sense for consistency purposes.

src/canvas-view/canvas/node-toolbar.component.js
amarie401: From my knowledge, I don't believe the handlers need to be passed back out. Can change these to one-way binds as well.

src/canvas-view/canvas-editor/toolbox-items.component.js
amarie401: Agree..

src/canvas-view/canvas/node-toolbar.component.js
```diff
@@ -0,0 +1,32 @@
...
+  'use strict';
+
+  angular.module('patternfly.canvas')
+    .component('nodeToolbar', {
+      templateUrl: 'canvas-view/canvas/node-toolbar.html',
+      bindings: {
+        node: '=',
+        nodeActions: '=',
+        nodeClickHandler: '=',
+        nodeCloseHandler: '='
```
jeff-phillips-18: We should add comments here (since there is no doc for this) as to what the callback arguments should be.

src/canvas-view/canvas/node-toolbar.component.js
```diff
@@ -0,0 +1,32 @@
...
+      },
+      controller: function NodeToolbarController () {
+        var ctrl = this;
+        ctrl.selectedAction = "none";
+
+        ctrl.actionIconClicked = function (action) {
+          ctrl.selectedAction = action;
+          if (angular.isFunction(ctrl.nodeClickHandler)) {
+            ctrl.nodeClickHandler({'action': action, 'node': ctrl.node});
```
jeff-phillips-18: Might be better to pass 2 arguments rather than an object.

src/canvas-view/canvas/canvas-component.js
```diff
@@ -0,0 +1,411 @@
...
+          var backspaceKeyCode = 8;
+          var deleteKeyCode = 46;
+          var ctrlKeyCode = 17;
+          var ctrlDown = false;
+          var aKeyCode = 65;
+          var escKeyCode = 27;
+
+          $document.find('body').keydown(function (evt) {
+            if (evt.keyCode === ctrlKeyCode) {
+              ctrlDown = true;
```
jeff-phillips-18: You can check evt.ctrlKey to determine if ctrl is pressed rather than keeping state.

src/canvas-view/canvas/canvas-component.js
```diff
@@ -0,0 +1,411 @@
...
+              ctrl.deselectAll();
+              $scope.$digest();
+            }
+
+            if (evt.keyCode === ctrlKeyCode) {
+              ctrlDown = false;
+              evt.stopPropagation();
+              evt.preventDefault();
+            }
+          });
```
jeff-phillips-18: Why keyup sometimes and keydown others? I would think we could check only on keydown for all of these.

src/canvas-view/canvas-editor/toolbox-items.component.js
cdcabrera: @jeff-phillips-18 @dtaylor113 since the syntax on the consumption end should be the same is there a contrast opinion to updating the syntax?

jeff-phillips-18: works for me.

src/canvas-view/canvas/node-toolbar.component.js
amarie401: I agree

src/canvas-view/canvas-editor/toolbox-items.component.js
amarie401: After looking into it, seems like ```ctrl.clickCallback ``` isn't a function so would be unable to check for a function in this instance

src/canvas-view/canvas-editor/toolbox-items.component.js
amarie401: Same as above ```ctrl.startDragCallback``` isn't a function either. Just being used for binding and to pass the arguments into from the it's parent function.

src/canvas-view/canvas-editor/toolbox-items.component.js
jeff-phillips-18: If it isn't a function, how is it being called?

amarie401: Or maybe I'm misunderstanding

amarie401:
@cdcabrera @jeff-phillips-18 made some requested changes. Still working on fixing the eventText issue.

src/canvas-view/canvas-editor/toolbox-items.component.js
```diff
@@ -0,0 +1,31 @@
...
+        clickCallback: '<',
+        searchText: '='
+      },
+      controller: function toolboxItemsController () {
+        var ctrl = this;
+
+        ctrl.clickCallbackfmDir = function (item) {
+          if (angular.isFunction(ctrl.clickCallback)) {
+            if (!item.disableInToolbox) {
+              ctrl.clickCallback(item);
```
jeff-phillips-18: I would && these together for readability.

src/canvas-view/canvas-editor/toolbox-items.component.js
```diff
@@ -0,0 +1,31 @@
...
+      bindings: {
+        items: '=',
+        startDragCallback: '<',
+        clickCallback: '<',
+        searchText: '='
+      },
+      controller: function toolboxItemsController () {
+        var ctrl = this;
+
+        ctrl.clickCallbackfmDir = function (item) {
```
jeff-phillips-18: This is an odd name for the function, what does the 'fm' mean?

amarie401: I have no idea. They were already named these lol

dtaylor113: That stood for 'from directive'.  I think at some point just 'clickCallback' wasn't descriptive enough.

src/canvas-view/canvas/canvas-component.js
```diff
@@ -0,0 +1,406 @@
...
+    // Component that generates the rendered chart from the data model.
+    //
+    .component('pfCanvas', {
+      templateUrl: "canvas-view/canvas/canvas.html",
+      bindings: {
+        chartDataModel: "=",
+        chartViewModel: "=?",
+        readOnly: "=?",
+        hideConnectors: "=?"
+      },
```
cdcabrera: We seem to flip back and forth between double quotes and single without a clear reason? Do you have a preference, or you up for making them consistent? If so encourage you to go back through and check them.

src/canvas-view/canvas/canvas-component.js
```diff
@@ -0,0 +1,406 @@
...
+              ctrl.deselectAll();
+              $scope.$digest();
+            }
+
+            if (evt.keyCode === ctrlKeyCode) {
+              evt.ctrlKey = false;
+              evt.stopPropagation();
+              evt.preventDefault();
+            }
+          });
```
cdcabrera: Just a note you don't need to change it. 

This could have also been flipped to a switch. However it probably wouldn't save us much...

```js
switch(evt.keyCode) {
  case deleteKeyCode:
  case backspaceKeyCode:
    ctrl.deleteSelected();
    $scope.$digest();
    break;

  case escKeyCode:
    ctrl.deselectAll();
    $scope.$digest();
    break;

  case ctrlKeyCode:
    evt.ctrlKey = false;
    evt.stopPropagation();
    evt.preventDefault();
    break;
}
```

src/canvas-view/canvas/node-toolbar.component.js
```diff
@@ -0,0 +1,34 @@
+(function () {
+  'use strict';
+
+  angular.module('patternfly.canvas')
+    .component('nodeToolbar', {
+      templateUrl: 'canvas-view/canvas/node-toolbar.html',
+      bindings: {
+        node: '=',
+        nodeActions: '=',
+        nodeClickHandler: '<', // @callback args - (action {string}, node {Object})
```
cdcabrera: We sure about calling them arguments... what about parameters? 

Also it may be a good idea to be consistent and apply similar comments to other callbacks that pass parameters back.

src/canvas-view/examples/canvas.js
```diff
@@ -163,7 +169,7 @@
                     "fontFamily": "PatternFlyIcons-webfont",
                     "fontContent": "\ue621"
                   }
-                ],
+                ]
```
cdcabrera: Good catch on going through the examples and cleaning up! Thank you!

src/canvas-view/canvas/canvas-component.js
```diff
@@ -0,0 +1,406 @@
...
+          var svgElem =  $element.get(0).children[0];
+          var matrix = svgElem.getScreenCTM();
+          var point = svgElem.createSVGPoint();
+          point.x = (x - evt.view.pageXOffset) / ctrl.zoomLevel();
+          point.y = (y - evt.view.pageYOffset) / ctrl.zoomLevel();
+
+          return point.matrixTransform(matrix.inverse());
+        };
+
+        ctrl.hideConnectors = ctrl.hideConnectors ? ctrl.hideConnectors : false;
```
cdcabrera: You'll have to check it to confirm, but this might be a good spot for some simplification... 
```js
ctrl.hideConnectors = ctrl.hideConnectors || false;
```

src/canvas-view/canvas/canvas-component.js
amarie401: I'm fine with double quotes. I can go back and make them consistent.

src/canvas-view/canvas/canvas-component.js
amarie401: I agree

src/canvas-view/canvas/canvas-component.js
amarie401: I'm fine with changing it to a switch. Looks nice and clean.

src/canvas-view/canvas/node-toolbar.component.js
amarie401: I thought args/params were pretty much the same thing? I actually documented the nodeClickHandler in ng-docs so might not need this comment here.

src/canvas-view/canvas/canvas-component.js
jeff-phillips-18: If the idea is that any value makes this true, then I would use:
```js
ctrl.hideConnectors = !!ctrl.hideConnectors;
```

src/canvas-view/canvas/canvas-component.js
```diff
@@ -0,0 +1,406 @@
...
+        };
+
+        ctrl.zoomLevel = function () {
+          return ctrl.chart.zoom.getLevel();
+        };
+
+        ctrl.$onInit = function () {
+          var backspaceKeyCode = 8;
+          var deleteKeyCode = 46;
+          var ctrlKeyCode = 17;
```
jeff-phillips-18: This is not needed.

src/canvas-view/canvas/canvas-component.js
```diff
@@ -0,0 +1,406 @@
...
+              ctrl.deleteSelected();
+              $scope.$digest();
+            }
+
+            if (evt.keyCode === escKeyCode) {
+              ctrl.deselectAll();
+              $scope.$digest();
+            }
+
+            if (evt.keyCode === ctrlKeyCode) {
```
dtaylor113: You should replace this with `evt.ctrlKey === true` like in the Ctrl-A usage.  Then you can get rid of the `ctrlkeyCode` variable like @jeff-phillips-18  suggests.  -thanks!

amarie401: Will do! Thanks :)

src/canvas-view/canvas/canvas-component.js
```diff
@@ -0,0 +1,406 @@
...
+          var ctrlKeyCode = 17;
+          var aKeyCode = 65;
+          var escKeyCode = 27;
+
+          $document.find('body').keydown(function (evt) {
+            if (evt.ctrlKey === true) {
+              evt.stopPropagation();
+              evt.preventDefault();
+            }
+
```
dtaylor113: You can get rid of the `if (evt.ctrlKey === true)...stopPropagation` statement since you are no longer keeping track of ctrl key state.

src/canvas-view/canvas/canvas-component.js
amarie401: It appears that it’s already considered a boolean, I could cast it but I think the syntax is a bit confusing

src/canvas-view/canvas/node-toolbar.component.js
amarie401: reworded the ng-docs to encompass the concept of both arguments and parameters for the nodeClickHandler in  the ```canvas.js``` example. Do we still need the comment?

src/canvas-view/canvas/canvas-component.js
```diff
@@ -0,0 +1,393 @@
...
+    //
+    // Component that generates the rendered chart from the data model.
+    //
+    .component('pfCanvas', {
+      templateUrl: 'canvas-view/canvas/canvas.html',
+      bindings: {
+        chartDataModel:'=',
+        chartViewModel: '=?',
+        readOnly: '=?',
```
cdcabrera: This is non-breaking, but we use a one way bind on the other readOnly binding would that work here too?

src/canvas-view/canvas/canvas-component.js
```diff
@@ -0,0 +1,393 @@
...
+
+        ctrl.$onInit = function () {
+          var backspaceKeyCode = 8;
+          var deleteKeyCode = 46;
+          var aKeyCode = 65;
+          var escKeyCode = 27;
+
+          $document.find('body').keydown(function (evt) {
+
+            if (evt.keyCode === aKeyCode) {
```
dtaylor113: Hi, you need `evt.ctrlKey === true` here in order to do Ctrl-A, at the moment just hitting 'a' alone will select all.

src/canvas-view/canvas/canvas-component.js
amarie401: Yes it does work here, will change for consistency.

src/canvas-view/canvas/canvas-component.js
amarie401: ahh yes good catch!

amarie401:
We don't seem to unit test this set of components, is that something we could aim for?

@cdcabrera  @jeff-phillips-18 @dtaylor113

cdcabrera:
Unit tests eh, well they might have been helpful with the conversion as a base line comparison. But since they aren't there we could two part it @dtaylor113 @jeff-phillips-18 , and were there any limitations before?

I'm leaning more towards integrating them into this PR, at least a few basics to verify things are loading/showing up. It'll increase the length of time on the PR a bit, but it'll head towards the sweet spot of a well rounded/thought out piece.

jeff-phillips-18:
@amarie401 I'm OK with adding unit tests as part of this PR or as a follow on.

amarie401:
Sounds good! Thanks guys!

dtaylor113:
> One question for @amarie401 @dtaylor113 @cdcabrera , is this a fix or a chore ?

I would think `chore` as underlying changes should be transparent to the end users.

amarie401:
Hmmm...I would say that it's a chore since it's a conversion and all the major pieces are still working.

jeff-phillips-18:
OK @amarie401, could you reword your commit please?

dtaylor113:
Wow, really thought I had added the unit tests, as there are extensive ones [here](https://github.com/dtaylor113/manageiq-ui-self_service/blob/master/tests/blueprints/blueprint-canvas.directive.spec.js) and [here](https://github.com/dtaylor113/manageiq-ui-self_service/blob/master/tests/blueprints/blueprint-editor.directive.spec.js)
But I guess these existing 'blueprint canvas' unit tests would need to be updated to working with the generalized a-pf pfCanvas component.  This should be added as a new issue and can be addressed in a follow on PR.  -thanks

cdcabrera:
I'm good with labeling this work as chore if we can work the unit tests in now... but my vote is for fix since it'll increment and allow consumers to pick their version

dtaylor113:
I'm ok with `fix(..)` as it may possibly introduce a regression.

amarie401:
In that case with the possibility of the rollback then I would agree on labeling it a fix

jeff-phillips-18:
OK, I can go along with that.

PR #645, opened 2017-10-12
feature(wizard): deprecate title parameter, simplify buttons

### LINKED ISSUE #639: Wizard should not use 'title' as a parameter, causes tooltip on the entire wizard.
the title parameter should be deprecated and allow callers to use a different parameter (wizardTitle?) so the entire wizard component does not show the title as a tooltip.

jeff-phillips-18:
Deprecate the title parameter in lieu of wizardTitle as using title
causes the entire wizard to have the title as tooltip.

Unwrap the wizard next and back buttons so they correctly pick up
patternfly styling.

Fixes #639

### DIFF
--- a/src/wizard/examples/wizard.js
+++ b/src/wizard/examples/wizard.js
@@ -1,0 +1,316 @@
+/**
+  * @ngdoc directive
+  * @name patternfly.wizard.component:pfWizard
+  * @restrict E
+  *
+  * @description
+  * Component for rendering a Wizard modal.  Each wizard dynamically creates the step navigation both in the header and the left-hand side based on nested steps.
+  * Use pf-wizard-step to define individual steps within a wizard and pf-wizard-substep to define portions of pf-wizard-steps if so desired.  For instance, Step one can have two substeps - 1A and 1B when it is logical to group those together.
+  * <br /><br />
+  * The basic structure should be:
+  * <pre>
+  * <pf-wizard>
+  *   <pf-wizard-step>
+  *     <pf-wizard-substep><!-- content here --></pf-wizard-substep>
+  *     <pf-wizard-substep><!-- content here --></pf-wizard-substep>
+  *   </pf-wizard-step>
+  *   <pf-wizard-step><!-- additional configuration can be added here with substeps if desired --></pf-wizard-step>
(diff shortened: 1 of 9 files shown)


src/wizard/examples/wizard.js
```diff
@@ -0,0 +1,309 @@
...
+    on-cancel="cancelDeploymentWizard()"
+    next-title="nextButtonTitle"
+    next-callback="nextCallback"
+    back-callback="backCallback"
+    wizard-done="deployComplete || deployInProgress"
+    sidebar-class="example-wizard-sidebar"
+    step-class="example-wizard-step"
+    loading-secondary-information="secondaryLoadInformation"
+    on-step-changed="stepChanged(step, index)">
+      <pf-wizard-step step-title="First Step" substeps="true" step-id="details" step-priority="0" show-review="true" show-review-details="true">
```
cdcabrera: @jeff-phillips-18 feel a little silly asking this, but was trying to get the uib-tooltip to show up on the buttons... 

Can't seem to get them to load after editing the example under ```src/wizard/examples/wizard.js``` and adding the following attributes to "pf-wizard-step"s like...
```html
<pf-wizard-step step-title="First Step" prev-tooltip="'1 PRIOR TIPS'" next-tooltip="'1 CHECK MY TIPS'" substeps="true"... />

<pf-wizard-step step-title="Second Step" prev-tooltip="'2 PRIOR TIPS'" next-tooltip="'2 CHECK MY TIPS'" substeps="false"... />

<pf-wizard-step step-title="Review" prev-tooltip="'3 PRIOR TIPS'" next-tooltip="'3 CHECK MY TIPS'" substeps="t..." />
```

Is there something I'm missing, or maybe a conflicting that needs to be removed to get them to load? If so then an example update might be in order.

jeff-phillips-18: Hmmm.... seems they are not working with or without these changes. I'll take a look.

jeff-phillips-18:
@cdcabrera the wizard module was not importing the correct ui bootstrap module. Fixed that.

Found a few other bugs along the way as well as far as the prev/next buttons enablement.

PR #648, opened 2017-10-13
fix(pfEmptyState): Ability for urlAction in empty state component

amarie401:
## Description
Ability for pfEmptyState component config to have url action to provide a user defined callback method. 

Specifically this will be helpful for this issue:
https://github.com/openshift/origin-web-catalog/issues/461

Which would allow the clear applied filters function to be applied to the url link when a user clicks the link on the empty state. Currently only a url can be provided.

[Rawgit](https://rawgit.com/amarie401/angular-patternfly/emptystate-filter-dist/dist/docs/index.html#/api/patternfly.views.component:pfEmptyState)

## PR Checklist

- [X] Unit tests are included

### DIFF
--- a/src/views/empty-state.component.js
+++ b/src/views/empty-state.component.js
@@ -11,6 +11,14 @@
  *   <li>.icon   - (string) class for main icon. Ex. 'pficon pficon-add-circle-o'
  *   <li>.title  - (string) Text for the main title
  *   <li>.info  - (string) Text for the main informational paragraph
+ *   <li>.helpLink - (object) Contains url specific properties and actions
+ *   <ul style='list-style-type: none'>
+ *     <li>.label - (string) Optional text label which appears before the urlLabel
+ *     <li>.urlLabel - (string) Optional text for the clickable portion of the link
+ *     <li>.url - (string) Optional text for url path
+ *     <li>.urlAction - (function) Optional function to invoke a url action when a callback method is specified.
+ *     When both urlAction and url are specified the component will first execute urlAction then nagivate to the url.
+ *   </ul>
  * </ul>
  * @param {array} actionButtons Buttons to display under the icon, title, and informational paragraph.
  *   <ul style='list-style-type: none'>
--- a/test/views/empty-state.spec.js
+++ b/test/views/empty-state.spec.js
@@ -2,8 +2,6 @@ describe('Component:  pfEmptyState', function () {
   var $scope;
   var $compile;
   var element;
-  var performedAction;
-  var updateCount;
 
   // load the controller's module
   beforeEach(function () {
@@ -24,6 +22,11 @@ describe('Component:  pfEmptyState', function () {
   };
 
   beforeEach(function () {
+
+    var performEmptyStateAction = function () {
+      $scope.eventText = 'Empty State Action Executed';
+    };
+
     $scope.config = {
       icon: 'pficon-add-circle-o',
       title: 'Empty State Title',
@@ -31,7 +34,8 @@ describe('Component:  pfEmptyState', function () {
       helpLink: {
         label: 'For more information please see',
         urlLabel: 'pfExample',
-        url : '#/api/patternfly.views.component:pfEmptyState'
+        url : '#/api/patternfly.views.component:pfEmptyState',
+        urlAction: performEmptyStateAction
       }
     };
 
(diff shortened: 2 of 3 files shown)


test/views/empty-state.spec.js
```diff
@@ -73,6 +79,7 @@ describe('Component:  pfEmptyState', function () {
     expect(element.find('.blank-state-pf-helpLink').text()).toContain('For more information please see');
     expect(element.find('a').text()).toContain('pfExample');
     expect(element.find('a').prop('href')).toContain('#/api/patternfly.views.component:pfEmptyState');
+    expect(element.find('.blank-state-pf-helpLink').click().length).toBe(1);
```
dtaylor113: I think a better way to test callback functions in unit tests would be:

```
element.find('.blank-state-pf-helpLink').click();
expect($scope.eventText).toBe('Empty State Action Executed');
```

src/views/empty-state.component.js
```diff
@@ -11,6 +11,14 @@
  *   <li>.icon   - (string) class for main icon. Ex. 'pficon pficon-add-circle-o'
  *   <li>.title  - (string) Text for the main title
  *   <li>.info  - (string) Text for the main informational paragraph
+ *   <li>.helpLink - (object) Contains url specific properties and actions
+ *   <ul style='list-style-type: none'>
+ *     <li>.label - (string) Text for label
```
dtaylor113: Perhaps "Text label which appears before the `urlLabel`

src/views/empty-state.component.js
```diff
@@ -11,6 +11,14 @@
...
+ *     <li>.urlLabel - (string) Text for url label
```
dtaylor113: Text for the clickable portion of the link

src/views/empty-state.component.js
amarie401: sounds good!

test/views/empty-state.spec.js
amarie401: Gotcha! I had something similar to that initially, thanks Dave! :)

src/views/empty-state.component.js
```diff
@@ -11,6 +11,14 @@
  *   <li>.icon   - (string) class for main icon. Ex. 'pficon pficon-add-circle-o'
  *   <li>.title  - (string) Text for the main title
  *   <li>.info  - (string) Text for the main informational paragraph
+ *   <li>.helpLink - (object) Contains url specific properties and actions
+ *   <ul style='list-style-type: none'>
+ *     <li>.label - (string) Text label which appears before the
+ *     <li>.urlLabel - (string) Text for the clickable portion of the link
+ *     <li>.url - (string) Text for url path
```
jeff-phillips-18: This is optional, right?

src/views/empty-state.component.js
```diff
@@ -11,6 +11,14 @@
...
+ *     <li>.urlAction - (function) Function to invoke a url action when a callback method is specified.
```
jeff-phillips-18: This is optional, right?

src/views/empty-state.component.js
```diff
@@ -151,5 +164,11 @@ angular.module('patternfly.views').component('pfEmptyState', {
         action.actionFn(action);
       }
     };
+
+    ctrl.handleUrlAction = function () {
+      if (ctrl.config.helpLink.urlAction) {
+        ctrl.config.helpLink.urlAction();
+      }
+    };
```
jeff-phillips-18: This is unnecessary. You can just use ng-click="$ctrl.config.helpLink.urlAction()", if it is not defined nothing will happen, no error.

src/views/empty-state.component.js
amarie401: Yes this is optional. Will make sure to note that

src/views/empty-state.component.js
amarie401: Yes. Will make sure to note that

amarie401: looks like helpLink and it's properties are under the "config Optional configuration object" for the config param. Would I still need to note that it's optional?

jeff-phillips-18: I would still mark them as optional for that object.

amarie401: Okay will do.

src/views/empty-state.component.js
```diff
@@ -11,6 +11,14 @@
  *   <li>.icon   - (string) class for main icon. Ex. 'pficon pficon-add-circle-o'
  *   <li>.title  - (string) Text for the main title
  *   <li>.info  - (string) Text for the main informational paragraph
+ *   <li>.helpLink - (object) Contains url specific properties and actions
+ *   <ul style='list-style-type: none'>
+ *     <li>.label - (string) Optional text label which appears before the
+ *     <li>.urlLabel - (string) Optional text for the clickable portion of the link
```
dtaylor113: How about "...before the <code>urlLabel</code>"

src/views/empty-state.component.js
dtaylor113: Still waiting on this minor fix.  After these textual changes, everything LGTM.

src/views/empty-state.component.js
amarie401: Whoops that would be helpful huh? :)

PR #654, opened 2017-10-19
fix(pfNotificationDrawer): Do not add #0 to URL on close or expand

### LINKED ISSUE #641: Closing the notification drawer shouldn't add `#0` to my URL
See commit https://github.com/patternfly/angular-patternfly/commit/8332bb772a1f4b20e434241356208d34d0e2ecf6

Adding the `#0` to the URL has some bad side effects:

1. It breaks the browser history. I now have to press the back button twice to go back.
2. It clears any previous hash value. This is bad for any apps that might store state in the URL (active filters, selected tab, etc.).
3. I think this also completely breaks Angular apps that don't use the HTML5 mode since the current route is stored in the URL fragment?

In angular, `<a href="">` is accessible and doesn't have this bad side effect.

https://docs.angularjs.org/api/ng/directive/a

@andresgalante I know this was added to many of the Patternfly examples, but IMHO I think we should revisit it for the reasons above.

At least for Angular apps, it's better to use `href=""`.

cc @jeff-phillips-18 @dtaylor113

jeff-phillips-18:
## Description
Remove the #0 from the href for the close and expand links in the drawer header. This required commenting out some tests but those are pretty simple and unlikely to break.

Fixes #641

### DIFF
--- a/src/notification/examples/notification-drawer.js
+++ b/src/notification/examples/notification-drawer.js
@@ -64,7 +64,7 @@
            <li class="drawer-pf-trigger dropdown" ng-class="{'open': !hideDrawer}">
              <a class="nav-item-iconic drawer-pf-trigger-icon" ng-click="toggleShowDrawer()">
                <span class="fa fa-bell" title="Notifications"></span>
-               <span ng-if="unreadNotifications" class="badge"> </span>
+               <span ng-if="unreadNotifications" class="badge badge-pf-bordered"> </span>
              </a>
            </li>
          </ul>
--- a/src/notification/notification-drawer.html
+++ b/src/notification/notification-drawer.html
@@ -1,7 +1,7 @@
 <div class="drawer-pf" ng-class="{'hide': $ctrl.drawerHidden, 'drawer-pf-expanded': $ctrl.drawerExpanded}">
   <div ng-if="$ctrl.drawerTitle" class="drawer-pf-title">
-    <a href="#0" ng-if="$ctrl.allowExpand" class="drawer-pf-toggle-expand fa fa-angle-double-left hidden-xs" ng-click="$ctrl.toggleExpandDrawer()"></a>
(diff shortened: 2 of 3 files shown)


amarie401:
I'm not entirely sure if we should comment out the unit tests. I think it provides more of a patch.

amarie401:
@spadgett do you have any input on this?

jeff-phillips-18:
well, angular supports href="" but there is a karma known issue with it. I don't think these tests are testing anything that would break on a change that didn't expect to change this and would be tested manually.

amarie401:
Agreed. It looks like @cdcabrera broke this down here https://github.com/patternfly/angular-patternfly/pull/621

Could upgrades to karma solve this?

jeff-phillips-18:
There is no karma fix and they don't plan on fixing it.

jeff-phillips-18:
Yes, in #621 we agreed to go with #0. What we didn't realize at the time was what @spadgett found, it changes the URL and breaks the browser history. Given the choices, skipping the automated test of very simple functionality seems to be the most reasonable solution.

amarie401:
I ran back threw and removed the eventFire piece and went with a regular JS click and that seems to do the trick.

```html
<a href="" ng-if="$ctrl.allowExpand" class="drawer-pf-toggle-expand fa fa-angle-double-left hidden-xs" ng-click="$ctrl.toggleExpandDrawer()"></a>
<a href="" ng-if="$ctrl.onClose" class="drawer-pf-close pficon pficon-close" ng-click="$ctrl.onClose()"></a>
```
```js
//eventFire(closeButton[0], 'click');
 closeButton[0].click();
 $scope.$digest();

 expect($scope.closed).toBe(true);
```

```js
//eventFire(expandToggle[0], 'click');
 expandToggle[0].click();
 $scope.$digest();

 expandedDrawer = element.find('.drawer-pf.drawer-pf-expanded');
 expect(expandedDrawer.length).toBe(1);
```

jeff-phillips-18:
what is the point?

jeff-phillips-18:
Thanks @amarie401 , updated.

PR #661, opened 2017-10-25
fix(pfCard): Add loading state for pfCard

### LINKED ISSUE #658: The base dashboard card should have a loading state
The base dashboard card (http://www.patternfly.org/pattern-library/cards/base-card/#/design) should have a loading state similar to this:
screen shot 2017-10-23 at 9 52 33 am

We need to verify with Visuals which spinner should be used, as i think we want animation here.   @jennyhaines what are your thoughts?

amarie401:
## Description
Adding a loading state spinner to pfCard to fix #658 

[Rawgit](http://rawgit.com/amarie401/angular-patternfly/card-loading-state-dist/dist/docs/index.html#/api/patternfly.card.component:pfCard%20-%20Timeframe%20Filters)

## PR Checklist

- [X] Unit tests are included

loadingstate

@beanh66 @jennyhaines @dtaylor113 @jeff-phillips-18 @serenamarie125

### DIFF
--- a/src/card/basic/card.component.js
+++ b/src/card/basic/card.component.js
@@ -5,9 +5,11 @@
  *
  * @param {string} headTitle Title for the card
  * @param {string=} subTitle Sub-Title for the card
+ * @param {string=} spinnerText Text for the card spinner
  * @param {boolean=} showTopBorder Show/Hide the blue top border. True shows top border, false (default) hides top border
  * @param {boolean=} showTitlesSeparator Show/Hide the grey line between the title and sub-title.
  * True (default) shows the line, false hides the line
+ * @param {boolean=} showSpinner Show/Hide the spinner for loading state. True shows the spinner, false (default) hides the spinner
  * @param {object=} footer footer configuration properties:<br/>
  * <ul style='list-style-type: none'>
  * <li>.iconClass  - (optional) the icon to show on the bottom left of the footer panel
@@ -35,7 +37,7 @@
  <file name="index.html">
    <div ng-controller="ChartCtrl">
      <label class="label-title">Card With Multiple Utilization Bars</label>
-     <pf-card head-title="System Resources" show-top-border="true" style="width: 65%">
+     <pf-card head-title="System Resources" show-spinner="dataLoading" spinner-text="Loading" show-top-border="true">
        <pf-utilization-bar-chart chart-data=data2 chart-title=title2 layout=layoutInline units=units2 threshold-error="85" threshold-warning="60"></pf-utilization-bar-chart>
        <pf-utilization-bar-chart chart-data=data3 chart-title=title3 layout=layoutInline units=units3 threshold-error="85" threshold-warning="60"></pf-utilization-bar-chart>
        <pf-utilization-bar-chart chart-data=data4 chart-title=title4 layout=layoutInline units=units4 threshold-error="85" threshold-warning="60"></pf-utilization-bar-chart>
@@ -44,37 +46,44 @@
    </div>
  </file>
  <file name="script.js">
- angular.module( 'demo', ['patternfly.charts', 'patternfly.card'] ).controller( 'ChartCtrl', function( $scope ) {
+ angular.module( 'demo', ['patternfly.charts', 'patternfly.card'] ).controller( 'ChartCtrl', function( $scope, $timeout ) {
+
+       $scope.dataLoading = true;
 
        $scope.title2 = 'Memory';
        $scope.units2 = 'GB';
 
-       $scope.data2 = {
-         'used': '25',
-         'total': '100'
-       };
-
        $scope.title3 = 'CPU Usage';
        $scope.units3 = 'MHz';
 
-       $scope.data3 = {
-         'used': '420',
-         'total': '500'
-       };
-
        $scope.title4 = 'Disk Usage';
        $scope.units4 = 'TB';
-       $scope.data4 = {
-         'used': '350',
-         'total': '500'
-       };
 
        $scope.title5 = 'Disk I/O';
-       $scope.units5 = 'I/Ops';
-       $scope.data5 = {
-         'used': '450',
-         'total': '500'
-       };
+         $scope.units5 = 'I/Ops';
+
+       $timeout(function () {
+         $scope.dataLoading = false;
+
+         $scope.data2 = {
+           'used': '25',
+           'total': '100'
+         };
+
+         $scope.data3 = {
+           'used': '420',
+           'total': '500'
+         };
+
+         $scope.data4 = {
+           'used': '350',
+           'total': '500'
+         };
+         $scope.data5 = {
+           'used': '450',
+           'total': '500'
+         };
+      }, 3000 );
 
        $scope.layoutInline = {
          'type': 'inline'
@@ -91,6 +100,8 @@ angular.module('patternfly.card').component('pfCard', {
     subTitle: '@?',
     showTopBorder: '@?',
     showTitlesSeparator: '@?',
+    showSpinner: '<?',
+    spinnerText: '@?',
     footer: '=?',
     filter: '=?'
   },
@@ -104,11 +115,9 @@ angular.module('patternfly.card').component('pfCard', {
         ctrl.currentFilter = ctrl.filter.filters[0];
       }
     }
-
     ctrl.footerCallBackFn = function () {
       ctrl.footerCallBackResult = ctrl.footer.callBackFn();
     };
-
     ctrl.filterCallBackFn = function (f) {
       ctrl.currentFilter = f;
       if (ctrl.filter.callBackFn) {
@@ -130,6 +139,7 @@ angular.module('patternfly.card').component('pfCard', {
 
     ctrl.$onInit = function () {
       ctrl.shouldShowTitlesSeparator = (!ctrl.showTitlesSeparator || ctrl.showTitlesSeparator === 'true');
+      ctrl.showSpinner = ctrl.showSpinner === true;
     };
   }
 });
--- a/src/card/basic/card.html
+++ b/src/card/basic/card.html
@@ -1,15 +1,21 @@
-<div ng-class="$ctrl.showTopBorder === 'true' ? 'card-pf card-pf-accented' : 'card-pf'">
+<div class="card-pf" ng-class="{'card-pf-accented': $ctrl.showTopBorder === 'true'}">
   <div ng-if="$ctrl.showHeader()" ng-class="$ctrl.shouldShowTitlesSeparator ? 'card-pf-heading' : 'card-pf-heading-no-bottom'">
-    <div ng-if="$ctrl.showFilterInHeader()" ng-include="'card/basic/card-filter.html'"></div>
+    <div ng-if="$ctrl.showFilterInHeader()" class="card-pf-footer-in-header" ng-class="{'hide-for-spinner': $ctrl.showSpinner}" ng-include="'card/basic/card-filter.html'"></div>
     <h2 class="card-pf-title">{{$ctrl.headTitle}}</h2>
   </div>
-
   <span ng-if="$ctrl.subTitle" class="card-pf-subtitle">{{$ctrl.subTitle}}</span>
 
-  <div class="card-pf-body">
-    <div ng-transclude></div>
+  <div class="card-pf-body" ng-class="{'show-spinner': $ctrl.showSpinner}">
+    <div ng-class="{'hide-for-spinner': $ctrl.showSpinner}" ng-transclude></div>
+    <div ng-if="$ctrl.showSpinner" class="spinner-container">
+      <div class="loading-indicator">
+        <span class="spinner spinner-lg" aria-hidden="true"></span>
+        <span ng-if="$ctrl.spinnerText" class="loading-text">{{$ctrl.spinnerText}}</span>
+        <label ng-if="!$ctrl.spinnerText" class="sr-only">Loading</label>
+      </div>
+    </div>
   </div>
-  <div ng-if="$ctrl.footer" class="card-pf-footer">
+  <div ng-if="$ctrl.footer" class="card-pf-footer" ng-class="{'hide-for-spinner': $ctrl.showSpinner}">
     <div ng-if="$ctrl.showFilterInFooter()" ng-include="'card/basic/card-filter.html'"></div>
     <p>
       <a ng-if="$ctrl.footer.href" href="{{$ctrl.footer.href}}" ng-class="{'card-pf-link-with-icon':$ctrl.footer.iconClass,'card-pf-link':!$ctrl.footer.iconClass}">
--- a/src/card/examples/card-timeframe.js
+++ b/src/card/examples/card-timeframe.js
@@ -5,9 +5,11 @@
  *
  * @param {string} headTitle Title for the card
  * @param {string=} subTitle Sub-Title for the card
+ * @param {string=} spinnerText Text for the card spinner
  * @param {boolean=} showTopBorder Show/Hide the blue top border. True shows top border, false (default) hides top border
  * @param {boolean=} showTitlesSeparator Show/Hide the grey line between the title and sub-title.
  * True (default) shows the line, false hides the line
+ * @param {boolean=} showSpinner Show/Hide the spinner for loading state. True shows the spinner, false (default) hides the spinner
  * @param {object=} footer footer configuration properties:<br/>
  * <ul style='list-style-type: none'>
  * <li>.iconClass  - (optional) the icon to show on the bottom left of the footer panel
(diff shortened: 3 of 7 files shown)


src/card/examples/card-timeframe.js
```diff
@@ -40,6 +41,9 @@
        <pf-card head-title="Card Title" sub-title="Card Subtitle" show-top-border="true" filter="filterConfigHeader" style="width: 50%">
          Card Contents
        </pf-card>
+     <label class="label-title">Timeframe filter in header & loading state</label>
+       <pf-card show-spinner="true" head-title="Card Title" sub-title="Card Subtitle" show-top-border="true" filter="filterConfigHeader" style="width: 50%">
```
dtaylor113: Please tie this to a $scope var which is initially 'true'.  Then in a $timeout delay, have it set to 'false' .  Just so we can  see the state change.

dtaylor113:
I'm wondering if we should also hide the filter dropdown (in header and/or footer) and the footer if the spinner is showing?  If the data isn't available yet, we probably shouldn't show a filter control yet.

beanh66:
@dtaylor113 Agreed, I would think filters and actions should be hidden as well, thanks!

src/card/basic/card.component.js
```diff
@@ -91,6 +92,7 @@ angular.module('patternfly.card').component('pfCard', {
     subTitle: '@?',
     showTopBorder: '@?',
     showTitlesSeparator: '@?',
+    showSpinner: '@?',
```
jeff-phillips-18: Better to make this a binding and use true/false rather that 'true'/'false'

src/card/basic/card.component.js
```diff
@@ -130,6 +130,9 @@ angular.module('patternfly.card').component('pfCard', {
 
     ctrl.$onInit = function () {
       ctrl.shouldShowTitlesSeparator = (!ctrl.showTitlesSeparator || ctrl.showTitlesSeparator === 'true');
+      if (angular.isUndefined(ctrl.showSpinner)) {
+        ctrl.showSpinner = 'false';
+      }
     };
```
jeff-phillips-18: Once this is a binding you can use
`ctrl.showSpinner = ctrl.showSpinner === true`

src/card/basic/card.html
```diff
@@ -3,10 +3,10 @@
     <div ng-if="$ctrl.showFilterInHeader()" ng-include="'card/basic/card-filter.html'"></div>
     <h2 class="card-pf-title">{{$ctrl.headTitle}}</h2>
   </div>
-
-  <span ng-if="$ctrl.subTitle" class="card-pf-subtitle">{{$ctrl.subTitle}}</span>
+  <span ng-class="{'hide': $ctrl.showSpinner === 'true'}" ng-if="$ctrl.subTitle" class="card-pf-subtitle">{{$ctrl.subTitle}}</span>
 
   <div class="card-pf-body">
+    <div ng-class="{'spinner spinner-lg': $ctrl.showSpinner === 'true'}"></div>
```
jeff-phillips-18: This div should only exist if $ctrl.showSpinner

src/card/basic/card.html
```diff
@@ -3,10 +3,10 @@
...
     <div ng-transclude></div>
```
jeff-phillips-18: This should not be shown if showSpinner is true. You may need to use ng-show though due to the transclude.

src/card/basic/card.component.js
amarie401: ```shouldn't I do ctrl.showSpinner = ctrl.showSpinner === false``` since we want it to be false by default?

amarie401:
@beanh66 should the top border be hidden as well? I updated the screenshot

src/card/examples/card-timeframe.js
```diff
@@ -45,10 +46,19 @@
           footer="footerConfig" filter="filterConfig" style="width: 50%">
         Card Contents
      </pf-card>
+     <label class="label-title">Loading State</label>
+     <pf-card show-spinner="dataLoading" head-title="Card Title" sub-title="Card Subtitle" show-top-border="true" filter="filterConfigHeader" style="width: 50%">
```
dtaylor113: I'd add "Card Contents" here, between the <pf-card> and </pf-card>, just to show that transclusion will come through when the spinner is off.

amarie401: Will do!

src/card/basic/card.component.js
jeff-phillips-18: so you want it to be true if they pass false?

What I have above says it is true only if the specifically pass true. Any other value would resolve to false (the default).

src/card/basic/card.component.js
```diff
@@ -130,6 +130,9 @@ angular.module('patternfly.card').component('pfCard', {
 
     ctrl.$onInit = function () {
       ctrl.shouldShowTitlesSeparator = (!ctrl.showTitlesSeparator || ctrl.showTitlesSeparator === 'true');
+      if (angular.isUndefined(ctrl.showSpinner)) {
```
jeff-phillips-18: No need to check if it is defined.

src/card/basic/card.html
```diff
@@ -1,15 +1,15 @@
 <div ng-class="$ctrl.showTopBorder === 'true' ? 'card-pf card-pf-accented' : 'card-pf'">
   <div ng-if="$ctrl.showHeader()" ng-class="$ctrl.shouldShowTitlesSeparator ? 'card-pf-heading' : 'card-pf-heading-no-bottom'">
-    <div ng-if="$ctrl.showFilterInHeader()" ng-include="'card/basic/card-filter.html'"></div>
+    <div ng-class="{'hide': $ctrl.showSpinner === true}" ng-if="$ctrl.showFilterInHeader()" ng-include="'card/basic/card-filter.html'"></div>
```
dtaylor113: This might be better written as:
`<div ng-if="$ctrl.showFilterInHeader() && !$ctrl.showSpinner" ...`
...or you could even move the logic into the $ctrl.showFilterInHeader() function.

src/card/basic/card.component.js
amarie401: Okay no prob!

src/card/basic/card.html
```diff
@@ -1,15 +1,15 @@
...
     <h2 class="card-pf-title">{{$ctrl.headTitle}}</h2>
   </div>
-
-  <span ng-if="$ctrl.subTitle" class="card-pf-subtitle">{{$ctrl.subTitle}}</span>
+  <span ng-class="{'hide': $ctrl.showSpinner === true}" ng-if="$ctrl.subTitle" class="card-pf-subtitle">{{$ctrl.subTitle}}</span>
 
   <div class="card-pf-body">
+    <div ng-show="$ctrl.showSpinner" ng-class="{'spinner spinner-lg': $ctrl.showSpinner}"></div>
```
jeff-phillips-18: This should be ng-if and no need to ng-class since it won't be shown without the class.

src/card/basic/card.html
```diff
@@ -1,15 +1,15 @@
 <div ng-class="$ctrl.showTopBorder === 'true' ? 'card-pf card-pf-accented' : 'card-pf'">
   <div ng-if="$ctrl.showHeader()" ng-class="$ctrl.shouldShowTitlesSeparator ? 'card-pf-heading' : 'card-pf-heading-no-bottom'">
-    <div ng-if="$ctrl.showFilterInHeader()" ng-include="'card/basic/card-filter.html'"></div>
+    <div ng-class="{'hide': $ctrl.showSpinner === true}" ng-if="$ctrl.showFilterInHeader()" ng-include="'card/basic/card-filter.html'"></div>
     <h2 class="card-pf-title">{{$ctrl.headTitle}}</h2>
   </div>
-
-  <span ng-if="$ctrl.subTitle" class="card-pf-subtitle">{{$ctrl.subTitle}}</span>
+  <span ng-class="{'hide': $ctrl.showSpinner === true}" ng-if="$ctrl.subTitle" class="card-pf-subtitle">{{$ctrl.subTitle}}</span>
```
dtaylor113: I think the standard we use is ng-if instead of ng-class/hide, see comment above.

src/card/basic/card.html
```diff
@@ -1,15 +1,15 @@
...
 
   <div class="card-pf-body">
+    <div ng-show="$ctrl.showSpinner" ng-class="{'spinner spinner-lg': $ctrl.showSpinner}"></div>
     <div ng-transclude></div>
```
jeff-phillips-18: This should be ng-show="!$ctrl.showSpinner"

src/card/basic/card.html
```diff
@@ -1,15 +1,15 @@
...
   </div>
-  <div ng-if="$ctrl.footer" class="card-pf-footer">
+  <div ng-class="{'hide': $ctrl.showSpinner === true}" ng-if="$ctrl.footer" class="card-pf-footer">
```
dtaylor113: I think the standard we use is ng-if instead of ng-class/hide, see comment above.

jeff-phillips-18: I'd rather use ng-show than adding the 'hide' class

src/card/basic/card.html
jeff-phillips-18: ng-show rather than adding the 'hide' class?

src/card/basic/card.html
dtaylor113: :+1:

test/card/basic/card.spec.js
```diff
@@ -85,6 +85,19 @@ describe('Component: pfCard', function() {
...
+      element = compileCard('<pf-card head-title="My card title" sub-title="My card subtitle title" show-top-border="false">Inner content goes here</pf-card>', $scope);
+      cardClass = angular.element(element).find('.spinner-lg');
+      expect(cardClass.length).toBe(0);
+
+      // setting to true should show spinner
+      element = compileCard('<pf-card head-title="My card title" show-spinner="true" sub-title="My card subtitle title" show-top-border="false">Inner content goes here</pf-card>', $scope);
+      cardClass = angular.element(element).find('.spinner-lg');
+      expect(cardClass.length).toBe(1);
+    });
+
```
jeff-phillips-18: Might be better to test the binding. Use show-spinner="dataLoading" then change the value and test again.

amarie401: Sounds good. I'll go back to this

amarie401:
After changing the ```ng-class``` to ```ng-if``` I noticed there is a bit of lag and the icon is still there even after the data loads. I tried changing it to ```ng-show``` instead and there still seems to be a bit of lag that wasn't present before. I'm not sure if it has to do with the ```ng-class``` allowing it still be apart of the DOM but just hiding it which would maybe allow for faster load time but I would think ng-show would also do the same? @jeff-phillips-18 @dtaylor113 

loading-state-2

src/card/basic/card.html
amarie401: Seems to be causing a lag with ng-if and ng-show

serenamarie125:
Although I know in the issue uor guidance was to not have associated "Loading" text, I'm rethinking that decision.  

Reasoning is that we have a similar question that just came in about a content view loading state - and in that situation, i think we need "Loading".  Thoughts on adding it here for consistency with what we should do in a view as well as consistency with No Data Available state having an icon/label pair as well @beanh66 @jeff-phillips-18 @jennyhaines ?

src/card/basic/card.html
dtaylor113: Should be one or the other.   I did have to change ng-ifs to ng-shows in order to not break OpenShift e2e/Protractor tests, so I'd go with ng-show.

beanh66:
@serenamarie125 that is fine with me. I was initially debating the label because I felt like it would be more consistent with the empty states we have. I can see where it is really necessary on larger views or cards. @jennyhaines would you be okay with that or would you rather see it optional maybe?

jeff-phillips-18:
@serenamarie125 @beanh66 I think we should allow passing a dataLoadingText value that can be shown along with the spinner if desired by the application.

amarie401:
I agree with having a dataLoadingText value

dtaylor113:
> After changing the ng-class to ng-if I noticed there is a bit of lag now and the icon is still there for a few seconds even after the data loads.

I think I've run into this before, and I think it was a spinner as well.   I believe I did have to replace `ng-if` and `ng-show` with `ng-class`.  Not saying this was right, but it did fix the problem.

It might have something to do with the angular lifecycle, may need to throw in a `$scope.$digest()` in the controller somewhere.  Looking into it....

beanh66:
@jeff-phillips-18 agreed, so are you saying it would be optional?

src/card/basic/card.component.js
```diff
@@ -44,37 +45,44 @@
...
-         'used': '450',
-         'total': '500'
-       };
+         $scope.units5 = 'I/Ops';
+
+       $timeout(function () {
+         $scope.dataLoading = false;
+
+         $scope.data2 = {
+           'used': '25',
```
dtaylor113: You don't need any of these dataN objects

src/charts/utilization-bar/utilization-bar-chart.component.js
```diff
@@ -172,8 +172,16 @@ angular.module('patternfly.charts').component('pfUtilizationBarChart', {
       prevChartData = angular.copy(ctrl.chartData);
       prevLayout = angular.copy(ctrl.layout);
 
+      if (!ctrl.chartData) {
```
dtaylor113: Why are there changes in `utilization-bar-chart.component.js`?  Is this a git/rebase/commit issue?

src/card/basic/card.component.js
```diff
@@ -44,37 +45,44 @@
...
-       $scope.data5 = {
-         'used': '450',
-         'total': '500'
-       };
+         $scope.units5 = 'I/Ops';
+
+       $timeout(function () {
+         $scope.dataLoading = false;
+
+         $scope.data2 = {
```
dtaylor113: Not sure you need the $scope.dataN objects in the $timeout.

dtaylor113:
LGTM, believe we are just waiting on the `dataLoadingText` stuff

src/card/basic/card.component.js
amarie401: @jeff-phillips-18 would it be fine to remove the data objects?

jeff-phillips-18: Yes, trying to mimic what an application would do. If the data is loading, then there is no data until it is loaded. This uncovered an issue were we expect the data to always be set.

jeff-phillips-18:
@beanh66 Yes, I think it should be optional. Text can be specified by the application or left out.

src/card/basic/card.component.js
dtaylor113: ok, thanks

src/card/basic/card.html
```diff
@@ -1,15 +1,16 @@
...
+  <span ng-if="$ctrl.subTitle && !$ctrl.showSpinner" class="card-pf-subtitle">{{$ctrl.subTitle}}</span>
 
-  <span ng-if="$ctrl.subTitle" class="card-pf-subtitle">{{$ctrl.subTitle}}</span>
-
-  <div class="card-pf-body">
-    <div ng-transclude></div>
+  <div class="card-pf-body" ng-class="{'show-spinner': $ctrl.showSpinner}">
+    <label ng-if="!$ctrl.spinnerText" class="sr-only">Loading</label>
+    <div ng-if="$ctrl.showSpinner" class="spinner spinner-lg" aria-hidden="true"></div>
+    <div ng-class="{'hide-body': $ctrl.showSpinner}" ng-transclude></div>
```
jeff-phillips-18: Shouldn't we be displaying $ctrl.spinnerText if it is set?

amarie401: Working on that now :)

amarie401: whoops did that backwards. Fixing.. :)

dtaylor113:
Should the "Loading" spinnerText be centered under the spinner?
image

test/card/basic/card.spec.js
```diff
@@ -85,6 +85,21 @@ describe('Component: pfCard', function() {
...
+      cardClass = angular.element(element).find('.spinner-lg');
+      expect(cardClass.length).toBe(0);
+
+      // When data is loading, spinner should be present
+      $scope.dataLoading = true;
+      element = compileCard('<pf-card head-title="My card title" show-spinner="dataLoading" sub-title="My card subtitle title" show-top-border="false">Inner content goes here</pf-card>', $scope);
+      cardClass = angular.element(element).find('.spinner-lg');
+      expect(cardClass.length).toBe(1);
+    });
+
```
dtaylor113: Pls add unit test for spinnerText, thanks

amarie401: Okay will do!

jeff-phillips-18:
I believe this should be a feat rather than a fix.

amarie401:
No problem, will change to a feat.

jeff-phillips-18:
@dtaylor113 The text should be to the right of the spinner. There are instances where the body content may not present enough vertical space to show the spinner and text stacked.

beanh66:
@jeff-phillips-18 I was thinking it would be centered under the icon, similar to the empty state. Could we do that when there is enough space and when there is not, maybe just hide the text? 

@serenamarie125 Thoughts?

jeff-phillips-18:
@beanh66 not sure we can tell if it will fit or not. I can't think of any way anyways.

jeff-phillips-18:
This is what we have now:
image

beanh66:
@jeff-phillips-18 Oh okay, got it. Well it looks a lot better being centered, I just didn't like the large separation between the icon/text originally :)

jeff-phillips-18:
The ... is just example text. Applications can do what they will.

amarie401:
I have no issue with the "..." since it is just example text and anything can be entered in by the user to suit their needs

amarie401:
@beanh66 I have removed the "..."

amarie401:
@beanh66 I have removed the "..." from the example text and changed it to "Loading" to be more consistent

beanh66:
Thanks, sorry for nit picking! 😆

PR #664, opened 2017-10-27
fix pfTableView: border style not in line with PatternFly

### LINKED ISSUE #663: pfTableView: border style not in line with PatternFly
The table borders are thicker than they should be.

image

akieling:
fix #663 

Make the pfTableView borders thinner according to PatternFly's [Table View](http://www.patternfly.org/pattern-library/content-views/table-view/).

image

@cshinn

### DIFF
--- a/src/pagination/pagination.html
+++ b/src/pagination/pagination.html
@@ -1,4 +1,4 @@
-<form class="content-view-pf-pagination list-view-pf-pagination clearfix" id="form1">
+<form class="content-view-pf-pagination table-view-pf-pagination list-view-pf-pagination clearfix" id="form1">
   <div class="form-group">
     <div uib-dropdown class="btn-group">
       <button uib-dropdown-toggle type="button" class="btn btn-default">
--- a/src/table/table.less
+++ b/src/table/table.less
@@ -15,6 +15,7 @@ table.dataTable.no-footer {
 }
 table.dataTable {
   margin: 0;
+  border-collapse: collapse;
   thead {
     .sorting_asc {
       background-image: none !important;


src/pagination/pagination.less
```diff
@@ -1,6 +1,9 @@
...
-  align-items: center;
-  .per-page-label {
-    padding-left: 10px;
+.content-view-pf-pagination {
+  border-top: none;
+  .form-group {
+    align-items: center;
+    .per-page-label {
+      padding-left: 10px;
+    }
```
jeff-phillips-18: Shouldn't this be in Patternfly styling? Not sure why the original was here. @dtaylor113 ?

jeff-phillips-18:
Commit message needs to be formatted for semantic release. See https://github.com/patternfly/angular-patternfly#git-commit-guidelines

akieling:
Formatted the commit message according to semantic releases

src/pagination/pagination.less
dtaylor113: Without the less, I get:

image

I'd have to dive deeper into the css used in the a-pf pagination component.  I'm fairly certain it is very close to the css used in pf core.

dtaylor113:
@jeff-phillips-18, are you ready to accept this PR?  Filed a issue for the less comment.  -thanks

src/pagination/pagination.less
akieling: Can this change be done in a separate PR so we can merge & close this one?

jeff-phillips-18: Looks like the real issue here is that the pagination component should have the 'table-view-pf-pagination' selector as well as the 'content-view-pf-pagination'. If we add that for the pagination component when used for a table this change would not be necessary.

akieling:
Your requested changes have been done.

dtaylor113:
@jeff-phillips-18, need your approval -thanks

PR #666, opened 2017-10-27
pfTableView: change templateFn to receive the row item as second parameter

### LINKED ISSUE #665: pfTableView: change templateFn to receive the row item as second parameter
Sometimes we need to create templates that use data from other columns, e.g., an anchor link that uses the item ID in the URL and the item name for display.
Let's pass the row item as a second parameter to the `templateFn` function. This change is backward compatible.

akieling:
fix #665 

@dtaylor113


jeff-phillips-18:
Commit message needs to be formatted for semantic release. See https://github.com/patternfly/angular-patternfly#git-commit-guidelines

akieling:
Formatted the commit message according to semantic release.

PR #668, opened 2017-10-31
Info card styling alignment

seanforyou23:
## Description
During work on [issue-111](https://github.com/patternfly/patternfly-ng/issues/111) in the patternfly-ng repo we noticed there was a styling discrepancy between the info status card examples. After speaking with a member of the design team it was decided to add an additional 15px of bottom margin to this component.

The SVG width/height was added because there was a bug where SVG's with no width/height defined within the xml didn't display at all. This gives them bounds of their own even when not specified.
screen shot 2017-10-31 at 10 40 33 am
screen shot 2017-10-31 at 10 40 43 am

## PR Checklist

- [ ] Unit tests are included

### DIFF
--- a/src/table/tableview/examples/table-view-basic.js
+++ b/src/table/tableview/examples/table-view-basic.js
@@ -20,7 +20,7 @@
   * <ul style='list-style-type: none'>
   *   <li>.header     - (string) Text label for a column header
   *   <li>.itemField    - (string) Item field to associate with a particular column.
-  *   <li>.templateFn - (function) (optional) Template function used to render each cell of the column. Pro: more performant than `htmlTemplate`. Con: doesn't support AngularJS directives in the template, therefore it doesn't support things like ng-click. Example: <pre>templateFn: value => `<span class="text-danger">${value}</span>`</pre>
+  *   <li>.templateFn - (function(value, item)) (optional) Template function used to render each cell of the column. Pro: more performant than `htmlTemplate`. Con: doesn't support AngularJS directives in the template, therefore it doesn't support things like ng-click. Example: <pre>templateFn: (value, item) => `<a href="${item.id}">${value}</a>`</pre>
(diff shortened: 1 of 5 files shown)


src/table/tableview/examples/table-view-basic.js
```diff
@@ -20,7 +20,7 @@
   * <ul style='list-style-type: none'>
   *   <li>.header     - (string) Text label for a column header
   *   <li>.itemField    - (string) Item field to associate with a particular column.
-  *   <li>.templateFn - (function) (optional) Template function used to render each cell of the column. Pro: more performant than `htmlTemplate`. Con: doesn't support AngularJS directives in the template, therefore it doesn't support things like ng-click. Example: <pre>templateFn: value => `<span class="text-danger">${value}</span>`</pre>
+  *   <li>.templateFn - (function(value, item)) (optional) Template function used to render each cell of the column. Pro: more performant than `htmlTemplate`. Con: doesn't support AngularJS directives in the template, therefore it doesn't support things like ng-click. Example: <pre>templateFn: (value, item) => `<a href="${item.id}">${value}</a>`</pre>
```
dtaylor113: Think you need to rebase.  These tableView changes should not be part of this PR.

jeff-phillips-18:
The checkboxes for designer and css reviewers are checked but they were not added as reviewers.

cshinn:
From a visual standpoint, I agree this is better. There are some issues with the status card itself that should be addressed either here or in a separate PR. The class names don't follow our pf-postfix standards for instance

beanh66:
@seanforyou23 these styling changes look good to me, but as discussed I think we decided to bubble this up to an issue in PF core.

src/table/tableview/examples/table-view-basic.js
seanforyou23: Understood, thanks for pointing this out. After speaking with some designers it looks like this issue is best addressed in PF core, so I believe this PR will be closed anyway. Let me know if there is anything else needed from me.

PR #679, opened 2017-11-21
updated .gitignore file

dtaylor113:
## Description
Updated .gitignore file to remove unused rules.


.gitignore
```diff
@@ -1,31 +1,12 @@
...
-coverage/
-
-/docs
-/dist
-node_modules
 npm-debug.log
-lib
-.bower
 
-.vscode
```
cdcabrera: If we're removing text editor/IDE configs, should we just go for all of them... ie. .idea, etc ? Also the comment about ```# generic (system) files/extensions we don't want``` could be wiped

dtaylor113: Sorry, didn't know what .vscode was, I'll add it back...and remove comment

dtaylor113:
@cdcabrera , updated this PR -thanks

PR #682, opened 2017-11-28
fix(pfListView): Fix determination of kebab menu placement

### LINKED ISSUE #677: Kebab menu on list view items in My Services is not always visible
👋 Hi Friends!

So we gotta pickle, [using the listview dropdown menu] (http://www.patternfly.org/angular-patternfly/#/api/patternfly.views.component:pfListView) [in conjunction with the toolbar](http://www.patternfly.org/angular-patternfly/#/api/patternfly.toolbars.component:pfToolbar) does this wacky thing where it opens upwards once, then downwards thereafter...

### What, this sounds crazy, **PROVE IT**
pat

A bugzilla ticket was made in this bug's honor: 
https://bugzilla.redhat.com/show_bug.cgi?id=1514080

jeff-phillips-18:
Update the logic that determines the kebab menu placement (up or down). Hide the kebab menu until the determination is made to avoid the menu flicker.

Fixes #677

### DIFF
--- a/src/views/listview/list-view.component.js
+++ b/src/views/listview/list-view.component.js
@@ -19,7 +19,7 @@ angular.module('patternfly.views').component('pfListView', {
     expandedContent: '?listExpandedContent'
   },
   templateUrl: 'views/listview/list-view.html',
-  controller: function ($window, $element) {
+  controller: function ($window, $element, $timeout) {
     'use strict';
     var ctrl = this;
     var prevPageConfig, prevItems;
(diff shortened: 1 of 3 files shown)


src/views/listview/list-view.component.js
```diff
@@ -123,7 +130,10 @@ angular.module('patternfly.views').component('pfListView', {
       // update the actions based on the current item
       ctrl.updateActions(item);
 
-      $window.requestAnimationFrame(function () {
+      // Hide the kebab until we determine dropup or dropdown to avoid flicker
+      ctrl.kebabMenuStyle = {visibility: 'hidden'};
```
cdcabrera: General question... do we have or use a "visibility: hidden" CSS class in core PF, or was using "ng-class" a no go since we're trying to simply hide the refresh/load tweak?

jeff-phillips-18: there is an invisible class in bootstrap I could use. Good thought, I'll update.

PR #683, opened 2017-12-01
chore(dependencies): Update dependencies to remove potential security vulnerabilities

jeff-phillips-18:


cdcabrera:
@jeff-phillips-18 @dtaylor113 @AllenBW we concerned about the remaining Node Security warning since this was meant to remove some of those warnings?

PR #687, opened 2017-12-07
fix(wizard): update wizard on parameter changes

jeff-phillips-18:
Change appropriate parameters to one-way bindings.

### DIFF
--- a/src/wizard/wizard.component.js
+++ b/src/wizard/wizard.component.js
@@ -3,27 +3,27 @@ angular.module('patternfly.wizard').component('pfWizard', {
   bindings: {
     title: '@',
     wizardTitle: '@',
-    hideIndicators: '=?',
+    hideIndicators: '<?',
     activeStepTitleOnly: '<?',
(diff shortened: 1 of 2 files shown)


src/wizard/wizard.component.js
```diff
@@ -3,27 +3,27 @@ angular.module('patternfly.wizard').component('pfWizard', {
...
+    nextCallback: '<?',
     onFinish: '&',
     onCancel: '&',
-    wizardReady: '=?',
-    wizardDone: '=?',
-    loadingWizardTitle: '=?',
-    loadingSecondaryInformation: '=?',
-    embedInPage: '=?',
+    wizardReady: '<?',
+    wizardDone: '<?',
```
cdcabrera: ```wizardReady``` and ```wizardDone``` don't need their statuses piped back out?

jeff-phillips-18: no, these are just to inform the wizard of the app's status.

PR #689, opened 2017-12-08
Bump patternfly version to 3.31.1 

jeff-phillips-18:
Fixes for wizard and accordion with latest patternfly

*Patternfly v.3.31.1 includes SASS support.

@skateman

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -44,7 +44,7 @@
     "angular-bootstrap": "0.14.x",
     "lodash": "3.x",
     "jquery": "2.x",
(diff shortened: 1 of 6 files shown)


dtaylor113:
@jeff-phillips-18  is bumping to patternfly v3.31.1 enough to provide SASS support in a-pf 3.31.1?  I was expecting to see .scss files or SASS grunt plug-in or something?

skateman:
@dtaylor113 no, but we're okay with the compiled CSS version here.

Issue #688, opened 2017-12-07
Add LESS-to-SASS Converter


dtaylor113:
https://www.npmjs.com/package/grunt-lassie

dtaylor113:
Hi @jeff-phillips-18, @amarie401, @TheRealJon ,

I've added 'lessToSass' and 'sass' task to angular-patternfly.  The resulting sass-2-css is almost identical to the original (less-2-css):  https://www.diffnow.com/?report=jh54y

@TheRealJon, wondering if all these [rules](https://github.com/dtaylor113/angular-patternfly/blob/add-less-to-sass/Gruntfile.js#L142) are necessary?

Results in a `dist/sass` directory.  Still have to add  the `grunt build --sass` feature to angular-patternfly.

TheRealJon:
@dtaylor113 All of them should not be necessary. There are a few that are specific to Ruby related tooling and also specific to the patternfly core repo. Once you submit a PR, I can comment on which rules are unnecessary, or I can submit a PR to your fork if you want.

TheRealJon:
As you may have noticed, I am the publisher of that grunt-lassie package. I forked an existing and abandoned converted (grunt-less-to-sass) and published it so that we can update it if we need to. Some of these conversion rules could be added to that tool so that we don't need to copy/paste rules between projects.

dtaylor113:
Thanks @TheRealJon, Either way works for me.  Feel free to post PR to my branch, or when I'm ready we can handle unneeded rules via PR comments.  -thanks

TheRealJon:
@dtaylor113 This PR should remove all the unneeded rules:
https://github.com/dtaylor113/angular-patternfly/pull/5

dtaylor113:
Thanks @TheRealJon !

PR #692, opened 2017-12-13
feat(sass): Added Support for Less to Sass Conversion

dtaylor113:
## Description

From the updated README.md:

#### Less to Sass Conversion

During the build process Less files are converted to Sass files in `/dist/sass`.  Then the Sass files are compiled into `/dist/sass/angular-patternfly.css`. If you would like to copy the Sass generated css into the main `/dist/styles` directory, execute:

```
grunt build --sass
```

This task will copy `/dist/sass/angular-patternfly.css` to `/dist/styles/angular-patternfly.css`.  Then the build process will minimize the css in `/dist/styles`.

The Less to Sass Conversion step will be accomplished and managed as a part of any Pull Request which includes Less file changes. Although contributors may want to build and test their style changes with Sass before submitting a Pull Request, this step should always be tested and validated by reviewers before a style change is merged and released. If a contributor is having issues with Sass conversion that they cannot resolve, Pull Request reviewers will need to ensure that the Sass conversion step is successfully accomplished, tested, and included in the Pull Request before it is approved and merged.

For more detailed information, please read [PatternFly Less to Sass Conversion](https://github.com/patternfly/patternfly#less-to-sass-conversion)

*Note:* When a Less file is added/deleted/renamed it needs to be updated in the main Less import file `/styles/angular-patternfly.less` and the main Sass import file `/styles/_angular-patternfly.scss`.

### New `dist/sass` directory and files.

image

Fixes: #688 

**The sass-2-css file is almost identical to the less-2-css file; just diffs in line breaks.  The minified versions are identical!**

**Thanks to @TheRealJon for leading the way in the less-2-sass conversion process :-)

### DIFF
--- a/eslint.yaml
+++ b/eslint.yaml
@@ -11,7 +11,7 @@ globals:
 rules:
   strict: [2, "function"]
   quotes: 0
-  camelcase: 2
+  camelcase: 1
   indent: [2, 2]
   new-cap:
     - 2
--- a/styles/build.scss
+++ b/styles/build.scss
@@ -1,0 +1,1 @@
+@import 'angular-patternfly';
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -5,6 +5,9 @@ module.exports = function (grunt) {
 
   function init () {
 
+    // '--sass' command line argument exists?
+    var sassBuild = grunt.option('sass');
+
     grunt.initConfig({
       availabletasks: {
         tasks: {
@@ -94,6 +97,23 @@ module.exports = function (grunt) {
           dest: 'dist/less/dependencies/patternfly',
           expand: true,
           flatten: true
+        },
+        distsass: {
+          src: ['styles/_angular-patternfly.scss'],
+          dest: 'dist/sass',
+          expand: true,
+          flatten: true
+        },
+        sassBuild:{
+          files: [
+            {
+              // copy css built from sass into dist/styles
+              expand: true,
+              cwd: 'dist/sass',
+              src: ['**/*.css', '**/*.map'],
+              dest: 'dist/styles/'
+            }
+          ]
         }
       },
       'string-replace': {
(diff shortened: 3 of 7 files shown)


TheRealJon:
Do we want a top level `_angular-patternfly.scss` file to import the rest of the partials?

dtaylor113:
> Do we want a top level _angular-patternfly.scss file to import the rest of the partials?

Do you mean in addition to `styles/build.scss`?

dtaylor113:
Don't know why Travis is failing with:

```

Running "sass:patternfly" (sass) task
>> 
Error: File to import not found or unreadable: ../node_modules/patternfly/dist/sass/patternfly/_color-variables.scss.
>> 
       Parent style sheet: /home/travis/build/patternfly/angular-patternfly/styles/build.scss
>> 
        on line 1 of styles/build.scss
>> 
>> @import '../node_modules/patternfly/dist/sass/patternfly/_color-variables.scss
>> 
   ^
```
`sass:patternfly` runs fine locally.   package-lock.json has patternfly at  v3.31.2 which should have the `../node_modules/patternfly/dist/sass/patternfly/_color-variables.scss` file.

TheRealJon:
> > Do we want a top level _angular-patternfly.scss file to import the rest of the partials?  

> Do you mean in addition to styles/build.scss?

Yes. The build.scss file is for dev/test purposes to build css from the sass (since sass ignores partials by default, you can't compile from a file that starts with `_`). The _angular-patternfly.scss partial would ship in dist so that it could be consumed. That way consumers don't have to import individual partials. Then build.scss could also just `@impirt 'angular-patternfly';`.

dtaylor113:
> The _angular-patternfly.scss partial would ship in dist so that it could be consumed. That way consumers don't have to import individual partials.

Ok, implemented this.  May need a little help documenting how app devs are supposed to consume `_angular-patternfly.scss`.  -thanks

dtaylor113:
Awesome!  Thanks @TheRealJon !!!

dtaylor113:
Arrrrg...Travis still throwing:
```
Running "sass:patternfly" (sass) task
>> Error: File to import not found or unreadable: patternfly/color-variables.
>>        Parent style sheet: /home/travis/build/patternfly/angular-patternfly/styles/_angular-patternfly.scss
>>         on line 1 of styles/_angular-patternfly.scss
>> >> @import 'patternfly/color-variables';
>>    ^
```

TheRealJon:
Yeah, I just noticed this. I don't know why. The build runs fine on my machine. I'll fiddle around with it and see if I can figure something out.

dtaylor113:
Thanks @TheRealJon the switch to `"patternfly": "^3.31.0"` did the trick!!!!

styles/build.scss
```diff
@@ -0,0 +1 @@
+@import 'angular-patternfly.scss';
```
TheRealJon: @dtaylor113 can you remove the `.scss` extension from this import statement so that all of the imports are standardized?

dtaylor113: ok, done

eslint.yaml
```diff
@@ -11,7 +11,7 @@ globals:
 rules:
   strict: [2, "function"]
   quotes: 0
-  camelcase: 2
+  camelcase: 1
```
jeff-phillips-18: Is there a reason for this change?

styles/build.scss
```diff
@@ -0,0 +1 @@
+@import 'angular-patternfly';
```
jeff-phillips-18: I think this file needs to be named angular-patternfly.scss for it to be intuitive for consumption. This likely means moving the partials into a sub-directory or adding all the partials here rather than having an _angular-patternfly.scss

eslint.yaml
dtaylor113: Yes, was throwing an error due to new task in gruntfile 'convert_within_custom_replacements' not being camelcase:
```
lessToSass: {
        convert_within_custom_replacements: 
```

styles/build.scss
dtaylor113: @TheRealJon recommended the main partial:
"Yes. The build.scss file is for dev/test purposes to build css from the sass (since sass ignores partials by default, you can't compile from a file that starts with _). The _angular-patternfly.scss partial would ship in dist so that it could be consumed. That way consumers don't have to import individual partials. Then build.scss could also just @import 'angular-patternfly';."

TheRealJon: It doesn't **have** to start with an underscore, but we are doing it that way in patternfly core, so I figured it would be better to keep things consistent. It's also done similarly in other libraries like bootstrap-sass and font-awesome-sass.

dtaylor113: Just tested renaming to "angular-patternfly.scss" and css generated correctly.   Just want to make sure that "...The angular-patternfly.scss partial would ship in dist so that it could be consumed. That way consumers don't have to import individual partials.."  is still correct.  In other words, app devs can sill consume (@import?)  'dist/angular-patternfly.scss', even though it doesn't begin with an underscore.  -thanks

jeff-phillips-18: OK, this is fine then.

jeff-phillips-18:
I think this should be a feat. This is worthy of a point release.

styles/build.scss
TheRealJon: Just as a reference:

[Sass Partials](http://sass-lang.com/guide#topic-4)
[Sass Imports](http://sass-lang.com/guide#topic-5)

Sass does a bit of magic with imports and file names, so those sections definitely cleared some confusion up for me.

PR #697, opened 2018-01-09
fix(pfEmptyState): Do not redirect when no helpLink url specified

dtaylor113:
## Description
When using the pfEmptyState component and specifying a `helpLink`, the `helpLink` will render a `<a>`.  The `helpLink` can have a `url` and/or a `urlAction`.  Problem was if `url` was not specified, it defaulted to `<a href`  which redirects to root `/`.  
This PR ensures that if no `helpLink.url` specified, that clicking on the `<a>` only executes the `urlAction` (ng-click), and does not redirect and change the browser url.  This is a requirement for OpenShift.

image

### DIFF
--- a/src/views/empty-state.component.js
+++ b/src/views/empty-state.component.js
@@ -107,7 +107,7 @@ angular.module('patternfly.views').component('pfEmptyState', {
   templateUrl: 'views/empty-state.html',
   controller: function ($filter) {
     'use strict';
-    var ctrl = this;
+    var ctrl = this, prevConfig;
 
     ctrl.defaultConfig = {
       title: 'No Items Available'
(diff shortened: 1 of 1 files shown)


src/views/empty-state.component.js
```diff
@@ -117,6 +117,10 @@ angular.module('patternfly.views').component('pfEmptyState', {
       if (angular.isUndefined(ctrl.config)) {
         ctrl.config = {};
       }
+      if (ctrl.config.helpLink && angular.isUndefined(ctrl.config.helpLink.url)) {
+        // if no url specified, set url to not redirect.  ie. just do urlAction
+        ctrl.config.helpLink.url = "javascript:void(0)";
+      }
       ctrl.updateConfig();
```
jeff-phillips-18: This needs to be done in $onChanges as well for the case where the helpLink was added to the config object after init. Maybe do this as part of updateConfig.

dtaylor113: Fixed, added a `$doCheck` since we need to do a deep compare on the config obj.

PR #703, opened 2018-01-22
fix(pfEmptyState): avoid 'unsafe:javascript(void)' for empty helpLink url

dtaylor113:
## Description
When `config.helplink.url` isn't specified, we were using `javascript:void()` in order to avoid the `<a href` from redirecting.  Unfortunately, angularjs adds a `unsafe:javascript:void()`.   To avoid this, we now use a `button` if `config.helplink.url` isn't specified.
Also moved the `ng-click` from the parent `<p>` to the `<a>` or the `<button>`

### DIFF
--- a/src/views/empty-state.html
+++ b/src/views/empty-state.html
@@ -8,8 +8,10 @@
   <p id="blank-state-pf-info-{{$id}}" class="blank-state-pf-info" ng-if="$ctrl.config.info">
     {{$ctrl.config.info}}
   </p>
-  <p id="blank-state-pf-helpLink-{{$id}}" class="blank-state-pf-helpLink" ng-if="$ctrl.config.helpLink" ng-click="$ctrl.config.helpLink.urlAction()">
-    {{$ctrl.config.helpLink.label}} <a href="{{$ctrl.config.helpLink.url}}">{{$ctrl.config.helpLink.urlLabel}}</a>.
+  <p id="blank-state-pf-helpLink-{{$id}}" class="blank-state-pf-helpLink-label" ng-if="$ctrl.config.helpLink">
+    {{$ctrl.config.helpLink.label}}
+    <a ng-if="$ctrl.config.helpLink.url" class="blank-state-pf-helpLink" ng-click="$ctrl.config.helpLink.urlAction()" href="{{$ctrl.config.helpLink.url}}">{{$ctrl.config.helpLink.urlLabel}}</a>
+    <button ng-if="!$ctrl.config.helpLink.url" class="btn blank-state-pf-helpbutton btn-link" ng-click="$ctrl.config.helpLink.urlAction()">{{$ctrl.config.helpLink.urlLabel}}</button>.
   </p>
   <div ng-if="$ctrl.hasMainActions()" class="blank-slate-pf-main-action">
     <button class="btn btn-primary btn-lg"
--- a/src/views/views.less
+++ b/src/views/views.less
@@ -108,6 +108,11 @@
   button {
     margin-right: 4px;
   }
+  .blank-state-pf-helpbutton {
+    padding-left: 0;
+    padding-right: 0;
+    vertical-align: baseline;
+  }
 }
 /* overriding pf base so that buttons have spaces between them */
 .pf-expand-placeholder {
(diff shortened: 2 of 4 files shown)


src/views/empty-state.html
```diff
@@ -9,7 +9,7 @@ <h4 id="blank-state-pf-title-{{$id}}" class="h1 blank-state-pf-title">
     {{$ctrl.config.info}}
   </p>
   <p id="blank-state-pf-helpLink-{{$id}}" class="blank-state-pf-helpLink" ng-if="$ctrl.config.helpLink" ng-click="$ctrl.config.helpLink.urlAction()">
-    {{$ctrl.config.helpLink.label}} <a href="{{$ctrl.config.helpLink.url}}">{{$ctrl.config.helpLink.urlLabel}}</a>.
+    {{$ctrl.config.helpLink.label}} <a class="pf-helplink" ng-attr-href="{{$ctrl.config.helpLink.url ? $ctrl.config.helpLink.url : undefined}}">{{$ctrl.config.helpLink.urlLabel}}</a>.
```
jwforres: this fails accessibility via keyboard navigation

spadgett: Any reason not to use...

```html
<a ng-if="$ctrl.config.helpLink.url" class="pf-helplink" ng-href="{{$ctrl.config.helpLink.url}}">{{$ctrl.config.helpLink.urlLabel}}</a>
<a ng-if="!$ctrl.config.helpLink.url" class="pf-helplink" href="" role="button">{{$ctrl.config.helpLink.urlLabel}}</a>
```

dtaylor113: Did something similar but with a href and a button.

spadgett: > Did something similar but with a href and a button.

even better :) thanks

src/views/empty-state.html
```diff
@@ -8,8 +8,10 @@ <h4 id="blank-state-pf-title-{{$id}}" class="h1 blank-state-pf-title">
   <p id="blank-state-pf-info-{{$id}}" class="blank-state-pf-info" ng-if="$ctrl.config.info">
     {{$ctrl.config.info}}
   </p>
-  <p id="blank-state-pf-helpLink-{{$id}}" class="blank-state-pf-helpLink" ng-if="$ctrl.config.helpLink" ng-click="$ctrl.config.helpLink.urlAction()">
-    {{$ctrl.config.helpLink.label}} <a href="{{$ctrl.config.helpLink.url}}">{{$ctrl.config.helpLink.urlLabel}}</a>.
+  <p id="blank-state-pf-helpLink-{{$id}}" class="blank-state-pf-helpLink-label" ng-if="$ctrl.config.helpLink">
+    {{$ctrl.config.helpLink.label}}
+    <a ng-if="$ctrl.config.helpLink.url" class="blank-state-pf-helpLink" ng-click="$ctrl.config.helpLink.urlAction()" href="{{$ctrl.config.helpLink.url}}">{{$ctrl.config.helpLink.urlLabel}}</a>
+    <button ng-if="!$ctrl.config.helpLink.url" class="blank-state-pf-helpbutton btn-link" ng-click="$ctrl.config.helpLink.urlAction()">{{$ctrl.config.helpLink.urlLabel}}</button>.
```
jeff-phillips-18: Add `btn` here

src/views/views.less
```diff
@@ -108,6 +108,12 @@
   button {
     margin-right: 4px;
   }
+  .blank-state-pf-helpbutton {
+    border-left: 0;
+    border-right: 0;
```
jeff-phillips-18: Adding `btn` above lets you remove the border setting.
However, you'll need to set `vertical-align: baseline`

src/views/empty-state.html
dtaylor113: fixed, thanks

src/views/views.less
dtaylor113: fixed

amarie401:
Build says angular-ui-router has been renamed, should that be updated?

jeff-phillips-18:
We should address the ui-router issue as a separate PR.

Issue #671, opened 2017-11-06
pfDonutPctChart tooltip label shows % as well as units


cshinn:
The demo charts on [pfDonutPctChart](http://www.patternfly.org/angular-patternfly/#/api/patternfly.charts.component:pfDonutPctChart) display both a percent symbol and units when hovered as shown here. 
<img width="276" alt="screen shot 2017-11-06 at 3 03 58 pm" src="https://user-images.githubusercontent.com/3300756/32461510-1355c184-c304-11e7-9866-ff0e6d541d7a.png">

They should either show only %, or the tooltip should be changed to show absolute numbers with units rather than a percentage.

TheRealJon:
Adding a `tooltip` attribute that will allow for selection between "percent", "scalar", or "both".

PR #705, opened 2018-01-25
fix(donut-pct-chart): Add tooltip attribute

TheRealJon:
## Description
Add a `tooltip` attribute to the donut-pct-chart component which allows for a selection between 3 preset tooltip formats:
Fixes https://github.com/patternfly/angular-patternfly/issues/671

## Screen Shots

**Percent Format** `tooltip = 'percent'` (default)

image

**Amount Format** `tooltip = 'amount'`

image

**Both Format** `tooltip = 'both'`

image

## PR Checklist

- [x] Unit tests are included
- [X] A CSS rep (@cshinn) is assigned as a reviewer (if there are visual changes in the UI)

### DIFF
--- a/src/charts/donut/donut-pct-chart-component.js
+++ b/src/charts/donut/donut-pct-chart-component.js
@@ -2,6 +2,7 @@ angular.module('patternfly.charts').component('pfDonutPctChart', {
   bindings: {
     config: '<',
     data: '<',
+    tooltip: '<',
     chartHeight: '<?',
     centerLabel: '<?',
     onThresholdChange: '&'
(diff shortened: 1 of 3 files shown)


beanh66:
@TheRealJon Thanks, the percent format and the amount format look great. I was wondering if we could make a small adjustment to the format with both. I noticed in the PF Design it shows both with a gap between, something like this: 
donut
Is it possible to make this change to be more consistent? I attached the design example for reference as well. 
screen shot 2018-01-29 at 8 07 31 am

TheRealJon:
@beanh66 I adjusted the format to match your screen shot:

image

TheRealJon:
@dtaylor113 @beanh66 Could you approve again? I just added some unit tests for the tooltip.

PR #707, opened 2018-01-29
chore(pfFilterPanel): Update ngDoc to show how to prevent duplicate keywords in filter

### LINKED ISSUE #704: pfFilterPanel allows identical keyword to be entered multiple times
Enterning the exact same keyword in the pfFilterPanel multiple times will keep adding new items to the current filters list. Clicking the remove button on one of those items then removes all of them.

Discovered in OpenShift Console: 
image

dtaylor113:
## Description
Fixes #704

### DIFF
--- a/src/filters/examples/filter-panel.js
+++ b/src/filters/examples/filter-panel.js
@@ -235,13 +235,25 @@
 
         $scope.onKeywordKeyPress = function(keyEvent) {
           if (keyEvent.which === 13 && $scope.filterPanelModel[0].value.length > 0) {
-            // store new keywoard filter value in values array
-            $scope.filterPanelModel[0].values.push($scope.filterPanelModel[0].value);
+            var currentKeyword = $scope.filterPanelModel[0].value;
+            if(!keywordFilterExists(currentKeyword)) {
+              // store new keywoard filter value in values array
+              $scope.filterPanelModel[0].values.push(currentKeyword);
+              applyFilters();
+            }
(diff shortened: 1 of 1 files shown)


dtaylor113:
> LGTM. Probably should remove the case sensitive code that's commented out?

I don't think so, it's an ngDoc, so it's showing how you could implement both.

Do you happen to remember what OpenShift wanted?  Case insensitive or not?  -thanks

TheRealJon:
Could we make the `keywordFilterExists` function an attribute with a default value so that this feature is customizable?

dtaylor113:
> Could we make the keywordFilterExists function an attribute with a default value so that this feature is customizable?

The whole `pfFilterPanel`, from the contents of the panel controls, to the applying of filters, is handled by the consuming application.   I don't think PF determined a 'default' behavior for this.   Also, OpenShift would like the existing filter chip/tag to 'flash' when an existing keyword filter is entered twice.  -not sure this is a feature we want in angular-patternfly and/or PF core/design.

TheRealJon:
Yeah, I misunderstood what this PR is. This just gives an example of how to implement a pfFilterPanel with case sensitive/insensitive de-duplication logic, right?

dtaylor113:
@TheRealJon , yes, that's correct.   I would like to push this behavior into the component, but the contents of the filter panel is completely up to the consumer  -we can't guarantee a text/keyword filter will be implemented by the customer.

Issue #693, opened 2017-12-18
pfToolbar: Displaying An Empty Filter Result Set in ngDoc example


dtaylor113:
Design doc:  https://github.com/patternfly/patternfly-design/blob/master/pattern-library/forms-and-controls/filter/design/design.md#displaying-a-null-result-set

Currently in the ngDoc for pfToolbar, when there are no filter results, we incorrectly show the pfEmptyState component for 'no data'.  This should be changed to show a "null filter result set" alert as show in the design doc above.

PR #714, opened 2018-02-13
chore(pfToolbar): updated ngdoc with 'no filter results' empty state instead of 'No Items Available'

jcaianirh:
## Description
Fix Issue [https://github.com/patternfly/angular-patternfly/issues/693](https://github.com/patternfly/angular-patternfly/issues/693)
Add an empty state div with the messaging for 0 filtered results to differentiate that of an actual empty set based on the patternfly docs listed in the issue.

**Before**:

before

**After**:

after2

## PR Checklist

- [ ] Unit tests are included

### DIFF
--- a/src/toolbars/examples/toolbar.js
+++ b/src/toolbars/examples/toolbar.js
@@ -84,7 +84,7 @@
         <pf-list-view config="listConfig"
                       page-config="pageConfig"
                       items="items"
-                      empty-state-config="emptyStateConfig">
+                      empty-state-config="items.length === 0 && filterConfig.appliedFilters.length > 0 ? noItemsConfig : emptyStateConfig">
           <div class="list-view-pf-description">
             <div class="list-group-item-heading">
               {{item.name}}
@@ -107,7 +107,7 @@
         <pf-card-view config="listConfig"
                       page-config="pageConfig"
                       items="items"
-                      empty-state-config="emptyStateConfig">
+                      empty-state-config="items.length === 0 && filterConfig.appliedFilters.length > 0 ? noItemsConfig : emptyStateConfig">
           <div class="col-md-12">
             <span>{{item.name}}</span>
           </div>
@@ -124,7 +124,7 @@
                        page-config="pageConfig"
                        columns="columns"
                        items="items"
-                       empty-state-config="emptyStateConfig">
+                       empty-state-config="items.length === 0 && filterConfig.appliedFilters.length > 0 ? noItemsConfig : emptyStateConfig">
         </pf-table-view>
       </div>
       <hr class="col-md-12">
@@ -285,6 +285,11 @@
         } else {
           $scope.items = $scope.allItems;
         }
+      };
+
+      var clearFilters = function() {
+        filterChange([]);
+        $scope.filterConfig.appliedFilters = [];
       };
 
       var filterChange = function (filters) {
@@ -485,6 +490,15 @@
            label: 'For more information please see',
            urlLabel: 'pfExample',
            url : '#/api/patternfly.views.component:pfEmptyState'
+        }
+      };
+
+      $scope.noItemsConfig = {
+        title: 'No Results Match the Filter Criteria',
+        info: 'The active filters are hiding all items.',
+        helpLink: {
+          urlLabel: 'Clear All Filters',
+          urlAction: clearFilters
         }
       };
 


src/toolbars/examples/toolbar.js
```diff
@@ -488,6 +520,15 @@
         }
       };
 
+      $scope.noItemsConfig = {
+        title: 'No results match.',
```
dtaylor113: Pls change to match verbiage in [pf core](http://www.patternfly.org/pattern-library/forms-and-controls/filter/#design) (null result set at bottom of page) "No results match the filter criteria"  -thanks

dtaylor113:
In Table View, the 'no items match filter' empty state is wider than the toolbar:

image

src/toolbars/examples/toolbar.js
```diff
@@ -104,7 +113,16 @@
...
       </div>
       <div class="col-md-12" ng-if="viewType == 'cardView' && showComponent">
-        <pf-card-view config="listConfig"
+        <div ng-if="items.length === 0">
+          <div class="pf-empty-state">
+            <div ng-if="filterConfig.appliedFilters.length > 0">
+              <pf-empty-state config="noItemsConfig"></pf-empty-state>
+            </div>
+          </div>
+        </div>
```
dtaylor113: Rather than repeating/adding `<pf-empty-state/>`'s to the demo HTML, since we are already passing in `empty-state-config`'s to each of the List,Card,Table Views, you can just dynamically set it:

```
<pf-list-view config="listConfig"
                      page-config="pageConfig"
                      items="items"
                      empty-state-config="items.length === 0 ? noItemsConfig : emptyStateConfig">
```
Same for the Card and Table views.   Since this will use the `<pf-empty-state/>' embedded in each of the [views](https://github.com/patternfly/angular-patternfly/blob/master/src/views/listview/list-view.html#L75) this will also fix the TableView's 'no results' empty state container being wider than the toolbar.

jcaianirh: good idea

src/toolbars/examples/toolbar.js
jcaianirh: Completed, thanks!

src/toolbars/examples/toolbar.js
jcaianirh: Completed, thanks!

jcaianirh:
Made requested updates by @dtaylor113

src/toolbars/examples/toolbar.js
```diff
@@ -81,10 +81,11 @@
         </pf-toolbar>
       </div>
       <div class="col-md-12" ng-if="viewType == 'listView' && showComponent">
-        <pf-list-view config="listConfig"
+        <pf-list-view
```
dtaylor113: nit.  not sure why config couldn't be inline with `pf-list-view`, seems like an extra line/whitespace to me.

jcaianirh: @dtaylor113 I guess a difference in style. I'll update it to reflect the style of the rest of the file.

jcaianirh: @dtaylor113 pushed config line changes

dtaylor113:
Hi @cshinn, @beanh66,  this PR updates a angular-patternfly example which shows a core patternfly 'pattern' -List/Card/Table view with 'no results match' empty state.   It would be great to get designer/styling feedback  -thanks

beanh66:
@dtaylor113 Interestingly we also just got something similar updated in the design repo: 
https://github.com/patternfly/patternfly-design/pull/567
screen shot 2018-02-14 at 2 26 45 pm

So I think if we want to be consistent, it would need to say "No Results Match the Filter Criteria" (no period needed). @cshinn do you agree?

amarie401:
I believe @jcaianirh has made that verbiage change to his code @beanh66. @jcaianirh could you provide an updated screenshot?

amarie401:
@jcaianirh could you remove the period and provide an updated screenshot? Thanks!

jcaianirh:
@amarie401 @beanh66 fixed capitalization, and removed the period for 'No Results Match the Filter Criteria' title string. Updated screenshot as well.

dtaylor113:
Hmm...according to @beanh66 latest design spec, think the link needs to be 'Clear All Filters'

amarie401:
Ahhh you're right, good catch @dtaylor113.

jcaianirh:
@dtaylor113 @amarie401 Thanks for the input. I updated the link text to 'Clear All Filters', and the commit message and pr message to use the chore indicator keyword because there is no effect on consumers.

PR #718, opened 2018-02-19
fix(pfTableView): Fixed sort col when show/hiding checkbox col

### LINKED ISSUE #715: pfTableView: default sorting doesn't change based on visibility of checkboxes column
**Current behavior**
When the table shows checkboxes, the default sorting is by the second column.
When the table doesn't show checkboxes, the default sorting is by the **second** column.

**Expected behavior**
When the table shows checkboxes, the default sorting is by the second column.
When the table doesn't show checkboxes, the default sorting is by the **first** column.

dtaylor113:
## Description
Fixes #715  
@akieling

Previously, the sort column was 'hard coded' regardless of whether the first checkbox column was shown or hidden:

**Initial View:**
image

**After Turn Off Checkbox Column:**
image

**After Fix, sort column does not change:**

image
image

* Tested and Works with multiple column sorts as well.

## PR Checklist

- [ ] Unit tests are included

### DIFF
--- a/src/table/tableview/examples/table-view-basic.js
+++ b/src/table/tableview/examples/table-view-basic.js
@@ -103,7 +103,9 @@
   angular.module('patternfly.tableview.demo').controller('TableCtrl', ['$scope', '$timeout', 'itemsService',
   function ($scope, $timeout, itemsService) {
           $scope.dtOptions = {
-            order: [[2, "asc"]],
+            // order column(s) should NOT account for 1st checkbox column, table component will adjust col. numbers accordingly
+            // Sort by City, then Name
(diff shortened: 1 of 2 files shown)


jcaianirh:
LGTM: This looks great, and def solves the problem with the checkbox display on or off. If we move to a design of user customizable columns (whether it be shifting order or toggling view state) we may need to use a key to id a column instead of the column order.

dtaylor113:
> If we move to a design of user customizable columns (whether it be shifting order or toggling view state) we may need to use a key to id a column instead of the column order.

Agree, that would be more flexible.  -thanks

PR #719, opened 2018-02-21
add pfBootstrapDatepicker

### LINKED ISSUE #711: angular-ui-bootstrap Datepicker component does not include PatternFly styling
Apparently pfDateTimepicker was dropped from the current PatternFly release in favor of a recommendation to use the angular-ui-boostrap Datepicker component (see http://angular-ui.github.io/bootstrap/versioned-docs/1.0.0/). While this meets the functional requirements for selecting dates, the styling of this component is inconsistent with the PatternFly design here: http://www.patternfly.org/pattern-library/forms-and-controls/datepicker/#design

The angular-ui-bootstrap component should be restyled to be more consistent with the PatternFly design.

nicolethoen:
## Description
fixes #711 
Styled the Angular UI datepicker so that it looks more like the patternfly datepicker.
Mostly removing the week numbers from the beginning of each week and styling the various buttons.
The user can also still pass dateOptions to the Angular UI datepicker to customize its behavior.

screen shot 2018-02-21 at 11 47 25 am 2
screen shot 2018-02-22 at 2 24 24 pm 2
screen shot 2018-02-21 at 11 48 10 am 2

## PR Checklist

- [x] Unit tests are included

### DIFF
--- a/src/datepicker/bootstrap-datepicker.component.js
+++ b/src/datepicker/bootstrap-datepicker.component.js
@@ -1,0 +1,38 @@
+angular.module('patternfly.datepicker').component('pfBootstrapDatepicker', {
+  bindings: {
+    date: '<',
+    format: '@?',
+    dateOptions: '<?',
+    isOpen: '<?',
+    popupPlacement: '@?'
+  },
+  templateUrl: 'datepicker/datepicker.html',
+  controller: function () {
+    'use strict';
+
+    var ctrl = this;
+
+    ctrl.defaultDateOptions = {
+      showWeeks : false,
+      formatDay : "d"
+    };
+    ctrl.defaultIsOpen = false;
+
+    ctrl.$onInit = function () {
+      ctrl.format = "MM/dd/yyyy";
+      ctrl.showButtonBar = true;
+      ctrl.popupPlacement = "auto bottom-left";
+
+      if (angular.isUndefined(ctrl.dateOptions)) {
+        ctrl.dateOptions = {};
+      }
+      _.defaults(ctrl.dateOptions, ctrl.defaultDateOptions);
+      _.defaults(ctrl.isOpen, ctrl.defaultIsOpen);
+    };
+
+    ctrl.$onChanges = function (changes) {
+      _.defaults(ctrl.isOpen, ctrl.defaultIsOpen);
+    };
+
+  }
+});
--- a/src/datepicker/datepicker.less
+++ b/src/datepicker/datepicker.less
@@ -1,0 +1,120 @@
+.uib-datepicker-popup {
+  padding: 4px;
+
+  *:focus {
+    outline: none;
+  }
+
+  button {
+    background: @color-pf-white;
+    border: none;
+    box-shadow: none;
+  }
+
+  th {
+    height: 30px;
+  }
+
+  .uib-title {
+    font-size: 14px;
+    font-weight: 500;
+    padding: 5px;
+
+    strong {
+      font-weight: normal;
+    }
+  }
+
+  .uib-left {
+    height: 30px;
+
+    .glyphicon {
+      display: none;
+    }
+
+    &::before {
+      content: "\00AB";
+    }
+  }
+
+  .uib-right {
+    height: 30px;
+
+    .glyphicon {
+      display: none;
+    }
+
+    &::before {
+      content: "\00BB";
+    }
+  }
+
+  .uib-day {
+
+    button.btn.btn-default {
+      height: 30px;
+      width: 30px;
+
+      &:hover {
+        background: @color-pf-blue-50;
+      }
+
+      &.active {
+        background: @color-pf-blue-400;
+        border-color: @color-pf-blue-600;
+        color: @color-pf-white;
+        box-shadow: none;
+        padding: 0;
+
+        .text-info { // this is today's date
+          color: @color-pf-white;
+        }
+      }
+
+      &:not(.active){
+        .text-info { // this is today's date
+          color: @color-pf-black-800;
+          background: @color-pf-gold-200;
+          height: 30px;
+          width: 30px;
+          padding: 7px;
+          display: inline-block;
+          margin: -1px -7px -2px -7px;
+
+          &:hover {
+            background: @color-pf-blue-50;
+          }
+        }
+      }
+    }
+
+  }
+
+  .uib-button-bar {
+    padding: 0;
+
+    .btn-group {
+      width: 100%;
+
+      .uib-clear {
+        display: none;
+      }
+
+      .uib-datepicker-current {
+        color: @color-pf-black;
+        width: 100%;
+        font-size: 14px;
+        font-weight: 500;
+        height: 30px;
+
+        &:hover {
+          background: @color-pf-blue-50;
+        }
+      }
+    }
+
+    .uib-close {
+      display: none;
+    }
+  }
+}
--- a/src/datepicker/datepicker.module.js
+++ b/src/datepicker/datepicker.module.js
@@ -1,0 +1,9 @@
+/**
+ * @name  patternfly datepicker
+ *
+ * @description
+ *   Datepicker module for patternfly.
+ *
+ */
+angular.module('patternfly.datepicker', ['ui.bootstrap']);
+
--- a/src/datepicker/examples/bootstrap.datepicker.js
+++ b/src/datepicker/examples/bootstrap.datepicker.js
@@ -1,0 +1,50 @@
+/**
+ * @ngdoc directive
+ * @name patternfly.datepicker.componenet:pfBootstrapDatepicker
+ * @element pf-bootstrap-datepicker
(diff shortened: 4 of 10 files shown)


beanh66:
@nicolethoen A couple of minor things:
- Seems like we need more spacing around the month name
- The fa-angle-double-left and right icons should be used for advancing through months
- The selected date should be shown in this format: 02/21/2018
- Today should not be bold and should match the font size of "February"

I can't tell by your images if you have hover states on the month selection and the "Today" selection but those should match the [core implementation in PatternFly](http://www.patternfly.org/pattern-library/forms-and-controls/datepicker/#code) if possible.

screen shot 2018-02-21 at 1 43 43 pm

Thanks!

src/datepicker/datepicker.module.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * @name  patternfly datepicker
+ *
+ * @description
+ *   Select module for patternfly.
```
jcaianirh: Add description apropos for the control, eg. 

Datepicker module for patternfly

src/datepicker/examples/bootstrap.datepicker.js
```diff
@@ -0,0 +1,52 @@
+/**
+ * @ngdoc directive
+ * @name patternfly.datepicker.componenet:pfBootstrapDatepicker
+ * @element pf-bootstrap-datepicker
+ *
+ * @param {date} date must be a Javascript Date - to be displayed in the input. Can be left empty.
+ * @param {string} format default date format for displayed dates
+ * @param {object} dateOptions an object to configure the uib-datepicker in one place (optional)
```
jcaianirh: As this is an object, some reference material might be helpful. See other object descriptions which describe the format/syntax of the options allowed for date.

src/datepicker/examples/bootstrap.datepicker.js
```diff
@@ -0,0 +1,52 @@
...
+ * @param {boolean} isOpen whether or not to show the datepicker (default false) (optional)
+ * @param {object} popupPlacement Placement of the datepicker popup relative to the input element. Passing in 'auto' separated by a space bfore the placement will enable auto positioning. (optional)
```
jcaianirh: Same as above vis a vis the object description, but additionally, with this option it might be nice to see the allowance for different placements in the demo area below. ie a radio button for say above/below.

src/datepicker/examples/bootstrap.datepicker.js
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @description
+ * A wrapper for the Angular UI {@link http://angular-ui.github.io/bootstrap/#!#datepickerPopup datepicker}.
```
dtaylor113: That link isn't  working for me, it goes to the top of Angular UI Bootstrap page.   I think `http://angular-ui.github.io/bootstrap/#datepickerPopup` will be enough to link-n-launch to the correct part of the page.
Hmm..still not working in some cases.  Funny, even their [GitHub](https://github.com/angular-ui/bootstrap/tree/master/src/datepickerPopup) page's 'here' and 'inline datepicker' links do not work correctly for me.

src/datepicker/datepicker.less
```diff
@@ -0,0 +1,66 @@
...
+
+    button.btn.btn-default {
+
+      &:hover {
+        background: @color-pf-blue-50;
+      }
+
+      &.active {
+        background: @color-pf-blue-400;
+        border-color: #003d5f;
```
jeff-phillips-18: This should be a color from the patternfly color palette.

src/datepicker/datepicker.module.js
nicolethoen: nice catch. and sublime use of apropos.

src/datepicker/examples/bootstrap.datepicker.js
nicolethoen: yeah - i've noticed the same problem as I've tried linking to the param documentation in response to Joe's comments... Not sure what to do about that.

src/datepicker/bootstrap-datepicker.component.js
```diff
@@ -0,0 +1,27 @@
+angular.module('patternfly.datepicker').component('pfBootstrapDatepicker', {
+  bindings: {
+    format: '@',
+    date: '<',
+    dateOptions: '@',
```
dtaylor113: `dateOptions` is an object, not a string, so it needs to be defined as `dateOptions: '<'`.   You'll notice when you make this change, your overrides in your example no longer work Ex:
```
$scope.dateOptions = {
        showWeeks : true
      };
```
:-(

This is because the `$onInit()` is actually called after the objects from the example are bound.   Try adding this to your component:
```
    ctrl.$onChanges = function (changesObj) {
      console.log("on changes");
    };
```
Then in the browser, source tab, open up angular-patternfly.js, find the component, set breakpoints in $onInit() and $onChanges().  You may want to remove the first datepicker example, just to see how the lifecycle of one example works.   You'll see that $onChanges() is called first with the objects from the example, then $onInit() is called afterwards  -in this case overriding the `$scope.dateOptions` set by the user/example.  

We typically have some sort of `if (!angular.isDefined(...)..`  or 'defaulting' operation in the $onInit() to prevent overriding.

Here's some [code](https://github.com/patternfly/angular-patternfly/blob/master/src/toolbars/toolbar-component.js#L15) which has some examples of $onInit(), $onChanges, and even $doCheck for deep comparisions.

Here is another [code example](https://github.com/patternfly/angular-patternfly/blob/master/src/table/tableview/table-view.component.js#L120) where we use lodash `._defaults(...)` to handle default props and user defined props.

dtaylor113: ^^ I'd probably stick with the last example and lodash `._defaults(...)` it's a nice solution.

dtaylor113:
You also have to change your commit msg to  be something like [this](https://github.com/patternfly/angular-patternfly#feature-release).  I think `feat(pfBootstrapDatePicker):  Initial Implementation`

dtaylor113:
Hi @nicolethoen, latest code looks good!  Here are my findings on updating attributes:
```
$timeout(function () {
  $scope.dateOptions.showWeeks = false;  // doesn't take effect on open popup, but does if re-opened
  $scope.popupPlacement = "upper-right"; // doesn't take effect on open popup, but does if re-opened
  // These do take effect immediately in open popup. 
  $scope.isOpen = "";
  $scope.date = new Date("Dec 1, 2000");
  $scope.format1 = "MM-dd-yy";
}, 2000);
```
So, for this particular component/usecase I think these are acceptable.

dtaylor113:
Can you pls look into changing the selected color to match pf design:

**PF Design:**
image

**Current Impl:**
image

The pf core example is showing the correct color:    http://www.patternfly.org/pattern-library/forms-and-controls/datepicker/#code

test/datepicker/bootstrap-datepicker.spec.js
```diff
@@ -0,0 +1,78 @@
+describe('Component:  pfBootstrapDatepicker', function () {
```
dtaylor113: Appreciate the unit tests!  :+1:

nicolethoen:
> The pf core example is showing the correct color: http://www.patternfly.org/pattern-library/forms-and-controls/datepicker/#code

so, the yellow color here is actually highlighting today's date - not the selected date. 

For some reason, the angular ui component doesn't distinguish between todays date and your selected date. So it will show today's date as if it were selected in blue until you select a date and it will highlight that in blue. I've had a hard time getting the yellow to stick to todays date, but I'll keep trying.

dtaylor113:
> I've had a hard time getting the yellow to stick to todays date, but I'll keep trying.

Ok, sounds good.  Not a show stopper :-)

dtaylor113:
> @nicolethoen The visual changes look good based on your updated screenshot. Thanks!

Hi @beanh66, could you please approve this PR if satisfied with the visuals.  -thanks!

@jeff-phillips-18, pls review again, I think your issue(s) have been addressed  -thanks!

beanh66:
@dtaylor113 @nicolethoen Yes this satisfies the visuals from what I can tell. I can't see if there are hover colors on the month and the current date. Are those matching the design? And does the month selection open up to a display of all months? Not sure if there is somewhere I can check out the current implementation. If not, the screenshots work. Thanks!

nicolethoen:
@beanh66 @dtaylor113 

Here i've tried capturing the latest changes with hover states.
screen shot 2018-02-23 at 3 02 18 pm
screen shot 2018-02-23 at 3 02 29 pm
screen shot 2018-02-23 at 3 02 41 pm

beanh66:
Thanks @nicolethoen! Hover states look great. It looks like the version with years or months should have more space around them and in the PF core implementation the months are abbreviated. Is that possible here? Posted a screenshot below: 
screen shot 2018-02-26 at 8 10 23 am

PR #724, opened 2018-03-05
Modal overlay

### LINKED ISSUE #673: Provide implementation for the Modal Overlay
Provide implementation for the modal overlay as documented here: http://www.patternfly.org/pattern-library/forms-and-controls/modal-overlay/

Note that @jeff-phillips-18 implemented this in OpenShift, but this has not yet been contributed back to PatternFly.  Also, I think he has some suggestions re: implementation that should be considered.

nicolethoen:
## Description
fixes #673 
Builds a Patternfly modal overlay component to be used by the Angular UI $uibModal service to create a modal overlay consistent with Patternfly design and behavior. The modal slides in and out from the top. 

screen shot 2018-03-05 at 2 20 18 pm
screen shot 2018-03-05 at 2 20 32 pm
screen shot 2018-03-05 at 2 21 18 pm

The user can still use the Angular UI modal and build a custom component rather that using the Angular Patternfly component.

## PR Checklist

- [x] Unit tests are included


src/modals/modal-overlay/modal-overlay.html
```diff
@@ -0,0 +1,15 @@
+<div class="modal-overlay" ng-attr-id="{{$ctrl.modalId}}" tabindex="-1" aria-labelledby="$ctrl.titleId" aria-hidden="true">
+  <div class="modal-header">
+    <button type="button" class="close" ng-click="$ctrl.cancel()" aria-hidden="true" aria-label="Close" ng-if="!$ctrl.hideCloseIcon">
+      <span class="pficon pficon-close"></span>
+    </button>
+    <h4 class="modal-title" ng-attr-id="{{$ctrl.titleId}}">{{$ctrl.modalTitle}}</h4>
+  </div>
+  <div class="modal-body" ng-bind-html="$ctrl.trustAsHtml($ctrl.modalBody)"></div>
```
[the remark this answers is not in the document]
nicolethoen: ah yes. on it.

dtaylor113:
Hi @nicolethoen, Looks good.   I just noticed when I click on the darkened background it closes the modal.  I don't think this is the expected behavior: 

"Background: When the modal overlay appears, the background should darken to bring attention to the primary content. While the rest of the page is no longer accessible, the background should still be legible to remain in context for the user."

http://www.patternfly.org/pattern-library/forms-and-controls/modal-overlay/#design

src/modals/modal-overlay/modal-overlay.js
```diff
@@ -0,0 +1,158 @@
...
+ *   <li>modalTitle - string modal title displayed in its header</li>
+ *   <li>titleId (optional) - string id of the title of the modal. Customizable so multiple modals can have unique ids.
+ * Set to 'modalTitle' by default</li>
+ *   <li>hideCloseIcon (optional) - boolean hide-close-icon Hides the 'x' close button, default is false</li>
+ *   <li>modalBody - html string to display in the modal</li>
+ *   <li>actionButtons - Array of button objects. Each button object can have the following properties:
+ *      <ul style='list-style-type: none'>
+ *        <li>.label      - the text to display on the button
+ *        <li>.class      - (optional) classes to add to the button
+ *        <li>.disabled   - (optional) boolean true if the button should be disabled by default
```
dtaylor113: We should make this consistent with params in other components.  Pls see: https://www.patternfly.org/angular-patternfly/#/api/patternfly.toolbars.component:pfToolbar

Params like 'isDisabled', 'actionFn', and 'onClick'

dtaylor113:
I think we need a 'max-height' for the modal to meet the 'Long Example', see bottom of:  http://www.patternfly.org/pattern-library/forms-and-controls/modal-overlay/#design.  

modal_overlay

Think we need an internal scrollbar for the main modal content, in order to always show the bottom buttons of the modal. -thanks

test/karma.conf.js
```diff
@@ -11,12 +11,14 @@ module.exports = function (config) {
       'node_modules/jquery/dist/jquery.js',
       'node_modules/datatables.net/js/jquery.dataTables.js',
       'node_modules/moment/moment.js',
+      'node_modules/bootstrap/js/modal.js',
```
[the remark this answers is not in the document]
dtaylor113: yeah, not sure, I do see unit tests (which is always appreciated :-)  but they continue to work if I take out that line.

beanh66:
@nicolethoen can we use Save instead of Okay for both primary buttons and remove the "test" button? Also wondering if the two column layout example can be tweaked a bit to show some form input (radio buttons or something simple) on the right column with the paragraph only on the left. Let me know if that works!

test/karma.conf.js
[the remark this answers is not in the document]
nicolethoen: yeah I'll take it out.

nicolethoen:
> http://www.patternfly.org/pattern-library/forms-and-controls/modal-overlay/#design

So, in the patternfly example: [https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/modals.html](here)
And in the react implementation: [https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?knob-Right%20aligned=false&knob-fluid=true&knob-Stacked=false&selectedKind=Modal%20Overlay&selectedStory=Modal&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs](here)
it seems that dismissing it by clicking the background is the expected behavior...

It is also the default behavior for the angular UI modal.

jeff-phillips-18:
It should be configurable here whether or not to dismiss on background click.

nicolethoen:
> It should be configurable here whether or not to dismiss on background click.

but I am not convinced that it should be. then this component would behave differently than the patternfly modal

dtaylor113:
> I am not convinced that it should be. then this component would behave differently than the patternfly modal

I think we need a ruling from PatternFly designers.  @cshinn, @beanh66   -thanks

cshinn:
hmmm, I think dismissing on background click could be dangerous in cases where the user has to enter and save a lot of information. It makes sense when the modal is primarily informational (like the about modal) but when it's a dialog, you could easily lose all your progress. That may be something worth bringing up in PF core as well. I also agree with @beanh66 that "save" is a more informative primary button than "ok" for a dialogue case.

beanh66:
I agree for the most part our modals do not close with a background click, so our examples should reflect that common use case. 

With that said, I agree with @jeff-phillips-18 it should be configurable because we do have rare use cases where clicking the background will dismiss a modal. For example, when we show a modal to tell users they have been inactive for too long and their session will end - clicking the background will dismiss that modal and allow a user to continue working in an active session.

jeff-phillips-18:
Applications should be able to decide. There is a callback that can be handled that can prevent closing if there are, say, unsaved changes.

nicolethoen:
screen shot 2018-03-09 at 9 27 46 am
screen shot 2018-03-09 at 9 27 12 am
screen shot 2018-03-09 at 9 27 05 am

src/modals/examples/modal-overlay.js
```diff
@@ -0,0 +1,138 @@
...
+ * @example
+ <example module="patternfly.modals">
+
+ <file name="index.html">
+ <div ng-controller="DemoModalOverlayCtrl">
+    <button ng-click="open()" class="btn btn-default">Launch Modal Overlay</button>
+    <pf-modal-overlay is-open="isOpen"
+          on-close="onClose()"
+          modal-id="modalId"
+          modal-body-path="modalBodyPath"
```
dtaylor113: Hi, is it possible to use angular's `transclusion` here?  See [pfFilterPanel](https://www.patternfly.org/angular-patternfly/#/api/patternfly.filters.component:pfFilterPanel) or [pfListView](https://www.patternfly.org/angular-patternfly/#/api/patternfly.views.component:pfListView) for examples on how we use transclusion.    [ng-transclude](https://github.com/patternfly/angular-patternfly/blob/master/src/filters/filter-panel/filter-panel.html#L7) in filter-panel.html.

src/modals/examples/modal-overlay.js
```diff
@@ -0,0 +1,138 @@
...
+          },
+          {
+            label: "OK",
+            class: "btn-primary"
+          }];
+ });
+
+ </file>
+
+ <file name="pf-modal-body.html">
```
dtaylor113: I guess I'm expecting to see this HTML in between `<pf-modal-overlay><contents of pf-modal-body.html></pf-modal-overlay>`, and have the `ng-transclude` attribute in the HTML template for the component.

dtaylor113:
Ideally, I'd like to see for the first example, which uses a form, that the first input "Field One" be required and the [Ok] button disabled when Field One is empty, and enabled when a value is entered.  I know that is going to be the first question app devs are going to ask about using this component.  -thanks!

dtaylor113:
It's interesting how OpenShift uses [$uibModal.open](https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/fromFile.js#L53).  Wonder if we just had our own `pfModal.open(...` which would take in all those params and use $uibModal, but just overwrite angular-bootstrap's modal css to match pf core.
This way we could avoid `transclusion` and just update the example to use an example `templateUrl' and `Controller`

src/modals/modal-overlay/modal-overlay.component.js
```diff
@@ -0,0 +1,122 @@
...
+      close: "&onClose",
+      isOpen: '<?'
+    },
+    templateUrl: 'modals/modal-overlay/modal-overlay.html',
+    controller: function ( $uibModal ) {
+      'use strict';
+
+      var ctrl = this;
+
+      // The ui-bootstrap modal only supports either template or templateUrl as a way to specify the content.
```
dtaylor113: Is this comment relevant anymore?  I think you removed transclusion support.

src/modals/examples/modal-overlay.js
```diff
@@ -0,0 +1,179 @@
...
+ * <li>.actionFn     - (optional) use defined function to call when the button is clicked
+ * <li>.isDisabled   - (optional) boolean true if the button should be disabled by default
+ * <li>.isCancel     - (optional) boolean true is the button should cancel and dismiss the modal
+ * </ul>
+ *
+ * @example
+ <example module="patternfly.modals">
+
+ <file name="index.html">
+ <div ng-controller="DemoModalOverlayCtrl">
```
dtaylor113: In order to get the Actions textarea to display correctly.  Add `class="example-container"` here

src/modals/examples/modal-overlay.js
```diff
@@ -0,0 +1,179 @@
...
+         backdrop-close="backdropClose"
+         title-id="titleId2"
+         modal-body-template="template2"
+         action-buttons="actionButtons2">
+    </pf-modal-overlay>
+   <div>
+      <label class="actions-label">Actions: </label>
+   </div>
+   <div>
+      <textarea rows="3">{{actionsText}}</textarea>
```
dtaylor113: In order to get the Actions textarea to display correctly.  Add `class="col-md-12"` here

src/modals/examples/modal-overlay.js
```diff
@@ -0,0 +1,179 @@
...
+ * away from the current page. It utilizes the {@link https://angular-ui.github.io/bootstrap/#modal Angular UI modal}
+ * and Patternfly styling. Use cases for Modal Overlays can vary a great deal, but some examples include the following:
+ * <ul>
+ * <li>To remind or prompt users
+ * <li>To load dialogs that require user input before advancing
+ * <li>To load tasks which require a user’s full attention, such as stepping through a wizard flow
+ * <li>To present important information or warnings
+ * </ul>
+ *
+ * @param {boolean} isOpen Flag indicating that the modal should be opened
```
dtaylor113: I realize this is how aboutModal works, but I keep tripping up on this.  `isOpen` or any `isFooBar` is suppose to just return a boolean True or False; app devs don't typically expect setting a `isWhatever` actually does an action.
Pls change to `showModal` to be clearer of it's usage.  -thanks

dtaylor113: ...and even `showModal` isn't implemented quite correctly as `showModal = false` won't actually hide the modal...although I think that would be easy to implement, however I don't think it's it's a common usecase.

dtaylor113:
> Applications should be able to decide [backdropClose]. There is a callback that can be handled that can prevent closing if there are, say, unsaved changes.

Hi, do we handle this usecase?  For the example, when user clicks on backdrop of the info example, capture/execute the example's `close(...)` function and just log to Actions textarea: 'User clicked on backdrop'.  
Ideally, it would be great to include a `[ ] I agree` checkbox on the info form, and if user clicks on backdrop without checking it, a red warning msg appears in the modal saying they must check the 'I agree' checkbox.

dtaylor113:
> Applications should be able to decide [backdropClose]. There is a callback that can be handled that can prevent closing if there are, say, unsaved changes.

Hi, I implemented a possible solution for 'backdrop-validation' [here](https://github.com/dtaylor113/angular-patternfly/commit/229b532da78aad5fab3a295b0eb4f83d01a02cad)
image

src/modals/examples/modal-overlay.js
```diff
@@ -0,0 +1,203 @@
...
+ * <ul>
+ * <li>To remind or prompt users
+ * <li>To load dialogs that require user input before advancing
+ * <li>To load tasks which require a user’s full attention, such as stepping through a wizard flow
+ * <li>To present important information or warnings
+ * </ul>
+ *
+ * @param {boolean} showModal Flag for setting modal to hide/show by default
+ * @param {function} onClose Function to call when the modal is closed<br/>
+ * Note: since 'on-close' is a '&', param passed must be named 'dismissCause'
```
dtaylor113: I think "Note: the parameter passed to the 'on-close' function must be named 'dismissCause'"  ...would be fine.

src/modals/modal-overlay/modal-overlay.component.js
```diff
@@ -0,0 +1,136 @@
...
+        ctrl.dismiss({$value: 'cancel'});
+      };
+
+      $scope.$on('modal.closing', function(event, reason, closed) {
+        if (reason === "backdrop click" || reason === "escape key press") {
+          event.preventDefault();
+          if ( typeof ctrl.onBackgroundClick === "function" && ctrl.onBackgroundClick()) {
+            ctrl.dismiss({$value: 'backdropClick'});
+          } else {
+            $scope.$digest();
```
dtaylor113: Pls add comment `onBackgroundClick' returned false, may have changed ctrl vars; execute $digest to refresh view`

src/modals/examples/modal-overlay.js
```diff
@@ -0,0 +1,203 @@
...
+          {
+            label: "Save",
+            class: "btn-primary custom-class",
+            actionFn: function() {
+                $scope.actionsText = "inputs {" +
+                    "\n    first: " + $scope.formScope.inputs.first +
+                    "\n    second: " + $scope.formScope.inputs.second +
+                    "\n    third: " + $scope.formScope.inputs.third +
+                    "\n}" + $scope.actionsText;
+                return true;
```
dtaylor113: Sorry, my bad, my poor attempt to do post submission validation, which isn't the correct pattern.   actionFn's do not need to return T/F.

src/modals/examples/modal-overlay.js
```diff
@@ -0,0 +1,203 @@
...
+      $scope.actionButtons2 = [
+          {
+            label: "Cancel",
+            isCancel: true
+          },
+          {
+            label: "OK",
+            class: "btn-primary",
+            actionFn: function() {
+              return true;
```
dtaylor113: Can remove

src/modals/modal-overlay/modal-overlay.component.js
```diff
@@ -0,0 +1,136 @@
...
+      ctrl.titleId = ctrl.resolve.titleId || "modalTitle";
+      ctrl.modalTitle = ctrl.resolve.modalTitle;
+      ctrl.hideCloseIcon = ctrl.resolve.hideCloseIcon || false;
+      ctrl.modalBodyTemplate = ctrl.resolve.modalBodyTemplate;
+      ctrl.modalBodyScope = ctrl.resolve.modalBodyScope;
+      ctrl.actionButtons = ctrl.resolve.actionButtons;
+      ctrl.isForm = ctrl.resolve.isForm;
+      ctrl.onBackgroundClick = ctrl.resolve.onBackgroundClick;
+
+      ctrl.ok = function (label, actionFn) {
```
dtaylor113: This can just be:
```
ctrl.ok = function (label, actionFn) {
        if (typeof actionFn === "function") {
          actionFn();
        }
        ctrl.close({$value: label});
      };
```

src/modals/examples/modal-overlay.js
```diff
@@ -0,0 +1,203 @@
...
+ * @param {function} onClose Function to call when the modal is closed<br/>
+ * Note: since 'on-close' is a '&', param passed must be named 'dismissCause'
+ * @param {string} modalId Id of the modal
+ * @param {string} modalTitle The title of the modal displayed in the header
+ * @param {string} modalBodyTemplate Path to html template for modal body
+ * @param {array} actionButtons array of button objects. Each button can have the following properties:<br/>
+ * <ul style='list-style-type: none'>
+ * <li>.label        - the text to display on the button
+ * <li>.class        - (optional) classes to add to the button
+ * <li>.actionFn     - (optional) user defined function to call when the button is clicked. Should return true to close modal, otherwise should return false.
```
dtaylor113: Pls remove: '...Should return true to close modal, otherwise should return false.'.   I don't think we need to support validation after Ok button is clicked, if modal is invalid, Ok button should be disabled.

nicolethoen:
@dtaylor113 Is there anything else I should do? Or can this be merged?

src/modals/examples/modal-overlay.js
```diff
@@ -0,0 +1,199 @@
...
+ * <ul>
+ * <li>To remind or prompt users
+ * <li>To load dialogs that require user input before advancing
+ * <li>To load tasks which require a user’s full attention, such as stepping through a wizard flow
+ * <li>To present important information or warnings
+ * </ul>
+ *
+ * @param {boolean} showModal Flag for setting modal to hide/show by default
+ * @param {function} onClose Function to call when the modal is closed<br/>
+ * Note: the parameter passed to the on-close function must be named 'dismissCause'
```
jeff-phillips-18: why?

src/modals/modal-overlay/modal-overlay.component.js
```diff
@@ -0,0 +1,131 @@
...
+              return ctrl.modalBodyScope;
+            },
+            actionButtons: function () {
+              return ctrl.actionButtons;
+            }
+          }
+        })
+          .result.then(
+          function (dismissCause) {
+            ctrl.close({'dismissCause': dismissCause}); // closed
```
jeff-phillips-18: Is there a reason to use a named parameter here?

dtaylor113:
@jeff-phillips-18 , I'll be merging this soon, last chance to comment, thanks :-)

styles/angular-patternfly.less
```diff
@@ -15,4 +15,4 @@
 @import "../src/canvas-view/canvas-editor/canvas-editor.less";
 @import "../src/pagination/pagination.less";
 @import "../src/datepicker/datepicker.less";
-
+@import "../src/modals/modal-overlay/modal-overlay.less";
```
jeff-phillips-18: Don't we need the equivalent for SASS?

dtaylor113: SASS is auto-created.

jeff-phillips-18: not committed?

dtaylor113: Based on angular-patternfly's README.md, I believe it is automatically handled as part of the PR: "The Less to Sass Conversion step will be accomplished and managed as a part of any Pull Request which includes Less file changes. Although contributors may want to build and test their style changes with Sass before submitting a Pull Request, this step should always be tested and validated by reviewers before a style change is merged and released. "
..still looking to verify what exactly a-pf's TravisCI build cmd is to see if it does a '--sass'

jeff-phillips-18: OK, looks like it is picking it up automatically. Thought you'd need to add the new scss file to the imports but looks like we are smaht enough to pick it up.

dtaylor113: Hi, @jeff-phillips-18 you are correct, in the README.md "Note: When a Less file is added/deleted/renamed it needs to be updated in the main Less import file /styles/angular-patternfly.less and the main Sass import file styles/_angular-patternfly.scss."
@nicolethoen, I think you need to also include the new .less file in `styles/_angular-patternfly.scss`.

jeff-phillips-18: Ah, but it is not added to the dist/sass/_angular-patternfly.scss file so it won't be automatically included for applications that simply import the single file rather than all the scss files.

Please add the import to https://github.com/patternfly/angular-patternfly/blob/master/styles/_angular-patternfly.scss

src/modals/modal-overlay/modal-overlay.component.js
dtaylor113: It how we pass the `dismissCause` back to user defined callback:  `$scope.onClose2 = function(dismissCause) ...`

src/modals/examples/modal-overlay.js
dtaylor113: because this is the named pair we used in your comment below.  Has to do with the binding being `close: "&onClose",`

PR #734, opened 2018-04-18
Fix documentation example of pfTopologyMap

Hyperkid123:
I have noticed that the online docs does'n have the image resource that i've used (and it broke the example obviously).

I have replaced it with online resource


Hyperkid123:
@ping @dtaylor113 Hi can u check this out please? No changes in code, just replaced image resource so it will work on the online demo(didn't realize my local image won't work...).

dtaylor113:
@Hyperkid123, good catch -thanks :+1:

PR #735, opened 2018-04-20
feat(wizard): Add ability to set initial focus component per page

jeff-phillips-18:
## Description
Adds the ability to pass selectors for wizard step and substep pages in order to set the initial focus component for the page. An array of selectors is passed and will be used in order until a match if found. If no selectors are passed or none of the selectors match a component, then the next button is focuses if enabled, otherwise the cancel button is focused.

### DIFF
--- a/src/wizard/wizard-step.component.js
+++ b/src/wizard/wizard-step.component.js
@@ -20,6 +20,7 @@
  * @param {string=} description The step description (optional)
  * @param {object} wizardData Data passed to the step that is shared by the entire wizard
  * @param {function()=} onShow The function called when the wizard shows this step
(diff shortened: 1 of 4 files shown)


src/wizard/wizard-step.component.js
```diff
@@ -232,6 +236,29 @@ angular.module('patternfly.wizard').component('pfWizardStep', {
             ctrl.selectedStep.onShow();
           }
 
+          // Give time for onShow to do its thing (maybe update the selectors), then time to display the elements
+          $timeout(function () {
```
dtaylor113: yeah, never a good sign when you have to do this, but...

PR #742, opened 2018-06-04
feat(bulkselection): ListView - Created bulk selection component

chalettu:
This is a component that goes along with a toolbar to allow for select all/ select none and partial select of lists of data.

### DIFF
--- a/src/bulkselection/bulkselection.component.js
+++ b/src/bulkselection/bulkselection.component.js
@@ -1,0 +1,59 @@
+angular.module('patternfly.bulkselection').component('pfBulkSelection', {
+
+  bindings: {
+    totalRecords: '=',
+    selectedRecordCount: '=',
+    updateSelectionFn: '<'
+  },
+  templateUrl: 'bulkselection/bulkselection.html',
+  controller: function () {
+    'use strict';
+
+    var ctrl = this;
+    var BULKSELECTION_NONE = 'none';
+    var BULKSELECTION_ALL = 'all';
+    var BULKSELECTION_PARTIAL = 'partial';
+    ctrl.menuIsOpen = false;
+
+    ctrl.$onInit = function () {
+      ctrl.setBulkSelectionState(BULKSELECTION_NONE);
+    };
+    ctrl.updateBulkSelection = function (bulkSelectAction) {
+      ctrl.menuIsOpen = false;
+      ctrl.setBulkSelectionState (bulkSelectAction);
+      ctrl.updateSelectionFn(ctrl.bulkSelection);
+    };
+    ctrl.toggleBulkSelection = function () {
+      if (ctrl.selectedRecordCount > 0 || ctrl.bulkSelection === BULKSELECTION_ALL) {
+        ctrl.updateBulkSelection(BULKSELECTION_NONE);
+      } else {
+        ctrl.updateBulkSelection(BULKSELECTION_ALL);
+      }
+    };
+    ctrl.setBulkSelectionState = function(state) {
+      ctrl.bulkSelection = state;
+      switch (state) {
+      case BULKSELECTION_NONE:
+        ctrl.menuClass = 'fa-square-o';
+        break;
+      case BULKSELECTION_ALL:
+        ctrl.menuClass = 'fa-check-square all-selected';
+        break;
+      case BULKSELECTION_PARTIAL:
+        ctrl.menuClass = 'fa-minus-square-o partially-selected';
+        break;
+      }
+    };
+    ctrl.$doCheck = function () {
+      var recordsSelectedCount = ctrl.totalRecords - ctrl.selectedRecordCount;
+      if (ctrl.selectedRecordCount === 0) {
+        ctrl.setBulkSelectionState(BULKSELECTION_NONE);
+      }
+      if (recordsSelectedCount > 0 && ctrl.selectedRecordCount > 0 ) {
+        ctrl.setBulkSelectionState(BULKSELECTION_PARTIAL);
+      } else if (recordsSelectedCount === 0) {
+        ctrl.setBulkSelectionState(BULKSELECTION_ALL);
+      }
+    };
+  }
+});
--- a/test/toolbars/toolbar.spec.js
+++ b/test/toolbars/toolbar.spec.js
@@ -4,12 +4,13 @@ describe('Directive:  pfToolbar', function () {
   var element;
   var $pfViewUtils;
   var performedAction;
+  var bulkSelectionState;
 
   // load the controller's module
   beforeEach(function () {
-    module('patternfly.toolbars', 'patternfly.views', 'patternfly.filters', 'toolbars/toolbar.html',
+    module('patternfly.toolbars', 'patternfly.views', 'patternfly.bulkselection', 'patternfly.filters', 'toolbars/toolbar.html',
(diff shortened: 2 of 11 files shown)


chalettu:
Screen shot of intended functionality.  
bulk selection final demo

chalettu:
@serenamarie125 , Here is the bulk selection component that is functioning as discussed.  I don't know what the process for UX review and approval is in this repo but could you help me figure out how to get this reviewed?

chalettu:
For some reason the screenshot video is not truly showing the color right but partially selected and select all states use @color-pf-blue-400 when the states happen.  
This component compliments the toolbar and filter components.  It has 3 states (None, All, Partially selected).  This component has two ways to interact with it.  One method is a toggle if you click the square icon it will 
1. Toggle select all/select none
2. If you select one or more results from the list before clicking toggle, it will deselect all.

serenamarie125:
@chalettu is it possible to add still screenshots showing that states of the icon in the bulk selector?  Jut want to make sure that it matches - here are the high fidelity mocks
Partial Selection
bulkselector-someselected 1
All Selected
bulkselector-allselected-rowhover

beanh66:
@chalettu thanks, this looks great! I do have two questions but I think they are more design related so may need @mcarrano to chime in. 

Curious if the checkbox is too small, because it seems too difficult to click the checkbox without clicking the dropdown. Also thinking we should hide the dropdown for cases where there are no other options available other than select all, so just show the checkbox for these cases. Thoughts?

src/bulkselection/bulkselection.component.js
```diff
@@ -0,0 +1,59 @@
...
+      case BULKSELECTION_ALL:
+        ctrl.menuClass = 'fa-check-square all-selected';
+        break;
+      case BULKSELECTION_PARTIAL:
+        ctrl.menuClass = 'fa-minus-square-o partially-selected';
+        break;
+      }
+    };
+    ctrl.$doCheck = function () {
+      var recordsSelectedCount = ctrl.totalRecords - ctrl.selectedRecordCount;
```
dtaylor113: This variable is a little confusing and is too similar to `ctrl.selectedRecordCount`.  Perhaps a better name would be `unselectedRecords`.

test/toolbars/toolbar.spec.js
```diff
@@ -300,7 +307,43 @@ describe('Directive:  pfToolbar', function () {
     active = element.find('.active');
     expect(active.length).toBe(1);
   });
+  it ('should have bulk selection menu shown', function() {
```
dtaylor113: Appreciate the unit tests!  :-)

cshinn:
I'm not sure how tricky it is to do, but I think it's important that the selector be moved further left so that all the checkboxes align. The visual line created by those will make its functionality a lot more clear.

dtaylor113:
If I apply a fitter (Ex: Name='F'), then 'Select All' from the dropdown, it only selects rows shown by the filter.  The dropdown icon is 'Select All' in this state.  
If I clear the filter, the icon does change to 'Partially Selected'.  Just want to verify this is the behavior we want?  

Wondering if we want either: 
a) The 'Select All' to truly select all records, even ones not shown by the current filter, or..
b) If 'Select All' only selects visible records shown by filter, then the dropdown icon should be 'partially selected' since not all records are actually selected.
c) Perhaps when a filter is applied, the dropdown choice should be something like 'Select All Filtered'

-thanks

mcarrano:
This looks good. In answer to your question @beanh66 , while having the dropdown, even when the only choices are Select All and None, is indeed redundant, we felt this was necessary to make it clear what's happening here. I did not feel like just a checkbox inside a button would be very intuitive on it's own. Unlike for tables, the checkbox is a bit removed from the column headers here.

@dtaylor113 raises a good question about how this interacts with filtering. I was assuming that the selection only applied to what's currently loaded into the view, but maybe that's not the right answer. What do we do for tables? It's exactly the same use case. If I select all rows in a table and a filter is applied, does it select items that are filtered from the table? I wouldn't expect it to.

dtaylor113:
Unfortunately, this doesn't play well with Table View.  Table View already has a 'Select All' checkbox, which when clicked highlights the table rows, whereas the bulk selection dropdown doesn't highlight the rows:
selectallwtable

Interestingly enough, the table's 'Select All' checkbox works the same way irt. filtering as the bulk dropdown -namely, it only selects the visible filtered records.

Wondering if we shouldn't show this dropdown in table view, OR..does this bulk selection dropdown replace the table's 'Select All' checkbox?

dtaylor113:
> If I select all rows in a table and a filter is applied, does it select items that are filtered from the table? I wouldn't expect it to.

@mcarrano the table's 'Select All' checkbox does only select the filtered rows.

mcarrano:
@dtaylor113 Thanks for the additional clarification. A couple of thoughts...

1- I don't think the Bulk Selector should apply in Table View as it does become redundant with the built in table selection.

2- I had forgotten about this whole business with whether select all applies to items outside the current page. I think we need to remain consistent with what we had decided earlier for Table View. But maybe this is a case where we can use the menu to allow an option to select everything (whether in view or not). @serenamarie125 what do you think?

dtaylor113:
> From patternfly.org Table View design: "Select All Rows: Selecting the checkbox in the header row selects all rows on the page. The total number of rows selected is shown near the table action buttons."
> 
> 'select all rows on the page' <--- this even applies to Pagination. If I am only showing 5 records per page and click 'Select All' checkbox, it only selects the 5 rows on the current page.

So I believe the bulk selector is correct irt. filtering and pagination, at least it is mirroring what the table's 'Select All' checkbox is doing.  I'm still questioning whether both should be available in the table view.

serenamarie125:
agreed that we don't show it in our current table view implementation

serenamarie125:
@chalettu ah, yes the checkbox in the selector should be aligned with the checkbox in the list view.  Good find everyone!

dtaylor113:
> agreed that we don't show it in our current table view implementation

@chalettu, we do this already for the Sort button which we do not show in TableView because we use the table column headers to do sorting.  Pattern should be very similar for the bulk selector:

https://github.com/patternfly/angular-patternfly/blob/master/src/toolbars/examples/toolbar.js#L340
https://github.com/patternfly/angular-patternfly/blob/master/src/toolbars/toolbar.html#L9

mcarrano:
Wanted to call attention to a related conversation that is happening in the design repo here: https://github.com/patternfly/patternfly-design/issues/684 regarding the intermediate state of the checkbox. I am recommending that we remove that state from the design as it would require us to create a custom checkbox in Core which we don't want to do. In this case, the bulk selector would show the checked state when all are selected and unchecked if no selection or partial selection. Clicking the checkbox would toggle between all and none selected as it does currently for the Table.  There is currently a design PR in progress that will need to be updated.

dtaylor113:
Hi @chalettu, this PR is very close to the recently added Bulk Selector component: https://www.patternfly.org/pattern-library/forms-and-controls/bulk-selector/#

I only see a few visual differences:
1. bulk selector dropdown needs to be vertically aligned with results label on the line below
2. bulk selector dropdown menu options need to be changed to 'Select All' and 'Select None'
3. bulk selector tooltip? (can't tell if it's implemented)

The other issue is that this bulk selector component should be hidden when used with the TableView: 
"This pattern should NOT be used when information is displayed in a table view. The table view component includes a bulk selector in its toolbar."

Just curious, are you still working on this PR?  
Thanks,
- Dave

PR #743, opened 2018-06-05
Fix for #740: pfTableView-with-Toolbar: cannot select all rows after clearing filters

### LINKED ISSUE #740: pfTableView-with-Toolbar: cannot select all rows after clearing filters
The checkbox that selects all table rows stops working after removing filters.

**Steps to reproduce**
1. Open https://www.patternfly.org/angular-patternfly/#/api/patternfly.table.component:pfTableView-with-Toolbar
1. Filter by status 'ok'
1. Remove filter
1. Click on checkbox that selects all rows

dtaylor113:
Hi @alexkieling,
The 'Select All' table checkbox seems to lose binding to the controller's 'toggleAll()' method after (a) clearing all or any one filter or (b) applying a second filter.  I spent an entire afternoon trying to figure out why this is happening; trying many suggested solutions such as changing to `ng-click` or using an object for the `ng-model` value.
The only solution I could find is calling `addNewComponentToDOM()` for the usecases mentioned above.  This is not ideal, as the entire table is redrawn in those situations.  However, the controller bindings continue to work after multiple filters are applied or cleared.

As you can see, applying the first filter doesn't cause the table to be redrawn, however adding multiple filters or clearing filters will cause the redraw:

table-select-all-w-filters

This is somewhat of a work-around as nothing has changed in the actual TableView component, just in the ngDoc example, so feel free to use the suggested solution if it's acceptable to you.

Fixes #740 ? :-)

Thanks,
- Dave

### DIFF
--- a/src/table/tableview/examples/table-view-with-toolbar.js
+++ b/src/table/tableview/examples/table-view-with-toolbar.js
@@ -360,6 +360,9 @@
         } else {
           $scope.items = $scope.allItems;
         }
+        if (filters && (filters.length === 0 || filters.length > 1)) {
+          $scope.addNewComponentToDOM();
+        }
       };
 
       var filterChange = function (filters) {


alexkieling:
Your solution is to remove and add the whole table using ngIf, correct? Did you try doing the same thing just for the checkbox?

dtaylor113:
> Your solution is to remove and add the whole table using ngIf, correct? 

Yes, in the demo code

> Did you try doing the same thing just for the checkbox?

No, I haven't tried that.  The 'Select All' checkbox is part of the transcluded HTML which is bound to underlying datatables.net component.  Not sure `ngif` there would help.  

angular-datatables has an example of the select all checkbox, I was going to compare it to how it is implemented in angular-patternfly:  http://l-lin.github.io/angular-datatables/archives/#!/rowSelect

dtaylor113:
@alexkieling do you have any issues with me merging this PR?  -thanks

alexkieling:
@dtaylor113 no issues, go ahead

Issue #745, opened 2018-06-13
pfVerticalNavigation: on mobile, move the masthead utility items to the vertical navigation


alexkieling:
On mobile (less than 768 px), the masthead utility items (Help, About, Logout, etc.) are hidden but the pfVerticalNavigation component doesn't support dynamically showing them in the vertical navigation. I suggest adding a new property to the `items` objects to flag what items should be hidden by default and shown on mobiles.
More details in the [PatternFly Vertical Navigation design docs](http://www.patternfly.org/pattern-library/navigation/vertical-navigation/#design
).

dtaylor113:
Hi @alexkieling,
Is this what you had in mind?  I added a `mobileOnly` flag to the items.

Note that **'Help'** and **'User'** vert. nav items only appear when in mobile (< 768px) mode, and are hidden when not in mobile mode:
![mobileonly](https://user-images.githubusercontent.com/12733153/41430999-96a34efc-6fdf-11e8-98a9-54bfaa9fc0da.gif)

alexkieling:
Yes, that's exactly what I had in mind. It seems to match the [PatternFly design](http://www.patternfly.org/pattern-library/navigation/vertical-navigation/#design) perfectly.
@tadayosi, what do you think?

tadayosi:
@alexkieling Looks good to me. Thanks.

dtaylor113:
Fixed via #747, not sure why it didn't auto-close

Issue #746, opened 2018-06-13
pfVerticalNavigation: masthead utility items are hidden when `ignore-mobile` is true


alexkieling:
**Current behavior**
Masthead utility items are hidden on mobile when `ignore-mobile` parameter is true
**Expected behavior**
Masthead utility items should not be hidden on mobile when `ignore-mobile` parameter is true

dtaylor113:
From @jeff-phillips-18, posting here so we don't lose this bit of work  -thanks Jeff

"I played around with the ignoreMobile stuff in A-PF
I have the `less` necessary to hide/show in the header based on resolution and the ignoreMobile flag 
Have to add `ignore-mobile` on `navbar-header` and `navbar-collapse`
then the less is:
```
.navbar-header {
  &.ignore-mobile {
    float: left;
  }
}

.navbar-collapse {
  &.ignore-mobile {
    width: auto;
    border-top: 0;
    box-shadow: none;

    .navbar-nav {
      float: left;
      margin: 0;

      &.navbar-right {
        float: right;
        .dropdown-menu {
          left: auto;
          right: 0;
        }
      }

      > li {
        float: left;
      }

      .open .dropdown-menu {
        position: absolute;
        margin-top: 0;
        background-color: @color-pf-white;
        border: 1px solid @color-pf-black-400;
        box-shadow: 0 6px 12px rgba(0, 0, 0, 0.175);
      }
    }
  }
}
```

PR #747, opened 2018-06-14
pfVerticalNavigation: on mobile, move the masthead utility items to the vertical navigation

dtaylor113:
## Description
Added a `mobileOnly` property to nav. item objects; when 'true' items only display in mobile mode. 
Addresses: #745 

Ex: The **Help** and **User** vert. nav items only appear when in mobile (< 768px) mode, and are hidden when not in mobile mode:
mobileonly

Note the masthead utility items (Help & User dropdowns) continue to be hidden at mobile; this PR allows app devs to mimic them 'moving' to the vert nav menu.

@alexkieling

## PR Checklist

### DIFF
--- a/src/navigation/examples/vertical-navigation-basic.js
+++ b/src/navigation/examples/vertical-navigation-basic.js
@@ -35,6 +35,8 @@
  *   <li>.tooltip      - (string) Tooltip to display for the badge
  *   <li>.badgeClass:  - (string) Additional class(es) to add to the badge container
  *   </ul>
+ * <li>.mobileOnly     - (boolean) When set to 'true', menu item will only be displayed when browser is in mobile mode (<768px).
+ * When <code>ignoreMobile</code> flag set to 'true', <code>mobileOnly</code> items are not displayed.
  * </ul>
  * @param {function} navigateCallback function(item) Callback method invoked on a navigation item click (one with no submenus)
  * @param {function} itemClickCallback function(item) Callback method invoked on an item click
@@ -76,7 +78,7 @@
               <span class="caret"></span>
             </a>
             <ul class="dropdown-menu" aria-labelledby="dropdownMenu2">
-              <li><a href="#">Preferences</a></li>
+              <li><a href="#">User Preferences</a></li>
               <li><a href="#">Logout</a></li>
             </ul>
           </li>
@@ -392,9 +394,19 @@
            href: "#/adipscing"
         },
         {
-           title: "Lorem",
-           iconClass: "fa fa-gamepad",
-           href: "#/lorem"
+           title: "Help",
+           iconClass: "fa pficon-help",
+           href: "#/help",
+           mobileOnly: true
+        },
+        {
+           title: "User",
+           iconClass: "fa pficon-user",
+           mobileOnly: true,
+           children: [
+              { title: "User Preferences" },
+              { title: "Logout" }
+           ]
         },
         {
            title: "Exit Demo"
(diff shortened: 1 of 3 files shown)


src/navigation/examples/vertical-navigation-basic.js
```diff
@@ -392,9 +394,19 @@
...
         },
         {
-           title: "Lorem",
-           iconClass: "fa fa-gamepad",
-           href: "#/lorem"
+           title: "Help",
+           iconClass: "fa pficon-help",
+           href: "#/help",
+           mobileOnly: "true"
+        },
```
jeff-phillips-18: This should have Help and About as children to be consistent with the header.

src/navigation/vertical-navigation.html
```diff
@@ -30,7 +30,7 @@
                     'force-hide-secondary-nav-pf': $ctrl.forceHidden,
                     'show-mobile-nav': $ctrl.showMobileNav}">
       <ul class="list-group">
-        <li ng-repeat="item in $ctrl.items" class="list-group-item"
+        <li ng-repeat="item in $ctrl.items | navItems:$ctrl.inMobileState" class="list-group-item"
```
jeff-phillips-18: Might be easer just to add:

```
'visible-xs': item.mobileOnly
```

to the `ng-class` rather than filtering.

src/navigation/examples/vertical-navigation-basic.js
```diff
@@ -392,9 +394,19 @@
...
+        {
+           title: "User",
+           iconClass: "fa pficon-user",
+           mobileOnly: "true",
+           children: [
+              { title: "Preferences" },
```
jeff-phillips-18: Can we change `Preferences` to `User Preferences` both here and in the header area? It was a UX request in the patternfly-react repo, so might as well update here as well.

src/navigation/examples/vertical-navigation-basic.js
jeff-phillips-18: These changes should be made in the `-router` example as well

src/navigation/examples/vertical-navigation-basic.js
```diff
@@ -35,6 +35,8 @@
  *   <li>.tooltip      - (string) Tooltip to display for the badge
  *   <li>.badgeClass:  - (string) Additional class(es) to add to the badge container
  *   </ul>
+ * <li>.mobileOnly     - (string) When set to 'true', menu item will only be displayed when browser is in mobile mode (<768px).
```
jeff-phillips-18: why use a string here rather than a boolean?

src/navigation/examples/vertical-navigation-basic.js
dtaylor113: > This should have Help and About as children to be consistent with the header.

Wasn't crazy about having sub-menus:
```
Help -> Help
        About
```
So left out 'About' as maybe something you wouldn't need at mobile.

src/navigation/vertical-navigation.html
dtaylor113: but is that acceptable performance wise?  Seems a waste to pull in/construct all that dom/html only to not display it.  Seems like it would be faster just filtering items.

src/navigation/examples/vertical-navigation-basic.js
jeff-phillips-18: OK, just thinking we should show consistency.

dtaylor113:
Hi @jeff-phillips-18, I believe I have addressed your comments, except the one about the Help-at-mobile menu, IMO I think leaving off the 'About' menu item represents a truer @mobile menu.

dtaylor113:
> Might want to address #746 while you are working here...

I'd prefer to do that in a follow up pr, thanks :-)

Issue #730, opened 2018-03-25
pfBootstrapDatepicker does not have an output binding


wywywywy:
At the moment you can input a Date to the pfBootstrapDatepicker, but it has no output because it is a one way binding, so this does not make the component very useful.

It should have an onUpdate & binding to a callback to the parent.

```
bindings: {
  onUpdate: '&'
}
```

Or am I missing something?

PR #749, opened 2018-07-02
fix(pfBootstrapDatepicker): added callback function when date changes

dtaylor113:
## Description
Added two-way binding on `Date` property and added the ability for user to define a callback function when date is changed via the picker.

Fixes: #730 

image
image

## PR Checklist


nicolethoen:
LGTM

dtaylor113:
> Do we need the two-way binding if we have the callback?

Probably not.  I had thought it was an ask in #730 but re-reading the issue maybe not.   Would two way binding help app devs if they had a `$watch` on the var.?

dtaylor113:
Hi @jeff-phillips-18, removed two way binding and added unit test  PTAL  -thanks

PR #752, opened 2018-07-23
fix(pfFilter): unsafe:javascript:void(0) redirect when clearing filters

### LINKED ISSUE #751: pfFilter - unsafe:javascript:void(0) redirect when clearing filter types
More context here: https://github.com/ManageIQ/manageiq-ui-service/issues/1456#issuecomment-407077066

This issue manifests when using safari, redirecting to the url `unsafe:javascript:void(0)` instead of clearing the active filters...

Looks like safari doesn't like this line... https://github.com/patternfly/angular-patternfly/blob/05baa47a34862deb0134a01fd92a7b4681a1ad94/src/filters/simple-filter/filter-results.html#L16

which might mean all uses of this might be blowing up 🤔

AllenBW:
fixes #751

Use `#0` instead for greater browser compatibility 

also solves the second half of this thing: https://github.com/ManageIQ/manageiq-ui-service/issues/1456

### DIFF
--- a/src/filters/filter-panel/filter-panel-results-component.js
+++ b/src/filters/filter-panel/filter-panel-results-component.js
@@ -57,8 +57,9 @@ angular.module('patternfly.filters').component('pfFilterPanelResults', {
       }
     }
 
-    function clearFilter (filter, value) {
+    function clearFilter (evt, filter, value) {
       var changedFilterId = filter.id;
+      evt.preventDefault();
 
       _.pull(filter.values, value);
 
@@ -71,7 +72,9 @@ angular.module('patternfly.filters').component('pfFilterPanelResults', {
       }
     }
 
-    function clearAllFilters () {
+    function clearAllFilters (evt) {
+      evt.preventDefault();
+
       ctrl.config.appliedFilters = [];
 
       if (ctrl.config.onFilterChange) {
--- a/src/filters/filter-panel/filter-panel-results.html
+++ b/src/filters/filter-panel/filter-panel-results.html
@@ -13,13 +13,13 @@
         <ul class="list-inline filter-pf-category-values">
           <li ng-repeat="value in filter.values">
             <span class="label label-info">{{value}}
-              <a href="javascript:void(0);"><span ng-click="$ctrl.clearFilter(filter, value)" class="pficon pficon-close"></span></a>
+              <a href="#"><span ng-click="$ctrl.clearFilter($event, filter, value)" class="pficon pficon-close"></span></a>
             </span>
           </li>
         </ul>
       </span>
       </li>
     </ul>
-    <p><a href="javascript:void(0);" ng-click="$ctrl.clearAllFilters()" ng-if="$ctrl.config.appliedFilters.length > 0">Clear All Filters</a></p>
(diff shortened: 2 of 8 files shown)


spadgett:
This will change back button behavior (you have to click back twice now).

AllenBW:
@spadgett down to modify the value changed! but its gotta change, safari is **not** a fan

spadgett:
`href=""` would fix both problems, but IIRC that breaks test in this repo.

Can you prevent default on the event when the link is clicked?

dtaylor113:
Hmm...we also use `javascript:void()` for pfEmptyState because in OpenShift we didn't want to redirect:  https://github.com/patternfly/angular-patternfly/pull/703

AllenBW:
Ok I had another idea... docs say `href` (alone) is valid... 

`<a id="link-1" href ng-click="value = 1">link 1</a> (link, don't reload)<br />`

and tests are happy... 

Thoughts? 🤔 

https://docs.angularjs.org/api/ng/directive/ngHref

spadgett:
Can you check if that breaks keyboard / screenreader accessibility of the
link?

AllenBW:
false alarm, gotta sort out https://github.com/patternfly/angular-patternfly/blob/bada308131135398987793798ed850911180892c/src/filters/filter-panel/filter-panel-results.html#L16 first

dtaylor113:
@AllenBW, I just tried your latest changes, broke unit tests:
```
PhantomJS 2.1.1 (Linux 0.0.0) ERROR
  Some of your tests did a full page reload!
PhantomJS 2.1.1 (Linux 0.0.0): Executed 136 of 393 ERROR (1.721 secs / 0 secs)
```

AllenBW:
@dtaylor113 see above comment, yah, href only works if there's an ngclick, reworking tests to be happy

AllenBW:
@dtaylor113 @spadgett, ok new idea, what if we *did not* use links, rather apply the button role to a span (it'll still do the pointer thing, work with the ng-click and everything passes)

spadgett:
> ok new idea, what if we did not use links, rather apply the button role to a span (it'll still do the pointer thing, work with the ng-click and everything passes)

It's not accessible even adding the button role.

I'm not sure why calling preventDefault on $event in the on click handler wouldn't work?

spadgett:
Why not 

```html
<a href="#" onClick="callback($event)">My action</a>
```

```js
callback (e) {
  e.preventDefault();
  // ...
}
```

AllenBW:
> It's not accessible even adding the button role.

?

spadgett:
`span` with `role="button"` is not keyboard or screenreader accessible.

AllenBW:
sadly `<button class="btn btn-link">` would change the visual appearance 😭

AllenBW:
I was hoping for a slightly less invasive fix, but whatever gets in soonest will be just 🍑y for meeeeee

@spadgett @dtaylor113 @jeff-phillips-18 👉 this should now be g2g

dtaylor113:
LGTM, thanks @AllenBW :-)

dtaylor113:
@AllenBW will merge today, just giving @spadgett a chance to weigh in.

spadgett:
works for me :)

PR #756, opened 2018-08-09
pfModalOverlay: allow async validations while keeping modal open, demo maxlength validation

### LINKED ISSUE #755: pfModalOverlay: no way to close modal for custom validations
If modal has a form and form validation depends on API response, the modal closes itself without waiting for API response and show error. There seems to be no way to stop closing of modal.

### LINKED ISSUE #757: Broken Semantic Release
The semantic release appears to be broken. The version of semantic release was recently updated to ^15.0.0, but that change never successfully published a package. The latest release (19 days ago) had to be published manually.

Note that ^15.0.0 is currently installing 15.8.1, which requires Node 8. Travis is currently using Node 4.

In addition, package.json references commands that no longer exist. For example, `pre` and `post` should be `--pre` and `--post`:

"semantic-release": "semantic-release pre && npm publish && semantic-release post",
"semantic-release-post": "semantic-release post",
"semantic-release-pre": "semantic-release pre" 

See changes here: https://github.com/patternfly/angular-patternfly/pull/750

dtaylor113:
## Description
Fixes #755

Added the ability to return `false` for actionBtn function(s) to keep modal open.  Allows app dev to click 'Save', show spinner & perform async validation, and if invalid display error msg, or if valid, programmatically close the modal:
Also demonstrated how to perform `ng-maxlength` validations.
modal_overlay

Also fixes #757 (Broken Semantic Release)

## PR Checklist

- [X] Unit tests are included


beanh66:
Following the [Errors and Validation](https://www.patternfly.org/pattern-library/forms-and-controls/errors-and-validation/#overview) pattern would include an inline notification at the top with the specific field-level errors listed out by each field. Below is an example of what I mean:
errors

dtaylor113:
Hi @beanh66, ok, sounds good.  The example here actually can have several error messages, how would that be displayed in the modal?

beanh66:
The summary is just in the inline notification so I think that could stay the same...input is not valid, fix the errors, etc. Then the additional errors would be highlighted below. Does that make sense? 

Something like this: 
errors-2

dtaylor113:
@beanh66, ok, thanks, that makes sense.  I'll work on it some more.  I may create another Issue/PR for this additional work because it's not what this PR was directly addressing.  -thanks

dtaylor113:
> I may create another Issue/PR for this additional work because it's not what this PR was directly addressing

...for the inline notification at the top that is, the field level help shouldn't be an issue for this PR.

dtaylor113:
Hi @beanh66, updated the screenshot/video in this PR's description. PTAL  -thanks!

dtaylor113:
@jeff-phillips-18, hoping you could take a look -thanks

src/modals/modal-overlay/modal-overlay.component.js
```diff
@@ -103,9 +109,12 @@ angular.module('patternfly.modals').component('pfModalOverlayContent', {
 
       ctrl.ok = function (label, actionFn) {
         if (typeof actionFn === "function") {
-          actionFn();
+          if (actionFn() !== false) {
```
jeff-phillips-18: Only concern would be an existing application returning false but not expecting the modal to remain open. Unlikely unless it is just happenstance. I think it's OK to assume that unlikelyhood is great enough to proceed.

dtaylor113: Yeah, I tend to agree that scenario is unlikely.  If an app was returning `false`, I'm not sure why or what controller would be processing a true or false?  The component executes the button actionFns and it wasn't expecting any return values.

